PR #4, opened 2021-10-27
api: add ReclaimSpaceJob api

Rakshith-R:
# ReclaimSpaceJob

ReclaimSpaceJob is a namespaced custom resource designed to invoke reclaim space operation on target volume.

```yaml
apiVersion: csiaddons.openshift.io/v1beta1
kind: ReclaimSpaceJob
metadata:
  name: sample-1
spec:
  target:
    pvc: pvc-1
  backOffLimit: 10
  activeDeadlineSeconds: 900
```

+ `target` represents the volume target on which the reclaim space operation will be performed.
    + `pvc` contains a string indicating the name of `PersistentVolumeClaim`.
+ `backOfflimit` specifies the number of retries before marking reclaim space operation as failed. If not specified, defaults to 6.
+ `activeDeadlineSeconds` specifies the duration in seconds relative to the startTime that the operation might be retried; value must be positive integer. If not specified, defaults to 600 seconds.

```
[rakshith@fedora kubernetes-csi-addons]$ k get reclaimspacejobs.csiaddons.openshift.io sample
NAME     NAMESPACE   AGE     RETRIES   RESULT
sample   rook-ceph   4h37m   16        Failed.
```

### DIFF
--- a/api/v1alpha1/reclaimspacejob_types.go
+++ b/api/v1alpha1/reclaimspacejob_types.go
@@ -1,0 +1,117 @@
+/*
+Copyright 2021 The Kubernetes-CSI-Addons Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+import (
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// OperationResult represents the result of reclaim space operation.
+type OperationResult string
+
+const (
+	// OperationResultSucceeded represents the Succeeded operation state.
+	OperationResultSucceeded OperationResult = "Succeeded"
+
+	// OperationResultFailed represents the Failed operation state.
+	OperationResultFailed OperationResult = "Failed"
+)
+
+// TargetSpec defines the targets on which the operation can be
+// performed.
+type TargetSpec struct {
+	// PersistentVolumeClaim specifies the target PersistentVolumeClaim name.
+	PersistentVolumeClaim string `json:"persistentVolumeClaim,omitempty"`
+}
+
+// ReclaimSpaceJobSpec defines the desired state of ReclaimSpaceJob
+type ReclaimSpaceJobSpec struct {
+	// Target represents volume target on which the operation will be
+	// performed.
+	// +kubebuilder:validation:Required
+	Target TargetSpec `json:"target"`
+
+	// BackOffLimit specifies the number of retries allowed before marking reclaim
+	// space operation as failed. If not specified, defaults to 6. Maximum allowed
+	// value is 60 and minimum allowed value is 0.
+	// +optional
+	// +kubebuilder:validation:Maximum=60
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:default:=6
+	BackoffLimit int32 `json:"backOffLimit"`
+
+	// RetryDeadlineSeconds specifies the duration in seconds relative to the
+	// start time that the operation may be retried; value MUST be positive integer.
+	// If not specified, defaults to 600 seconds. Maximum allowed
+	// value is 1800.
+	// +optional
+	// +kubebuilder:validation:Maximum=1800
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:default:=600
+	RetryDeadlineSeconds int64 `json:"activeDeadlineSeconds"`
+}
+
+// ReclaimSpaceJobStatus defines the observed state of ReclaimSpaceJob
+type ReclaimSpaceJobStatus struct {
+	// Result indicates the result of ReclaimSpaceJob.
+	Result OperationResult `json:"result,omitempty"`
+
+	// Message contains any message from the ReclaimSpaceJob.
+	Message string `json:"message,omitempty"`
+
+	// ReclaimedSpace indicates the amount of space reclaimed.
+	ReclaimedSpace resource.Quantity `json:"reclaimedSpace,omitempty"`
+
+	// Conditions are the list of conditions and their status.
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	// Retries indicates the number of times the operation is retried.
+	Retries        int32        `json:"retries,omitempty"`
+	StartTime      *metav1.Time `json:"startTime,omitempty"`
+	CompletionTime *metav1.Time `json:"completionTime,omitempty"`
+}
+
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:status
+//+kubebuilder:printcolumn:JSONPath=".metadata.namespace",name=Namespace,type=string
+//+kubebuilder:printcolumn:JSONPath=".metadata.creationTimestamp",name=Age,type=date
+//+kubebuilder:printcolumn:JSONPath=".status.retries",name=Retries,type=integer
+//+kubebuilder:printcolumn:JSONPath=".status.result",name=Result,type=string
+//+kubebuilder:printcolumn:JSONPath=".status.reclaimedSpace",name=ReclaimedSpace,type=string,priority=1
+
+// ReclaimSpaceJob is the Schema for the reclaimspacejobs API
+type ReclaimSpaceJob struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   ReclaimSpaceJobSpec   `json:"spec,omitempty"`
+	Status ReclaimSpaceJobStatus `json:"status,omitempty"`
+}
+
+//+kubebuilder:object:root=true
+
+// ReclaimSpaceJobList contains a list of ReclaimSpaceJob
+type ReclaimSpaceJobList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []ReclaimSpaceJob `json:"items"`
+}
+
+func init() {
+	SchemeBuilder.Register(&ReclaimSpaceJob{}, &ReclaimSpaceJobList{})
+}
--- a/docs/reclaimspace.md
+++ b/docs/reclaimspace.md
@@ -1,0 +1,23 @@
+# ReclaimSpaceJob
+
+ReclaimSpaceJob is a namespaced custom resource designed to invoke reclaim space operation on target volume.
+
+```yaml
+apiVersion: csiaddons.openshift.io/v1alpha1
+kind: ReclaimSpaceJob
+metadata:
+  name: sample-1
+spec:
+  target:
+    persistentVolumeClaim: pvc-1
+  backOffLimit: 10
+  retryDeadlineSeconds: 900
+```
+
+
++ `target` represents volume target on which the operation will be performed.
+    + `persistentVolumeClaim` contains a string indicating the name of `PersistentVolumeClaim`.
++ `backOfflimit` specifies the number of retries before marking reclaim space operation as failed. If not specified, defaults to 6. Maximum allowed value is 60 and minimum allowed value is 0.
++ `retryDeadlineSeconds` specifies the duration in seconds relative to the start time that the operation may be retried; value must be positive integer. If not specified, defaults to 600 seconds. Maximum allowed value is 1800.
+
+Further extensions like `ReclaimSpaceCronJob` and controller to automatically create `ReclaimSpaceCronJob` based on annotation on PersistentVolumeClaim are being planned.


Rakshith-R:
For testing, this branch can be used https://github.com/Rakshith-R/kubernetes-csi-addons/tree/init-final

CRDs : 
https://github.com/Rakshith-R/kubernetes-csi-addons/blob/init-final/config/crd/bases/csiaddons.openshift.io_csiaddonsjobs.yaml
sample: https://github.com/Rakshith-R/kubernetes-csi-addons/blob/init-final/config/samples/csiaddons_v1beta1_csiaddonsjob.yaml

Changed it to strongly typed format and added `target` section to hold pvc/pv ( maybe snapshot later on).

```yaml
apiVersion: csiaddons.openshift.io/v1beta1
kind: CSIAddonsJob
metadata:
  name: sample-1
spec:
  operation:
    reclaimSpace:
      target:
        pvc: pvc-1
      parameters:
        key1: value1
        key2: value2
  backOffLimit: 4
  activeDeadlineSeconds: 500
  ```

File: docs/csiaddonsjob.md [hunk 1]
```diff
@@ -0,0 +1,44 @@
...
+```
+
++ `operation` contains the required information to invoke operation on target volume.
+    + `reclaimSpace` contains the required information to invoke `reclaimSpace`  operation on target volume.
+      + `target` represents the volume target on which the operation can be performed. Only one target needs to be specified.
+        + `pvc` contains a string indicating the name of `PersistentVolumeClaim`.
+        + `pv` contains a string indicating the name of `PersistentVolume`.
+      + `parameters` contains key-value pairs that are passed down to the driver.
++ `backOfflimit` specifies the number of retries before marking this operation failed.
++ `activeDeadlineSeconds` specifies the duration in seconds relative to the startTime that the operation might be active; value must be positive integer.
```
Madhu-1, line 44: what is activeDeadLine here? if the operator didn't receive any response at this time it will resend the request to the driver again?

docs/csiaddonsjob.md
```diff
@@ -0,0 +1,44 @@
...
+
+CSIAddonsJob is a namespaced custom resource designed to invoke operation to be performed on a volume.
+
+```yaml
+apiVersion: csiaddons.openshift.io/v1beta1
+kind: CSIAddonsJob
+metadata:
+  name: sample-1
+spec:
+  operation:
```
Madhu-1, line 11: dont we need a schedule here? so that every interval job will be triggered?

api/v1beta1/csiaddonsjob_types.go
```diff
@@ -0,0 +1,117 @@
...
+	// +kubebuilder:validation:Optional
+	ActiveDeadlineSeconds int64 `json:"activeDeadlineSeconds"`
+
+	// BackOffLimit Specifies the number of retries before marking this operation failed.
+	// +kubebuilder:validation:Optional
+	BackoffLimit int32 `json:"backOffLimit"`
+}
+
+// CSIAddonsJobStatus defines the observed state of CSIAddonsJob
+type CSIAddonsJobStatus struct {
```
Madhu-1, line 82: this will provide a last of last operation what if someone wants to see the status of all? condition suits better here

docs/csiaddonsjob.md
```diff
@@ -0,0 +1,44 @@
...
+    reclaimSpace:
```
Madhu-1, line 12: can this be a generic one as below? if we keep it generic one we don't need to introduce a new CR for every operation we support.

```yaml
operation: reclaimSpace
target:
    pv: pvc-1232-4255-5675
parameters:
   key1: value1
   key2: value2
```

api/v1beta1/csiaddonsjob_types.go
```diff
@@ -0,0 +1,117 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// TargetSpec defines the targets on which the operation can be
+// performed.
+type TargetSpec struct {
+	// PVC represents PersistentVolumeClaim name.
+	PVC string `json:"pvc,omitempty"`
+
+	// PVC represents PersistentVolume name.
```
Madhu-1, line 29: PV represents

docs/csiaddonsjob.md
```diff
@@ -0,0 +1,44 @@
...
+---
+apiVersion: csiaddons.openshift.io/v1beta1
+kind: CSIAddonsJob
+metadata:
+  name: sample-2
+spec:
+  operation:
+    reclaimSpace:
+      target:
+        pv: pvc-1232-4255-5675
```
Madhu-1, line 29: as this is a namespaced resource one validation we need to add is that PV is bound to a PVC that belongs to the same namespace? or remove this from spec for now and add it later when it's required.

docs/csiaddonsjob.md
Rakshith-R: > what is activeDeadLine here? if the operator didn't receive any response at this time it will resend the request to the driver again?

I'll tweak the description. 
its supposed to be : 
>+ `activeDeadlineSeconds` specifies the duration in seconds relative to the startTime within which the operation might be retried; value must be positive integer.

This check is performed just before a request is sent to the driver. 

> if the operator didn't receive any response at this time it will resend the request to the driver again?

Nothing has been decided on this yet.

docs/csiaddonsjob.md
Rakshith-R: > dont we need a schedule here? so that every interval job will be triggered?

`csiAddonsJob` will run only once, whereas `csiAddonsCronJob` will run at regular interval which will be added soon.

api/v1beta1/csiaddonsjob_types.go
Rakshith-R: > this will provide a last of last operation what if someone wants to see the status of all? condition suits better here

Yes, I am looking into how status / condition fields are updated/used according to api specification, I'll update it accordingly.

docs/csiaddonsjob.md
Rakshith-R: >can this be a generic one as below? if we keep it generic one we don't need to introduce a new CR for every operation we support.

Previous iteration had similar design, it can be found here https://github.com/Rakshith-R/kubernetes-csi-addons/blob/api-job-generic/docs/csiaddonsjob.md . 
But @nixpanic preferred strongly typed CR 
refer discussion :https://github.com/csi-addons/kubernetes-csi-addons/pull/4#discussion_r738982807

docs/csiaddonsjob.md
Rakshith-R: > as this is a namespaced resource one validation we need to add is that PV is bound to a PVC that belongs to the same namespace? 

Yes, the plan is to have some kind of validation to verify user can actually perform operation on that PV.

docs/csiaddonsjob.md
Madhu-1: as we add more operations this will get extended a lot?

docs/csiaddonsjob.md
Madhu-1: drop PV now and add when there is a real requirement for it? we can always get to the PV if CR in a namespace has the corresponding PVC name.

docs/csiaddonsjob.md
nixpanic: yes, we'll extend this for each operation. Every operation may have different parameters. Maybe we want to add a `fsck` operation, and that may well need a `dry-run`, `force` or `clone-as-backup` option that is not applicable for `reclaimSpace`.

Madhu-1: as the parameters are opaque the user can specify the parameter as per the operations? so that a single operation can be reused instead of extending the spec?

nixpanic: That makes it more difficult to validate, and generated documentation could easily be incomplete. With a strongly typed CRD, it is very clear what operations are available, and what options they accept.

nixpanic: Also, from my understanding, it is not an issue to extend a CRD, as long as the existing members/values stay valid. Therefore, it should not be problematic to add more operations later.

Rakshith-R: I prefer generic approach too,IMO:

- CSIAddonsJob CR will just be a yaml/json config file which the user has to write by referring the **documentation**(which will be responsible for clearly stating the support for operations and their options). 
- Validation may inturn be in form of having 
  - enums set on the fields (like operation: <supported_types> where only 'reclaimSpace','fsck' will be accepted).  
  - admission webhooks which reject at CR create request with clear messages and which can be easily programmed.
  - At the controller level, through error messages(this can be driver specific validation, while the other two will be generic). 
- I prefer the CR & CRD to be simple and the need to extend/change it to be very minimal. 
  

With strongly typed CR,CRD: 

- The need of change will be very often, with change in very operation/ option.
- The code for the api may indeed become large and confusing(very low reusability)
- The documentation may get confusing too.
- versioning of the api may get complicated too (?).

Madhu-1: Agree with the above points that what I have in my mind to suggest is to keep it generic instead of strongly typed CR's. IMHO it may look simple for now but when we start adding more operations it becomes complex.

nixpanic: For each operation we need to extend things somewhere. It is just that the location(s) where the extensions need to be done should be kept as closely together (ideally a single place).

I prefer to extend things at a location so that other parts can use the (re)generated results. If the CRD can be validated easily, the admission webhook would mostly be using generated code. The same counts for (generated) documentation of the valid operations and their parameters. Anything that is done should work together with [Server-Side Apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/) as well.

Requiring users to review the documentation which contains a description like a switch-statement, depending on the (string :fearful:?) value that they provide as an argument, will be very user unfriendly.

Remember that we expect users (persons, not automation) to create the CR. Input validation is where a lot of bugs are  happening, and reducing that from occurring is important. Any validation that we manually need to write has a high chance on containing issues, whereas a strong-typed automatically generated validation should have fewer problems.

For me, the interactions that the user needs to do are most important. Users will be able to make less mistakes when the CRD is strongly typed. Recent versions of `kubectl` give reasonable errors when the CR contains invalid values. This uses standard Kubernetes mechanisms, just like any other object that the user may create. The common usage, standard error reporting and re-using existing functions outweighs the required updating of the CRD (which would not happen *that* often once common storage maintenance operations are included) IMHO.

BlaineEXE:
> Is there a design doc that provides more details? It would help to understand the goals and requirements of the add-ons. Specific examples of the scenarios that will be solved will also be helpful.

Agreed.

Rakshith-R:
> Is there a design doc that provides more details? It would help to understand the goals and requirements of the add-ons. Specific examples of the scenarios that will be solved will also be helpful.

Here is the link to the overall design doc:
https://hackmd.io/@JYGiDmq1TX6sgAT6cA9JPg/H1rZF_nEY

This pr focuses on adding user facing CR based on which the CSI-Addons controller will perform `SpaceReclaim` operation (more operations will be added later on).

 Most of the discussions happening on the pr involves around whether to have strongly typed API for each operation or have single generic structure. 
 
 Each CR will only execute one operation.
 
 ```yaml
 apiVersion: csiaddons.openshift.io/v1beta1
kind: CSIAddonsJob
metadata:
  name: sample-1
spec:
  operation:
    reclaimSpace:
      target:
        pvc: pvc-1
      parameters:
        key1: value1
  backOffLimit: 4
  activeDeadlineSeconds: 500
  ---
  # or 
  apiVersion: csiaddons.openshift.io/v1beta1
kind: CSIAddonsJob
metadata:
  name: sample-1
spec:
  operation:
    name: "reclaimSpace"
    target:
      name: pvc-1
      kind: PersistentVolumeClaim
    parameters:
      key1: value1
      key2: value2
  backOffLimit: 4
  activeDeadlineSeconds: 500
 ```

related references :

- https://github.com/csi-addons/kubernetes-csi-addons/blob/main/README.md
- https://github.com/csi-addons/kubernetes-csi-addons/pull/6
- https://github.com/csi-addons/spec/pull/28
- https://github.com/csi-addons/spec/pull/30

@travisn @BlaineEXE Please take a look 
Thanks.

travisn:
A few overall questions about the approach:
- What if the volume is in use at the time the trim is triggered? Is there any risk of corruption?
- Is the volume mounted by some new pod temporarily on the same node? Or how will the volume be accessed?
- Are these operations something that could be done by a sidecar? Using an independent CRD seems like a very disconnected way to apply some operation on a volume. If there is a sidecar in the same pod, the application has more control.

nixpanic:
> A few overall questions about the approach:
> 
>     * What if the volume is in use at the time the trim is triggered? Is there any risk of corruption?

There is no risk for corruption. If a CSI-driver runs `fstrim`, it should do so on the already mounted volume+filesystem on a node where a Pod has write-access to the filesystem. For storage cluster operations (think Ceph `rbd sparsify` command), the CSI-driver should consider functions that the storage cluster supports safely.

>     * Is the volume mounted by some new pod temporarily on the same node? Or how will the volume be accessed?

In case the volume is mounted already, the NodePlugin of the CSI-driver on the node where the Pod with the volume is running can execute filesystem-local operation (like `fstrim`) or block-level operations (like `blkdiscard`).

>     * Are these operations something that could be done by a sidecar? Using an independent CRD seems like a very disconnected way to apply some operation on a volume. If there is a sidecar in the same pod, the application has more control.

The side-car need to have a lot of privileges to run `fstrim` or similar operations. As this is an extension to the CSI specification, the CSI-driver's NodePlugin and Controller (provisioner) seem appropriate components to execute these additional storage maintenance tasks.

api/v1beta1/csiaddonsjob_types.go
```diff
@@ -0,0 +1,117 @@
...
+// OperationState defines the state of the operation.
+type OperationState string
+
+const (
+	// Succeeded represents the Succeeded operation state.
+	Succeeded OperationState = "Succeeded"
+
+	// Failed represents the Failed operation state.
+	Failed OperationState = "Failed"
+)
```
BlaineEXE, line 62: Nit: I would suggest having these be `OperationStateSucceeded` and `OperationStateFailed`. The Kubernetes resource objects follow the naming convention of including type name with value.

api/v1beta1/csiaddonsjob_types.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// OperationSpec defines the desired operation of CSIAdoonsJobSpec.
+type OperationSpec struct {
+	// ReclaimSpace repesents ReclaimSpace operation.
+	ReclaimSpace ReclaimSpaceSpec `json:"reclaimSpace,omitempty"`
+}
+
+// OperationState defines the state of the operation.
+type OperationState string
```
BlaineEXE, line 54: Currently, this says operation "state" which definitionally means that it shows the ongoing status of a thing, however, the only values below are final result states. 

Should this be left as it is to support a "Processing" state or some other thing in the future?

Or would it be better to change this to `OperationResult` in order to indicate it should only represent the final result?

api/v1beta1/csiaddonsjob_types.go
```diff
@@ -0,0 +1,117 @@
...
+type CSIAddonsJobSpec struct {
+	// Operation represents the operation to be performed on the volume.
+	// Current Supported operation is "ReclaimSapce".
+	// +kubebuilder:validation:Required
+	Operation OperationSpec `json:"operation"`
+
+	// ActiveDeadlineSeconds specifies the duration in seconds relative to the
+	// startTime that the operation might be active; value must be positive integer.
+	// +kubebuilder:validation:Optional
+	ActiveDeadlineSeconds int64 `json:"activeDeadlineSeconds"`
```
BlaineEXE, line 74: What does it mean if this is zero? Does a default get used? The doc doesn't say.

api/v1beta1/csiaddonsjob_types.go
```diff
@@ -0,0 +1,117 @@
...
+
+	// BackOffLimit Specifies the number of retries before marking this operation failed.
+	// +kubebuilder:validation:Optional
+	BackoffLimit int32 `json:"backOffLimit"`
```
BlaineEXE, line 78: What does it mean if this is zero? Does a default get used? The doc doesn't say.

docs/csiaddonsjob.md
BlaineEXE: > That makes it more difficult to validate, and generated documentation could easily be incomplete. With a strongly typed CRD, it is very clear what operations are available, and what options they accept.

If the intent is to make strongly-typed CRDs, then why are we naming this with such a generic name: "CSIAddonsJob"?

IMO, if we want strongly-typed, then changing this to CSIReclaimSpaceAddonJob would be better, and we should create a new type of job for each operation. Having a generic job with strongly-typed internals is making two different design decisions in different ways for the same resource.

For me, I think `CSIAddonsJob` is a useful generic, and generics should not be strongly-typed. I consider this somewhat analogous to a StorageClass, which is also generic and takes all kinds of opaque and untyped data.

If we need immediate/good kubectl feedback about it, a validating webhook can be used.

api/v1beta1/csiaddonsjob_types.go
```diff
@@ -0,0 +1,117 @@
...
+	Succeeded OperationState = "Succeeded"
+
+	// Failed represents the Failed operation state.
+	Failed OperationState = "Failed"
+)
+
+// CSIAddonsJobSpec defines the desired state of CSIAddonsJob
+type CSIAddonsJobSpec struct {
+	// Operation represents the operation to be performed on the volume.
+	// Current Supported operation is "ReclaimSapce".
```
BlaineEXE, line 67: Why are we hard-coding support for add-ons into the `csi-addons` spec here? Is the intent for `csi-addons` not to be an extensible interface? The interface is no longer very extensible or independent once we start hard-coding add-on implementations into the add-on spec. I think this is anti-pattern to the intent of the CSI ecosystem and falls back into an easy but poorly-designed system.

BlaineEXE (changes requested):

api/v1beta1/csiaddonsjob_types.go
Rakshith-R: :+1:  I'll add defaults to `BackoffLimit` and `ActiveDeadlineSeconds` and document it.

docs/csiaddonsjob.md
Rakshith-R: >Having a generic job with strongly-typed internals is making two different design decisions in different ways for the same resource.

I agree with the above points mentioned by Blaine. 
IMO too we should either have a:
-  generic `CSIAddonsJob` (My preference)
- or Strongly typed Specific `CSIReclaimSpaceAddonJob` type(and add other jobs following similar pattern but with different type + controller)

@nixpanic wdyt?

api/v1beta1/csiaddonsjob_types.go
nixpanic: The different operations in the spec follow the same design as [Volumes in a Pod](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#volume-v1-core). Each Volume has a single `<type>VolumeSource` attribute, everything is strongly typed there.

Not only validation of the Spec is easier when it is strongly typed. An other important advantage is that users can easily see what is possible, without having to figure out what parameters can be used for a specific `string` formatted operation.

Using a CRD per operation would be my 2nd choice, but I wonder if that makes things less user-friendly (a scheduled version like `CronJob` is needed in addition to the `Job`). The number of CRDs might grow a lot, extending the singe CRD with a new operation might be easier for users?

BlaineEXE: I think this assumes that the design for Pod volumes is (1) the correct analog to use and (2) that the pod volume design doesn't have flaws we should consider. I have been in COSI design meetings with one of the people involved with creating the PVC and PV design, and he often expresses regret over some of the design decisions made there. It is good reminder to review whether prior art that exists in Kubernetes is actually "good" and whether a better design might meet needs better and allow for more flexibility.

In my estimation, this design would be more closely approximated the design of StorageClasses and not pod volumes. Even considering pod volumes as inspiration, I really don't think the pod volume API is above reproach. For the benefit that exists from strong typing, I also see a lot of the provider-specific stuff in that API as technical debt and not a feature. We are seeing the kubernetes community move away from provider-specific APIs with CSI in order to use the more generic CSI interface. 

I still think having a generic and non-prescriptive interface creates a better design because it does not create tight coupling between the CSI addons API and arbitrary implementations of csi addons. I think a design with looser coupling would be a more preferable one for pod volumes were kuberenetes ever to do a v2 API, and I am pretty certain that the storage sig would push to change that API in v2.

And also to be sure, the design of StorageClasses is not above reproach. String-string maps of arbitrary parameters can be a little challenging to use. Based on seeing the trends in Kubernetes of moving away from tightly coupled APIs (CSI for a large example), I would encourage us to have generic API interfaces that are not tied directly to specific implementations of CSI addons but that still have an amount of specificity that can help developers.

BlaineEXE: To be a little more transparent about some of the thoughts driving my considerations, I think that if we want the CSI addons API to be used and accepted by organizations outside of Red Hat, we would be better off having generic API components that follow the trend of StorageClasses more than the trend of Pod volumes. This will make it easier for outside organizations to consume the API without having to do work to push code changes to this API in addition to implementing their own addon.

If these addons are useful for people but we make the addons hard to develop and consume for those outside the Kubernetes ecosystem, it is more likely that a Kubernetes special interest group (SIG) will crop up to fill the gap, and we may have to pivot to match what the Kubernetes community is doing in the future. Or it's possible that CSI addons aren't interesting to other organizations, and we are the only ones who want to do this, and so it may not be a big issue. I truly don't know.

All that said, I think that organizations within Red Hat will still benefit from having a decoupled design, as it should mean addons can be created a little more quickly and without requiring as much back-and-forth between different engineering teams. But ultimately, Red Hat has to cooperate with Red Hat to meet corporate deliverables, and so that is not as big of a concern.

docs/reclaimspacejob.md
```diff
@@ -0,0 +1,21 @@
+# ReclaimSpaceJob
+
+ReclaimSpaceJob is a namespaced custom resource designed to invoke reclaim space operation on target volume.
+
+```yaml
+apiVersion: csiaddons.openshift.io/v1beta1
```
nixpanic, line 6: beta, or alpha?

nixpanic (commented):
Looks good to me, I do not see any issues with this.

You could extend the documentation a little, mentioning planned extensions like ReclaimSpaceCronJob and being able to set an annotation on the PVC so that the controller can create the ReclaimSpaceCronJob CR.

Maybe rename the doc to just `docs/reclaimspace.md` so that extensions to the job/operation are all listed together. Will be easier for users :+1:

api/v1alpha1/reclaimspacejob_types.go
```diff
@@ -0,0 +1,114 @@
...
+	// +optional
+	// +kubebuilder:default:=6
+	BackoffLimit int32 `json:"backOffLimit"`
+
+	// ActiveDeadlineSeconds specifies the duration in seconds relative to the
+	// start time that the operation might be retried; value must be positive integer.
+	// If not specified, defaults to 600 seconds.
+	// +optional
+	// +kubebuilder:default:=600
+	ActiveDeadlineSeconds int64 `json:"activeDeadlineSeconds"`
```
Madhu-1, line 60: i was discussing this with @Rakshith-R a couple of hours back. do we need this one? as we are not calling any RPC calls does this helps in any way

api/v1alpha1/reclaimspacejob_types.go
```diff
@@ -0,0 +1,114 @@
...
+	// Target represents volume target on which the operation will be
+	// performed.
+	// +kubebuilder:validation:Required
+	Target TargetSpec `json:"target"`
+
+	// BackOffLimit specifies the number of retries allowed before marking reclaim
+	// space operation as failed. If not specified, defaults to 6.
+	// +optional
+	// +kubebuilder:default:=6
+	BackoffLimit int32 `json:"backOffLimit"`
```
Madhu-1, line 53: Need to have maximum number validation for this one?

api/v1alpha1/reclaimspacejob_types.go
```diff
@@ -0,0 +1,114 @@
...
+	ReclaimedSpace resource.Quantity `json:"reclaimedSpace,omitempty"`
+
+	// Conditions are the list of conditions and their status.
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	// ObervedGeneration is the last generation change the controller has dealt with.
+	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+
+	// Retries indicates the number of times the operation is retried.
+	Retries        int64        `json:"retries,omitempty"`
```
Madhu-1, line 81: it can be uint8 or 32? what is the maximum retry we are allowing?

api/v1alpha1/reclaimspacejob_types.go
```diff
@@ -0,0 +1,114 @@
...
+	Result Result `json:"result,omitempty"`
+
+	// Message contains any message from the ReclaimSpaceJob.
+	Message string `json:"message,omitempty"`
+
+	// ReclaimedSpace indicates the amount of space reclaimed.
+	ReclaimedSpace resource.Quantity `json:"reclaimedSpace,omitempty"`
+
+	// Conditions are the list of conditions and their status.
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
```
Madhu-1, line 75: what are the conditions we are going to add as we already have result in the status?

Madhu-1 (commented):
just a few questions. LGTM

api/v1alpha1/reclaimspacejob_types.go
Rakshith-R: >what are the conditions we are going to add as we already have result in the status?

`status.result` will contain the final state `success` or `failure`, 

`status.conditions` will contain a single condition 
`type: success , status: true/false, reason/message:""` which will be used to indicate intermediate failures.

api/v1alpha1/reclaimspacejob_types.go
Rakshith-R: :+1:  added, we can adjust it later if needed.

api/v1alpha1/reclaimspacejob_types.go
Rakshith-R: changed it to `int32`, haven't seen any usage of unsigned int. Added validation should make sure it in required range. 

https://pkg.go.dev/k8s.io/api/batch/v1#JobSpec
I have set the `BackOffLimit` max to 60 right now, but we can adjust it later based on testing.

api/v1alpha1/reclaimspacejob_types.go
Madhu-1: reason can be added to the message?

api/v1alpha1/reclaimspacejob_types.go
Rakshith-R: > i was discussing this with @Rakshith-R a couple of hours back. do we need this one? as we are not calling any RPC calls does this helps in any way

Changed it to `RetryDeadlineSeconds`, IMO, it will be helpful to limit the retry of operation based on duration from the start of operation.

Similar to how a job is killed if it exceeds `ActiveDeadlineSeconds`
 https://pkg.go.dev/k8s.io/api/batch/v1#JobSpec

api/v1alpha1/reclaimspacejob_types.go
Rakshith-R: > reason can be added to the message?

I see that both are required for default metav1.Condition, 
https://pkg.go.dev/k8s.io/apimachinery@v0.21.3/pkg/apis/meta/v1#Condition

```
    // reason contains a programmatic identifier indicating the reason for the condition's last transition.
    // Producers of specific condition types may define expected values and meanings for this field,
    // and whether the values are considered a guaranteed API.
    // The value should be a CamelCase string.
    // This field may not be empty.
    // +required
    // +kubebuilder:validation:MaxLength=1024
    // +kubebuilder:validation:MinLength=1
    // +kubebuilder:validation:Pattern=`^[A-Za-z]([A-Za-z0-9_,:]*[A-Za-z0-9_])?$`
    Reason string `json:"reason" protobuf:"bytes,5,opt,name=reason"`
    // message is a human readable message indicating details about the transition.
    // This may be an empty string.
    // +required
    // +kubebuilder:validation:MaxLength=32768
    Message string `json:"message" protobuf:"bytes,6,opt,name=message"`
```
I think it will be useful to have it. 
We can use a custom condition or remove it altogether if we don't find it to be useful later on.

api/v1alpha1/reclaimspacejob_types.go
```diff
@@ -0,0 +1,117 @@
...
+
+package v1alpha1
+
+import (
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// OperationState defines the state of the operation.
+type Result string
```
yati1998, line 25: Also the name result doesn't specifies which result is it about. Maybe we can come up something meaningful. I guess you need to modify the comment first.

api/v1alpha1/reclaimspacejob_types.go
```diff
@@ -0,0 +1,117 @@
...
+*/
+
+package v1alpha1
+
+import (
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// OperationState defines the state of the operation.
```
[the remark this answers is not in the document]
Rakshith-R: Changed it to `OperationResult`

api/v1alpha1/reclaimspacejob_types.go
Madhu-1: but we are not going to kill the running fstrim/ operations here but we just stop the reconcile.

Rakshith-R: >but we are not going to kill the running fstrim/ operations here but we just stop the reconcile.

Yes, We will be stopping the retries here.

Madhu-1: am not sure this really helps. if it was killing/stopping the operation this makes sense.

Rakshith-R: > am not sure this really helps. if it was killing/stopping the operation this makes sense

It's just killing/stopping the next retry of the operation, both results in making sure the operation does not exceed time duration limit, one is less restrict than another. I believe we do need a way to limit the execution of operation to a particular approximate time period.

Madhu-1: am still not convinced for handing this deadline seconds @nixpanic what is your thought on this one? i don't have anything else except this one.

nixpanic: `ActiveDeadlineSeconds` sounds reasonable to me. I expect the operation to be in a pending/active state until it either returns a response or until this timeout expires. Upon expiration of the timeout, an error can be returned to the caller, and the retry logic can be applied. At this point the operation is not active anymore.

There is no need to _kill_ the running operation in this case, the deadline is most likely hit when the side-car or csi-driver faces an issue (restart, network troubles, ...).

Rakshith-R:
```

	// RetryDeadlineSeconds specifies the duration in seconds relative to the
	// start time that the operation may be retried; value MUST be positive integer.
	// If not specified, defaults to 600 seconds. Maximum allowed
	// value is 1800.
	// +optional
	// +kubebuilder:validation:Maximum=1800
	// +kubebuilder:validation:Minimum=0
	// +kubebuilder:default:=600
	RetryDeadlineSeconds int64 `json:"activeDeadlineSeconds"`
```

The latest iteration has `RetryDeadlineSeconds`, which is checked at the start of reconcile only. 

We'll not be using this to set timeouts at sidecar or driver level.

This is wholly to restrict operation to within a limited duration. Backofflimit only limits number of retries.

It is also designed to work in addition to `startDeadlineSeconds` of CronJob, this will make sure the operation will not be performed outside `StartDeadlineSeconds + RetryDeadlineSeconds`.
(Similar to k8s cronjob)

This will ensure reclaimspace io intensive operation does not slip outside maintenance period.

@Madhu-1 @nixpanic

nixpanic: > This will ensure reclaimspace io intensive operation does not slip outside maintenance period.

:point_up: is difficult to achieve. Once the operation is started, it is not really possible to cancel/kill it (except when killing the csi-driver pod doing the operation). My understanding is that this parameter is an _assumed to be inactive when no response is received after the deadline_ timer, just so that the retry logic can kick in.

`RetryDeadlineSeconds` or `ActiveDeadlineSeconds` both work for me, as long as I understand what the option should do.

nixpanic (approved):

api/v1alpha1/reclaimspacejob_types.go
Rakshith-R: We can discuss more during the implementation of the reclaimspacejob & reclaimspacecronjob controllers and make modifications accordingly to the API later too. 
The usage can be better understood from the controllers' implementation. 

This sounds okay to you @Madhu-1 ?

Madhu-1 (approved):

PR #6, opened 2021-10-29
api: adds CSIAddonsNode api

yati1998:
This commit adds CSIAddonsNode API.
Currently the API is designed such that we will have
single CSIAddonsNode per side-car.

Alternatively, we can have single CSIAddonsNode per host.

Related: csi-addons/spec#28

### DIFF
--- a/api/v1beta1/csiaddonsnode_types.go
+++ b/api/v1beta1/csiaddonsnode_types.go
@@ -1,0 +1,103 @@
+/*
+Copyright 2021 The Kubernetes-CSI-Addons Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1beta1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// CSIAddonsNodeState defines the state of the operation.
+type CSIAddonsNodeState string
+
+const (
+	// Connected represents the Connected state.
+	CSIAddonsNodeStateConnected CSIAddonsNodeState = "Connected"
+
+	// Failed represents the Connection Failed state.
+	CSIAddonsNodeStateFailed CSIAddonsNodeState = "Failed"
+)
+
+type CSIAddonsNodeDriver struct {
+	// Name is the name of the CSI driver that this object refers to.
+	// This must be the same name returned by the CSI-Addons GetIdentity()
+	// call for that driver. The name of the driver is in the format:
+	// `example.csi.ceph.com`
+	Name string `json:"name"`
+
+	// EndPoint is url that contains the ip-address to which the CSI-Addons
+	// side-car listens to.
+	EndPoint string `json:"endpoint"`
+
+	// NodeID is the ID of the node to identify on which node the side-car
+	// is running.
+	NodeID string `json:"nodeID"`
+}
+
+// CSIAddonsNodeSpec defines the desired state of CSIAddonsNode
+type CSIAddonsNodeSpec struct {
+	// Driver is the information of the CSI Driver existing on a node.
+	// If the driver is uninstalled, this can become empty.
+	Driver CSIAddonsNodeDriver `json:"driver"`
+}
+
+// CSIAddonsNodeStatus defines the observed state of CSIAddonsNode
+type CSIAddonsNodeStatus struct {
+	// State represents the state of the CSIAddonsNode object.
+	// It informs whether or not the CSIAddonsNode is Connected
+	// to the CSI Driver.
+	State CSIAddonsNodeState `json:"state,omitempty"`
+
+	// Messgae is a human-readable message indicating details about why the CSIAddonsNode
+	// is in this state.
+	// +optional
+	Message string `json:"message,omitempty"`
+
+	// Reason is a brief CamelCase string that describes any failure and is meant
+	// for machine parsing and tidy display in the CLI.
+	// +optional
+	Reason string `json:"reason,omitempty"`
+}
+
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:status
+//+kubebuilder:printcolumn:JSONPath=".metadata.namespace",name=namespace,type=string
+//+kubebuilder:printcolumn:JSONPath=".metadata.creationTimestamp",name=Age,type=date
+//+kubebuilder:printcolumn:JSONPath=".spec.driver.name",name=DriverName,type=string
+//+kubebuilder:printcolumn:JSONPath=".spec.driver.endpoint",name=Endpoint,type=string
+//+kubebuilder:printcolumn:JSONPath=".spec.driver.nodeID",name=NodeID,type=string
+
+// CSIAddonsNode is the Schema for the csiaddonsnode API
+type CSIAddonsNode struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   CSIAddonsNodeSpec   `json:"spec,omitempty"`
+	Status CSIAddonsNodeStatus `json:"status,omitempty"`
+}
+
+// +kubebuilder:object:root=true
+
+// CSIAddonsNodeList contains a list of CSIAddonsNode
+type CSIAddonsNodeList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []CSIAddonsNode `json:"items"`
+}
+
+func init() {
+	SchemeBuilder.Register(&CSIAddonsNode{}, &CSIAddonsNodeList{})
+}
--- a/docs/csiaddonsnode.md
+++ b/docs/csiaddonsnode.md
@@ -1,0 +1,19 @@
+# CSIAddonsNode
+
+CSIAddonsNode is a custom resource designed to allow the CSI-Addons Controller(s) to discover CSI-Addons side-cars that are running alongside of CSI-driver components. The information provided in a CSIAddonsNode CR contain details on how to connect to the CSI-Addons side-car.
+
+```yaml
+apiVersion: csiaddons.openshift.io/v1beta1
+kind: CSIAddonsNode
+metadata:
+  name: csiaddonsnode-sample
+spec:
+    driver:
+        name: example.csi.ceph.com
+        endpoint: <url>
+        nodeID: node-1
+```
++ `driver` contains the required information about the CSI driver.
+    + `name` contains the name of the driver. The name of the driver is in the format: `rbd/cephfs.csi.ceph.com`
+    + `endpoint` contains the url that contains the ip-address to which the CSI-Addons side-car listens to.
+    + `nodeID` contains the ID of node to identify on which node the side-car is running.


api/v1beta1/csiaddonsnode_types.go
```diff
@@ -0,0 +1,109 @@
...
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+type CSIAddonsNodeDriver struct {
+	// This is the name of the CSI driver that this object refers to.
+	// This must be the same name returned by the CSI GetPluginame() call for
+	// that driver.
+	Name string `json:"name" protobuf:"bytes,1,opt,name=name"`
+
+	// Type represents the type of the driver i.e., rbd/cephfs.
+	Type string
```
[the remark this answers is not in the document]
yati1998: @nixpanic Can you please elaborate? It is present only in the CSIAddonsNodeDriver struct. And I think it will be good to have the `Type` to mention if it's the rbd/cephfs.  Not sure what exactly you meant here.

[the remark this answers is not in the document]
nixpanic: consumers of the `CSIAddonsNode` do not need to know anything about the storage system. They would use the `drivername` to match `provisioner` in the StorageClass and get details to connect to the storage system that way.

A `Type` might be interesting for users (people) viewing the CR, but they would also see the `drivername`. I expect that users viewing the CR recognize the drivernames and know what storage system is used with it.

api/v1beta1/csiaddonsnode_types.go
```diff
@@ -0,0 +1,101 @@
...
+	Failed CSIAddonsNodeState = "Failed"
+)
+
+// CSIAddonsNodeSpec defines the desired state of CSIAddonsNode
+type CSIAddonsNodeSpec struct {
+	// Driver is the information of the CSI Driver existing on a node.
+	// If the driver is uninstalled, this can become empty.
+	// +patchMergeKey=name
+	// +patchStrategy=merge
+	Driver CSIAddonsNodeDriver `json:"driver" patchStrategy:"merge" patchMergeKey:"name" protobuf:"bytes,1,rep,name=drivers"`
```
nixpanic, line 49: I think part of the `patchStrategy` and `patchMergeKey` options are related to a `drivers` list, but this currently is a single driver instance.

api/v1beta1/csiaddonsnode_types.go
```diff
@@ -0,0 +1,101 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type CSIAddonsNodeDriver struct {
+	// Name is the name of the CSI driver that this object refers to.
+	// This must be the same name returned by the CSI-Addons GetIdentity()
+	// call for that driver. The name of the driver is in the format:
+	// `rbd/cephfs.csi.ceph.com`
+	Name string `json:"name" protobuf:"bytes,1,opt,name=name"`
```
nixpanic, line 25: not sure the `protobuf` annotations are needed?

api/v1beta1/csiaddonsnode_types.go
```diff
@@ -0,0 +1,101 @@
...
+
+	// EndPoint is url that contains the ip-address to which the CSI-Addons
+	// side-car listens to.
+	EndPoint string
```
nixpanic, line 29: this misses json annotations

api/v1beta1/csiaddonsnode_types.go
```diff
@@ -0,0 +1,101 @@
...
+	// +patchStrategy=merge
+	Driver CSIAddonsNodeDriver `json:"driver" patchStrategy:"merge" patchMergeKey:"name" protobuf:"bytes,1,rep,name=drivers"`
+}
+
+// CSIAddonsNodeStatus defines the observed state of CSIAddonsNode
+type CSIAddonsNodeStatus struct {
+	// State represents the state of the CSIAddonsNode object.
+	// It informs whether or not the CSIAddonsNode is Connected
+	// to the CSI Driver.
+	State CSIAddonsNodeState `json:"status,omitempty"`
```
nixpanic, line 57: attribute is `State`, in json it is `status`, is that commonly done like that?

docs/csiaddonsnode.md
```diff
@@ -0,0 +1,17 @@
...
+
+```yaml
+apiVersion: csiaddons.openshift.io/v1beta1
+kind: CSIAddonsNode
+metadata:
+  name: csiaddonsnode-sample
+spec:
+    driver:
+        name: rbd/cephfs.csi.ceph.com
+        endpoint: <url>
```
nixpanic, line 13: is `endpoint` all lower case?

docs/csiaddonsnode.md
```diff
@@ -0,0 +1,17 @@
...
+CSIAddonsNode is a custom resource designed to allow the CSI-Addons Controller(s) to discover CSI-Addons side-cars that are running alongside of CSI-driver components. The information provided in a CSIAddonsNode CR contain details on how to connect to the CSI-Addons side-car.
+
+```yaml
+apiVersion: csiaddons.openshift.io/v1beta1
+kind: CSIAddonsNode
+metadata:
+  name: csiaddonsnode-sample
+spec:
+    driver:
+        name: rbd/cephfs.csi.ceph.com
```
Rakshith-R, line 12:
```suggestion
  name: csiaddonsnode-sample
spec:
    driver:
        name: rbd/cephfs.csi.ceph.com
        nodeID: node-1
```
Please add nodeID(optional) to identify on which node the side-car is running on.

nixpanic: Indeed, we still need to know the `nodeID` so that we can identify which node has a volume attached. The endpoint is not sufficient for that.

Because this is not a driver-list, but a single driver, the `nodeID` probably fits in the `CSIAddonsNodeSpec` directly.

api/v1beta1/csiaddonsnode_types.go
```diff
@@ -0,0 +1,101 @@
+/*
+Copyright 2021.
```
Madhu-1, line 2: don't we have author for this one like https://github.com/csi-addons/kubernetes-csi-addons/pull/4/files#diff-ca1a7e8d22d8c8a4d2ef31d0898a86dd1fb989c820d3a7e88883f6d0e2d96cdbR2

api/v1beta1/csiaddonsnode_types.go
```diff
@@ -0,0 +1,101 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type CSIAddonsNodeDriver struct {
+	// Name is the name of the CSI driver that this object refers to.
+	// This must be the same name returned by the CSI-Addons GetIdentity()
+	// call for that driver. The name of the driver is in the format:
+	// `rbd/cephfs.csi.ceph.com`
```
Madhu-1, line 24: no need to have rbd/cephfs specific examples here keep `example.csidriver.com`

api/v1beta1/csiaddonsnode_types.go
Madhu-1: only JSON is needed.

api/v1beta1/csiaddonsnode_types.go
```diff
@@ -0,0 +1,101 @@
...
+	Name string `json:"name" protobuf:"bytes,1,opt,name=name"`
+
+	// EndPoint is url that contains the ip-address to which the CSI-Addons
```
Madhu-1, line 27: it's not always the IP address isn't it?

api/v1beta1/csiaddonsnode_types.go
```diff
@@ -0,0 +1,101 @@
...
+// CSIAddonsNodeState defines the state of the operation.
+type CSIAddonsNodeState string
+
+const (
+	// Connected represents the Connected state.
+	Connected CSIAddonsNodeState = "Connected"
+
+	// Failed represents the Connection Failed state.
+	Failed CSIAddonsNodeState = "Failed"
+)
```
Madhu-1, line 41: move this to beginning of the file.

api/v1beta1/csiaddonsnode_types.go
Madhu-1: do we need to specify the patch strategy,mergekey, etc here?

api/v1beta1/csiaddonsnode_types.go
Madhu-1, line 29: call this as the address instead of endpoint?

api/v1beta1/csiaddonsnode_types.go
```diff
@@ -0,0 +1,101 @@
...
+	// +optional
+	Reason string `json:"reason,omitempty" protobuf:"bytes,3,opt,name=reason"`
+}
+
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:status
+//+kubebuilder:printcolumn:JSONPath=".metadata.namespace",name=namespace,type=string
+//+kubebuilder:printcolumn:JSONPath=".metadata.creationTimestamp",name=Age,type=date
+//+kubebuilder:printcolumn:JSONPath=".spec.driver.name",name=DriverName,type=string
+//+kubebuilder:printcolumn:JSONPath=".spec.driver.type",name=DriverType,type=string
```
Madhu-1, line 75: what is driver type?

api/v1beta1/csiaddonsnode_types.go
```diff
@@ -0,0 +1,101 @@
...
+//+kubebuilder:printcolumn:JSONPath=".spec.driver.endpoint",name=Endpoint,type=string
+//+kubebuilder:printcolumn:JSONPath=".status.state",name=state,type=string
+//+kubebuilder:printcolumn:JSONPath=".status.message",name=message,type=string
+//+kubebuilder:printcolumn:JSONPath=".status.reason",name=reason,type=string
```
Madhu-1, line 79: the status should not be printed when someone does `oc get object` it's more for describe/yaml output

docs/csiaddonsnode.md
Madhu-1, line 12: have the example of generic name not rbd/cephfs specific.

api/v1beta1/csiaddonsnode_types.go
yati1998: Missed to edit. Will update.

docs/csiaddonsnode.md
yati1998: yes.

api/v1beta1/csiaddonsnode_types.go
yati1998: will add it.

api/v1beta1/csiaddonsnode_types.go
yati1998: Not sure, but AFAIK, it's the IP and the port that defines the address to which CSIAddons sidecar will listen, something like this: `tcp://10.0.123.67:1234`. Am I missing anything here @nixpanic .

api/v1beta1/csiaddonsnode_types.go
yati1998: I don't mind, but it will mostly be a URL containing the IP address and port and not just the address!!

api/v1beta1/csiaddonsnode_types.go
yati1998: Forgot to remove it, planed to have driver type(rbd/cephfs) but since it is clear from the name, it's not required.

docs/csiaddonsnode.md
yati1998: @nixpanic  IMO, nodeID should be the part of `CSIAddonsNodeDriver` ad the description of it says that it contains all the information of the CSI Driver existing on the node, which will include nodeID too. Let me know if you still want that to be added in CSIAddonsNodeSpec directly.

api/v1beta1/csiaddonsnode_types.go
```diff
@@ -0,0 +1,104 @@
...
+
+// CSIAddonsNodeState defines the state of the operation.
+type CSIAddonsNodeState string
+
+const (
+	// Connected represents the Connected state.
+	Connected CSIAddonsNodeState = "Connected"
+
+	// Failed represents the Connection Failed state.
+	Failed CSIAddonsNodeState = "Failed"
```
nixpanic, line 31: See https://github.com/csi-addons/kubernetes-csi-addons/pull/4#discussion_r750553245 for a comment on standardized naming

nixpanic (dismissed):

api/v1beta1/csiaddonsnode_types.go
```diff
@@ -0,0 +1,104 @@
...
+	// `example.csi.ceph.com`
+	Name string `json:"name"`
+
+	// EndPoint is url that contains the ip-address to which the CSI-Addons
+	// side-car listens to.
+	EndPoint string `json:"endpoint"`
+
+	// NodeID is the ID of the node to identify on which node the side-car
+	// is running.
+	// +optional
```
Madhu-1, line 47: @yati1998 @nixpanic if this is optional how the controller can reach the driver on the particular node?

nixpanic (approved):

Madhu-1 (approved):

PR #10, opened 2021-12-14
api: add NetworkFence api

Yuggupta27:
# Network Fence Custom Resource 

NetworkFence is a cluster-scoped custom resource that allows Kubernetes to invoke "Network fence" operation on a storage provider.

### Fence Operation

```yaml
apiVersion: csiaddons.openshift.io/v1alpha1
kind: NetworkFence
metadata:
  name: network-fence-sample
spec:
  driver: example.driver
  cidr: 10.90.89.66/32
  secret:
    name: fence-secret
    namespace: default
  parameters:
    key: value

```

> **Note**: Creation of a NetworkFence CR blocks access to the corresponding CIDR block; which is then unblocked the CR deletion.

```bash
[yuggupta@fedora kubernetes-csi-addons](fencing)$ kubectl get networkfence
NAME                   DRIVER              CIDR             AGE    RESULT
network-fence-sample   example.driver      10.90.89.66/32   121m   
```


docs/networkfence.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Fence Operation
+
+```yaml
+apiVersion: csiaddons.openshift.io/v1alpha1
+kind: NetworkFence
+metadata:
+  name: network-fence-sample
+spec:
+  operation: fence
```
Madhu-1, line 16: how you will send a request to a specific CSI driver what is the identifier here? suppose 2 drivers are registered for fencing how a user can specify to which CSI driver he needs to invoke for fencing?

docs/networkfence.md
```diff
@@ -0,0 +1,47 @@
...
+  cidr:
```
Madhu-1, line 17: don't we need secrets to talk to the CSI driver? how it will be handled?

api/v1alpha1/networkfence_types.go
```diff
@@ -0,0 +1,110 @@
+/*
+Copyright 2021.
```
Madhu-1, line 2: the authors is missing in the license?

Madhu-1 (changes requested):

api/v1alpha1/networkfence_types.go
nixpanic: Still todo?

api/v1alpha1/networkfence_types.go
```diff
@@ -0,0 +1,110 @@
...
+	// OperationResultFailed represents the Failed operation state.
+	FencingOperationResultFailed FencingOperationResult = "Failed"
+)
+
+// NetworkFenceSpec defines the desired state of NetworkFence
+type NetworkFenceSpec struct {
+	// Operation specifes whether fence or unfence operation is to be performed
+	// on the corresponding CIDR blocks.
+	// It's can be set to either "fence" or "unfence".
+	Operation string `json:"operation,omitempty"`
```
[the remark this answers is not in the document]
nixpanic: This would also be something I would expect to happen. Create an instance of the `NetworkFence` CR to execute the underlaying fence operation. When the CR is deleted the unfence operation would be executed.

Can the CIDR list be modified after the initial creation of the CR? I guess that would be practical, so that a single CSI-node can be fenced/unfenced by editing the CR, while keeping the state of others in tact.

api/v1alpha1/networkfence_types.go
```diff
@@ -0,0 +1,110 @@
...
+
+	// Retries indicates the number of times the operation is retried.
+	Retries int32 `json:"retries,omitempty"`
+}
+
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:status
+
+//+kubebuilder:printcolumn:name="Operation",type="string",JSONPath=".spec.operation"
+//+kubebuilder:printcolumn:name="Clients",type="string",JSONPath=".spec.cidr"
```
nixpanic, line 84: `Clients` is probably not the right name?

docs/networkfence.md
nixpanic: Indeed, would it use the provisioner or node-plugin Secrets linked from a StorageClass, or maybe from a PV?

I guess it is possible to link to some secret, but that makes it not very user friendly.

This should be a privileged operation too. What is the plan to not allow anyone with permissions to the Kubernetes cluster to fence random CIDRs?

nixpanic (changes requested):

docs/networkfence.md
Yuggupta27: Thanks for pointing that out. Updated now.

api/v1alpha1/networkfence_types.go
Yuggupta27: Updated now.

api/v1alpha1/networkfence_types.go
Yuggupta27: Updated it to CIDR now

Yuggupta27:
Thanks for the suggestions @Madhu-1 @nixpanic, Updated the changes as well as the PR description.

docs/networkfence.md
Yuggupta27: > This should be a privileged operation too. What is the plan to not allow anyone with permissions to the Kubernetes cluster to fence random CIDRs?


Added secrets as a part of the CR now. Since blocking access is an admin operation, the `NetworkFence` CR creation will be done by high access users(admin most probably) and not by a common user.  The secrets used, thus, will have privileged access.

api/v1alpha1/networkfence_types.go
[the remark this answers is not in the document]
Yuggupta27: Having an update operation for the CR may cause issues during the unfence operation; as we perform unfence on a CIDR when the `NetworkFence` CR corresponding to it is deleted.
1. Assuming we have a CR with CIDR `10.23.53.42/24` , this will result in network fencing of `10.23.53.42/24`
2. Now, if an update is done on the CR, to say `10.15.35.64/24`, this will now fence `10.15.35.64/24`.
3. On deletion of the CR now, only `10.15.35.64/24` will be unfenced and the first CIDR(`10.23.53.42/24` ) will still remain fenced.

wdyt? @nixpanic @Madhu-1

api/v1alpha1/networkfence_types.go
```diff
@@ -0,0 +1,99 @@
...
+	// Namespace specifies the namespace in which the secret
+	// is located.
+	Namespace string `json:"namespace,omitempty"`
+}
+
+// NetworkFenceSpec defines the desired state of NetworkFence
+type NetworkFenceSpec struct {
+	// Provisioner is the name of storage provisioner.
+	// +kubebuilder:validation:Required
+	Provisioner string `json:"provisioner"`
```
nixpanic, line 47: Maybe use StorageClass instead of provisioner?

api/v1alpha1/networkfence_types.go
```diff
@@ -0,0 +1,99 @@
...
+	Name string `json:"name,omitempty"`
+
+	// Namespace specifies the namespace in which the secret
+	// is located.
+	Namespace string `json:"namespace,omitempty"`
+}
+
+// NetworkFenceSpec defines the desired state of NetworkFence
+type NetworkFenceSpec struct {
+	// Provisioner is the name of storage provisioner.
```
Rakshith-R, line 45: Let this be `driverName` instead of `provisionerName`?
And we would need `cluterID` too ?

api/v1alpha1/networkfence_types.go
```diff
@@ -0,0 +1,99 @@
...
+	// +kubebuilder:validation:Required
+	Provisioner string `json:"provisioner"`
+
+	// Cidr contains a CIDR block, on which the corresponding fence/unfence
```
Rakshith-R, line 49:
```suggestion
	// Cidr contains a CIDR block which is to fenced.
```

api/v1alpha1/networkfence_types.go
```diff
@@ -0,0 +1,99 @@
...
+	// operation are to be performed.
+	Cidr string `json:"cidr,omitempty"`
```
Rakshith-R, line 51: Make it a required parameter

Rakshith-R (changes requested):

api/v1alpha1/networkfence_types.go
```diff
@@ -0,0 +1,102 @@
...
+)
+
+// SecretSpec defines the secrets to be used for the network fencing operation.
+type SecretSpec struct {
+	// Name specifies the name of the secret.
+	Name string `json:"name,omitempty"`
+
+	// Namespace specifies the namespace in which the secret
+	// is located.
+	Namespace string `json:"namespace,omitempty"`
```
Rakshith-R, line 40:
```suggestion
	// Name specifies the name of the secret.
	// +kubebuilder:validation:Required
	Name string `json:"name,omitempty"`

	// Namespace specifies the namespace in which the secret
	// is located.
	// +kubebuilder:validation:Required
	Namespace string `json:"namespace,omitempty"`
```

api/v1alpha1/networkfence_types.go
Yuggupta27: > And we would need clusterID too ?

Yes, that will be part of additional parameters for SP

api/v1alpha1/networkfence_types.go
Yuggupta27: Using driver name for identifying the corresponding CSI driver now :+1:

Rakshith-R (changes requested):

api/v1alpha1/networkfence_types.go
```diff
@@ -0,0 +1,102 @@
...
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type FencingOperationResult string
+
+const (
+	// OperationResultSucceeded represents the Succeeded operation state.
```
yati1998, line 26: OperationResultSucceeded -> FencingOperationResultSucceeded

api/v1alpha1/networkfence_types.go
```diff
@@ -0,0 +1,102 @@
...
+	FencingOperationResultSucceeded FencingOperationResult = "Succeeded"
+
+	// OperationResultFailed represents the Failed operation state.
```
yati1998, line 29: OperationResultFailed  -> FencingOperationResultFailed

yati1998 (commented):
LGTM, preserving merge rights for others.

api/v1alpha1/networkfence_types.go
```diff
@@ -0,0 +1,104 @@
...
+	// +kubebuilder:validation:Required
+	Driver string `json:"driver"`
+
+	// Cidr contains a CIDR block, which is required to be fenced.
+	// +kubebuilder:validation:Required
+	Cidr string `json:"cidr"`
+
+	// Secret is a kubernetes secret, which is required to perform the fence/unfence operation.
+	// +kubebuilder:validation:Required
+	Secret SecretSpec `json:"secret"`
```
Madhu-1, line 57: this is not mandatory and its optional.

api/v1alpha1/networkfence_types.go
```diff
@@ -0,0 +1,104 @@
...
+// SecretSpec defines the secrets to be used for the network fencing operation.
+type SecretSpec struct {
+	// Name specifies the name of the secret.
+	// +kubebuilder:validation:Required
+	Name string `json:"name"`
+
+	// Namespace specifies the namespace in which the secret
+	// is located.
+	// +kubebuilder:validation:Required
+	Namespace string `json:"namespace"`
```
Madhu-1, line 42: this is not mandatory and its optional.

Yuggupta27: Thanks! Updated now

api/v1alpha1/networkfence_types.go
yati1998: @Yuggupta27 seems you have missed this?

Madhu-1 (dismissed):

Madhu-1:
@Yuggupta27 please fix CI

Madhu-1:
@nixpanic @Rakshith-R PTAL. there is a requested change from your side.

api/v1alpha1/networkfence_types.go
```diff
@@ -0,0 +1,103 @@
...
+
+// NetworkFenceSpec defines the desired state of NetworkFence
+type NetworkFenceSpec struct {
+	// Driver contains  the name of CSI driver.
+	// +kubebuilder:validation:Required
+	Driver string `json:"driver"`
+
+	// Cidr contains a CIDR block, which is required to be fenced.
+	// +kubebuilder:validation:Required
+	Cidr string `json:"cidr"`
```
nixpanic, line 51: do you want this as a single `CIDR`, or a list?

docs/networkfence.md
```diff
@@ -0,0 +1,32 @@
+# NetworkFence
+
+NetworkFence is a cluster-scoped custom resource that allows Kubernetes to invoke "Network fence" operation on a storage provider.
+
+The user needs to specify the kind of operation they want to perform (fence/unfence), and then provide the list of CIDR blocks on which the mentioned operation is to be performed.
```
nixpanic, line 5: here is `list of CIDR blocks`, but it is not a list in the example.

nixpanic (changes requested):

docs/networkfence.md
Yuggupta27: Nice catch! The description was old and was not updated. Updated now :)

api/v1alpha1/networkfence_types.go
Yuggupta27: In case multiple CIDR blocks are required to be network fenced, one Custom Resource per CIDR block is required so as to keep the operational complexity low.
As each CIDR block then has its own range of IPs; due to which keeping multiple CIDRs in one CR will highly increase nested complexity. 

Although, in the future, adding multiple CIDRs to a CR is definitely a goal.

nixpanic: _to keep the operational complexity low_ is not a good reason to make the interface less user-friendly. As the CSI-Addons specification offers a list of CIDRs, it probably is better to follow that approach.

Yuggupta27: Thanks for the suggestion! Going with list of blocks :+1:  Updated now!

api/v1alpha1/networkfence_types.go
```diff
@@ -0,0 +1,103 @@
...
+
+// NetworkFenceSpec defines the desired state of NetworkFence
+type NetworkFenceSpec struct {
+	// Driver contains  the name of CSI driver.
+	// +kubebuilder:validation:Required
+	Driver string `json:"driver"`
+
+	// Cidr contains a list of CIDR blocks, which are required to be fenced.
+	// +kubebuilder:validation:Required
+	Cidr []string `json:"cidr"`
```
nixpanic, line 51: now that this is a list :partying_face: the name needs to be `cidrs` (with `s`)

docs/networkfence.md
```diff
@@ -0,0 +1,34 @@
...
+    name: fence-secret
+    namespace: default
+  parameters:
+    key: value
+```
+
+> **Note**: Creation of a NetworkFence CR blocks access to the corresponding CIDR block; which is then unblocked the CR deletion.
+
++ `provisioner`: specifies the name of storage provisioner.
++ `cidr`: refers to the CIDR block on which the mentioned fence/unfence operation is to be performed.
```
nixpanic, line 30: change to `cidrs` and `CIDR blocks`

nixpanic (changes requested):
few minor changes needed for `cidr` -> `cidrs`. Looks good otherwise

docs/networkfence.md
Yuggupta27: sure, updating

Yuggupta27: Updated now

api/v1alpha1/networkfence_types.go
Yuggupta27: Thanks! Updated now

nixpanic (approved):

Madhu-1 (approved):

humblec (approved):

PR #20, opened 2021-12-17
changes the conflicting names

yati1998:
This commit changes `StorageConsumption` to
`StorageUsage` as this was conflicting with other proto file.
This commit also changes Node/ControllerReclaimSpaceRequest
to single request as the structure was the same, resulting in error.


nixpanic (changes requested):
I do not think this is the right approach. It probably is possible to change the package name (now `reclaimspace`) to something else that is not used in the csi-addons spec repository. Maybe just `proto` and place other (future) `.proto` files in the same directory/package.

internal/service/reclaimspace/reclaimspace.proto
```diff
@@ -8,16 +8,16 @@ option go_package = "github.com/csi-addons/kubernetes-csi-addons/internal/servic
 service ReclaimSpace {
     // ControllerReclaimSpace is a procedure that gets called on the CSI
     // sidecar.
-    rpc ControllerReclaimSpace (ControllerReclaimSpaceRequest) returns(ReclaimSpaceResponse) {}
+    rpc ControllerReclaimSpace (ReclaimSpaceRequest) returns(ReclaimSpaceResponse) {}
     
     // NodeReclaimSpace is a procedure that gets called on the CSI
     // sidecar.
-    rpc NodeReclaimSpace (NodeReclaimSpaceRequest) returns(ReclaimSpaceResponse) {}
+    rpc NodeReclaimSpace (ReclaimSpaceRequest) returns(ReclaimSpaceResponse) {}
```
Rakshith-R, line 15: Changing package name and regenerating the files seems to work. 
@yati1998 can you modify package name instead please?

@nixpanic @yati1998  I am also inclined to keep a single `RecaimSpaceRequest` similar to `ReclaimSpaceResponse` since they share the same parameter.

`StorageUsage` can be reverted back to `StorageConsumption`

Rakshith-R (dismissed):

yati1998:
@nixpanic @Rakshith-R made the changes for now but I thinking for some meaningful name instead of simply proto.

internal/service/reclaimspace/reclaimspace.proto
nixpanic: There needs to be a way for the side-car to make a distinction between NodeReclaimSpace and ControllerReclaimSpace. Using different procedures makes that the cleanest.

Rakshith-R: Yes, I referred to the `Request` struct, holding pv_name.

yati1998: yes, the rpc's names are different only are arguments are same. In my opinion, it will be fine, isn't it???

nixpanic:
> @nixpanic @Rakshith-R made the changes for now but I am still thinking for some meaningful name instead of simply proto.

I think `proto` is fine for now. That package should only contain the `.proto` files and generated `.pb.go` files. The directory structure could then look like`

```
/
  +-- internal/
  |    +-- proto/
  |           +-- reclaimspace.proto
  |           +-- reclaimspace.pb.go
  |           +-- reclaimspace_grpc.pb.go
  |
  +-- controller/internal/
  |     +-- reclaimspace/
  |           +-- reclaimspace.go
  |           +-- reclaimspace_test.go
  |
  +-- sidecar/internal/
        +-- reclaimspace/
              +-- reclaimspace.go
              +-- reclaimspace_test.go
```

internal/service/proto/proto.proto
```diff
@@ -1,23 +1,23 @@
 syntax = "proto3";
-package reclaimspace;
+package proto;
 
-option go_package = "github.com/csi-addons/kubernetes-csi-addons/internal/service/reclaimspace";
+option go_package = "github.com/csi-addons/kubernetes-csi-addons/internal/service/proto";
```
nixpanic, line 4: the filename should stay `reclaimspace.proto`. The package can just be `proto`. When fencing is added, the `fence.proto` file (and generated `.go` files) should then be included in the same `proto` package.

I am not sure if the `service` part in the directory structure will be beneficial. Maybe it is cleaner to drop that?

yati1998:
> > @nixpanic @Rakshith-R made the changes for now but I am still thinking for some meaningful name instead of simply proto.
> 
> I think `proto` is fine for now. That package should only contain the `.proto` files and generated `.pb.go` files. The directory structure could then look like`
> 
> ```
> /
>   +-- internal/
>   |    +-- proto/
>   |           +-- reclaimspace.proto
>   |           +-- reclaimspace.pb.go
>   |           +-- reclaimspace_grpc.pb.go
>   |
>   +-- controller/internal/
>   |     +-- reclaimspace/
>   |           +-- reclaimspace.go
>   |           +-- reclaimspace_test.go
>   |
>   +-- sidecar/internal/
>         +-- reclaimspace/
>               +-- reclaimspace.go
>               +-- reclaimspace_test.go
> ```

can't have reclaimspace.proto as it conflicts with the one in the spec field, either we need to change the `StorageConsumption`  or the filename. Orelse there will be error while building it.

nixpanic:
I am not sure what you mean by this:

> can't have reclaimspace.proto as it conflicts with the one in the spec field, either we need to change the `StorageConsumption` or the filename. Orelse there will be error while building it.

What conflict is that? Can you share the error message?

If the packages for the files are different, it is (at least to me) surprising that there are conflicts.

internal/proto/reclaimspace.proto
```diff
@@ -1,23 +1,23 @@
 syntax = "proto3";
-package reclaimspace;
+package proto;
 
-option go_package = "github.com/csi-addons/kubernetes-csi-addons/internal/service/reclaimspace";
+option go_package = "github.com/csi-addons/kubernetes-csi-addons/internal/service/proto";
```
nixpanic, line 4: I think `service` can be removed from this line too?

nixpanic (dismissed):

Madhu-1 (approved):

nixpanic (approved):

PR #21, opened 2021-12-21
ci: add GitHub Workflow for building and pushing container-images

nixpanic:
Two CI jobs are added:

1. Test building container images, so that Mergify can require 
  `build_controller` and `build_sidecar` as CI jobs.

2. The container images should get pushed to quay.io after a PR has been
  merged. This currently builds the container images for all the
  architectures that the golang base container and Docker Actions support.

### DIFF
--- a/.github/workflows/build-push.yaml
+++ b/.github/workflows/build-push.yaml
@@ -1,0 +1,68 @@
+---
+name: Publish container images in quay.io
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  push_controller:
+    name: Push controller container image to quay.io
+    if: github.repository == 'csi-addons/kubernetes-csi-addons'
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


Madhu-1 (commented):
there is a lot of duplication of code this can be reduced take a look at https://github.blog/changelog/2021-08-25-github-actions-reduce-duplication-with-action-composition/ could be considered as an enhancement if it needs more time to figure out.

nixpanic:
> there is a lot of duplication of code this can be reduced take a look at https://github.blog/changelog/2021-08-25-github-actions-reduce-duplication-with-action-composition/ could be considered as an enhancement if it needs more time to figure out.

That is a good idea, but it would need some experimenting before I feel comfortable doing it that way. Created #24 for tracking.

Madhu-1 (approved):

yati1998 (approved):

PR #32, opened 2021-12-22
Implementation of Reclaim Space operation

yati1998:
This commit adds the implementation of Reclaim Space
Operation.

### DIFF
--- a/sidecar/internal/service/reclaimspace.go
+++ b/sidecar/internal/service/reclaimspace.go
@@ -17,12 +17,18 @@ package service
 
 import (
 	"context"
+	"path/filepath"
 
 	"github.com/csi-addons/kubernetes-csi-addons/internal/proto"
+	kube "github.com/csi-addons/kubernetes-csi-addons/sidecar/internal/kubernetes"
 	csiReclaimSpace "github.com/csi-addons/spec/lib/go/reclaimspace"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
+	"github.com/kubernetes-csi/csi-lib-utils/accessmodes"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
 )
@@ -35,15 +41,17 @@ type ReclaimSpaceServer struct {
 	controllerClient csiReclaimSpace.ReclaimSpaceControllerClient
 	nodeClient       csiReclaimSpace.ReclaimSpaceNodeClient
 	kubeClient       *kubernetes.Clientset
+	stagingPath      string
 }
 
 // NewReclaimSpaceServer creates a new ReclaimSpaceServer which handles the proto.ReclaimSpace
 // Service requests.
-func NewReclaimSpaceServer(c *grpc.ClientConn, kc *kubernetes.Clientset) *ReclaimSpaceServer {
+func NewReclaimSpaceServer(c *grpc.ClientConn, kc *kubernetes.Clientset, sp string) *ReclaimSpaceServer {
 	return &ReclaimSpaceServer{
 		controllerClient: csiReclaimSpace.NewReclaimSpaceControllerClient(c),
 		nodeClient:       csiReclaimSpace.NewReclaimSpaceNodeClient(c),
 		kubeClient:       kc,
+		stagingPath:      sp,
 	}
 }
 
@@ -61,22 +69,50 @@ func (rs *ReclaimSpaceServer) ControllerReclaimSpace(
 	pvName := req.GetPvName()
 	klog.Info(pvName)
 
-	// TODO: get following params using kubeClient and pvName
+	pv, err := rs.kubeClient.CoreV1().PersistentVolumes().Get(ctx, pvName, metav1.GetOptions{})
+	if err != nil {
+		klog.Errorf("Failed to get pv: %v", err)
+		return nil, status.Errorf(codes.InvalidArgument, "failed to get pv %q", pvName)
+	}
+
+	if pv.Spec.CSI == nil {
+		return nil, status.Errorf(codes.InvalidArgument, "pv %q is not a CSI volume", pvName)
+	}
+
+	volID := pv.Spec.CSI.VolumeHandle
+	volAttributes := pv.Spec.CSI.VolumeAttributes
 
 	csiReq := &csiReclaimSpace.ControllerReclaimSpaceRequest{
-		VolumeId:   "",
-		Parameters: map[string]string{},
-		Secrets:    map[string]string{},
+		VolumeId:   volID,
+		Parameters: volAttributes,
 	}
-	res, err := rs.controllerClient.ControllerReclaimSpace(ctx, csiReq)
+
+	// FIXME: use ControllerPublishSecret instead, but it is not set in the PV
+	if pv.Spec.CSI.NodeStageSecretRef != nil {
+		// Get the secrets from the k8s cluster
+		csiReq.Secrets, err = kube.GetSecret(ctx, rs.kubeClient, pv.Spec.CSI.NodeStageSecretRef.Name, pv.Spec.CSI.NodeStageSecretRef.Namespace)
+		if err != nil {
+			klog.Errorf("Failed to get secret: %v", err)
+			return nil, status.Error(codes.InvalidArgument, err.Error())
+		}
+	}
+
+	csiRes, err := rs.controllerClient.ControllerReclaimSpace(ctx, csiReq)
 	if err != nil {
 		return nil, err
 	}
+	if csiRes == nil {
+		return nil, status.Error(codes.InvalidArgument, "nil value returned as the response of ControllerReclaimSpace")
+	}
+	res := &proto.ReclaimSpaceResponse{}
+	if csiRes.PreUsage != nil {
+		res.PreUsage = &proto.StorageConsumption{UsageBytes: csiRes.PreUsage.UsageBytes}
+	}
+	if csiRes.PostUsage != nil {
+		res.PostUsage = &proto.StorageConsumption{UsageBytes: csiRes.PostUsage.UsageBytes}
+	}
 
-	return &proto.ReclaimSpaceResponse{
-		PreUsage:  &proto.StorageConsumption{UsageBytes: res.PreUsage.UsageBytes},
-		PostUsage: &proto.StorageConsumption{UsageBytes: res.PostUsage.UsageBytes},
-	}, nil
+	return res, nil
 }
 
 // NodeReclaimSpace fetches required information from kubernetes cluster and calls
@@ -88,22 +124,56 @@ func (rs *ReclaimSpaceServer) NodeReclaimSpace(
 	pvName := req.GetPvName()
 	klog.Info(pvName)
 
-	// TODO: get following params using kubeClient and pvName
+	pv, err := rs.kubeClient.CoreV1().PersistentVolumes().Get(ctx, pvName, metav1.GetOptions{})
+	if err != nil {
+		klog.Errorf("Failed to get pv: %v", err)
+		return nil, status.Errorf(codes.InvalidArgument, "failed to get pv %q", pvName)
+	}
+
+	if pv.Spec.CSI == nil {
+		return nil, status.Errorf(codes.InvalidArgument, "pv %q is not a CSI volume", pvName)
+	}
+	volID := pv.Spec.CSI.VolumeHandle
+	csiMode, err := accessmodes.ToCSIAccessMode(pv.Spec.AccessModes, true)
+	if err != nil {
+		klog.Errorf("Failed to map access mode: %v", err)
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+	stPath := filepath.Join(rs.stagingPath, pvName, "globalmount")
 
 	csiReq := &csiReclaimSpace.NodeReclaimSpaceRequest{
-		VolumeId:          "",
+		VolumeId:          volID,
 		VolumePath:        "",
-		StagingTargetPath: "",
-		VolumeCapability:  &csi.VolumeCapability{},
-		Secrets:           map[string]string{},
+		StagingTargetPath: stPath,
+		VolumeCapability: &csi.VolumeCapability{
+			AccessMode: &csi.VolumeCapability_AccessMode{
+				Mode: csiMode,
+			},
+		},
 	}
-	res, err := rs.nodeClient.NodeReclaimSpace(ctx, csiReq)
+
+	if pv.Spec.CSI.NodeStageSecretRef != nil {
+		// Get the secrets from the k8s cluster
+		csiReq.Secrets, err = kube.GetSecret(ctx, rs.kubeClient, pv.Spec.CSI.NodeStageSecretRef.Name, pv.Spec.CSI.NodeStageSecretRef.Namespace)
+		if err != nil {
+			klog.Errorf("Failed to get secret: %v", err)
+			return nil, status.Error(codes.InvalidArgument, err.Error())
+		}
+	}
+	csiRes, err := rs.nodeClient.NodeReclaimSpace(ctx, csiReq)
 	if err != nil {
 		return nil, err
 	}
+	if csiRes == nil {
+		return nil, status.Error(codes.InvalidArgument, "nil value returned as the response of NodeReclaimSpace")
+	}
+	res := &proto.ReclaimSpaceResponse{}
+	if csiRes.PreUsage != nil {
+		res.PreUsage = &proto.StorageConsumption{UsageBytes: csiRes.PreUsage.UsageBytes}
+	}
+	if csiRes.PostUsage != nil {
+		res.PostUsage = &proto.StorageConsumption{UsageBytes: csiRes.PostUsage.UsageBytes}
+	}
 
-	return &proto.ReclaimSpaceResponse{
-		PreUsage:  &proto.StorageConsumption{UsageBytes: res.PreUsage.UsageBytes},
-		PostUsage: &proto.StorageConsumption{UsageBytes: res.PostUsage.UsageBytes},
-	}, nil
+	return res, nil
 }
--- a/sidecar/main.go
+++ b/sidecar/main.go
@@ -33,12 +33,18 @@ import (
 func main() {
 	var (
 		defaultTimeout     = time.Minute * 3
+		defaultStagingPath = "/var/lib/kubelet/plugins/kubernetes.io/csi/pv/"
 		timeout            = flag.Duration("timeout", defaultTimeout, "Timeout for waiting for response")
 		csiAddonsAddress   = flag.String("csi-addons-address", "/run/csi-addons/socket", "CSI Addons endopoint")
 		nodeID             = flag.String("node-id", "", "NodeID")
+		stagingPath        = flag.String("stagingpath", defaultStagingPath, "stagingpath")
 		controllerEndpoint = flag.String("controller-endpoint", "",
 			"The TCP network port where the gRPC server for controller request, will listen (example: `8080`)")
 	)
+	klog.InitFlags(nil)
+	if err := flag.Set("logtostderr", "true"); err != nil {
+		klog.Exitf("failed to set logtostderr flag: %v", err)
+	}
 	flag.Parse()
 
 	csiClient, err := client.New(*csiAddonsAddress, *timeout)
@@ -73,7 +79,7 @@ func main() {
 
 	sidecarServer := server.NewSidecarServer(*controllerEndpoint)
 	sidecarServer.RegisterService(service.NewIdentityServer(csiClient.Client))
-	sidecarServer.RegisterService(service.NewReclaimSpaceServer(csiClient.Client, kubeClient))
+	sidecarServer.RegisterService(service.NewReclaimSpaceServer(csiClient.Client, kubeClient, *stagingPath))
 	sidecarServer.RegisterService(service.NewNetworkFenceServer(csiClient.Client, kubeClient))
 
 	sidecarServer.Start()
--- a/vendor/github.com/kubernetes-csi/csi-lib-utils/accessmodes/access_modes.go
+++ b/vendor/github.com/kubernetes-csi/csi-lib-utils/accessmodes/access_modes.go
@@ -1,0 +1,128 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package accessmodes
+
+import (
+	"fmt"
+
+	"github.com/container-storage-interface/spec/lib/go/csi"
+	v1 "k8s.io/api/core/v1"
+)
+
+// ToCSIAccessMode maps PersistentVolume access modes in Kubernetes to CSI
+// access modes. Which mapping is used depends if the driver supports the
+// SINGLE_NODE_MULTI_WRITER capability.
+func ToCSIAccessMode(pvAccessModes []v1.PersistentVolumeAccessMode, supportsSingleNodeMultiWriter bool) (csi.VolumeCapability_AccessMode_Mode, error) {
+	if supportsSingleNodeMultiWriter {
+		return toSingleNodeMultiWriterCapableCSIAccessMode(pvAccessModes)
+	}
+	return toCSIAccessMode(pvAccessModes)
+}
+
+// toCSIAccessMode maps PersistentVolume access modes in Kubernetes to CSI
+// access modes.
+//
+// +------------------+-------------------------+----------------------------------------+
+// |  K8s AccessMode  |     CSI AccessMode      |           Additional Details           |
+// +------------------+-------------------------+----------------------------------------+
+// | ReadWriteMany    | MULTI_NODE_MULTI_WRITER |                                        |
+// | ReadOnlyMany     | MULTI_NODE_READER_ONLY  | Cannot be combined with ReadWriteOnce  |
+// | ReadWriteOnce    | SINGLE_NODE_WRITER      | Cannot be combined with ReadOnlyMany   |
+// | ReadWriteOncePod | SINGLE_NODE_WRITER      | Cannot be combined with any AccessMode |
+// +------------------+-------------------------+----------------------------------------+
+func toCSIAccessMode(pvAccessModes []v1.PersistentVolumeAccessMode) (csi.VolumeCapability_AccessMode_Mode, error) {
+	m := uniqueAccessModes(pvAccessModes)
+
+	switch {
+	// This mapping exists to enable CSI drivers that lack the
+	// SINGLE_NODE_MULTI_WRITER capability to work with the
+	// ReadWriteOncePod access mode.
+	case m[v1.ReadWriteOncePod]:
+		if len(m) > 1 {
+			return csi.VolumeCapability_AccessMode_UNKNOWN, fmt.Errorf("Kubernetes does not support use of ReadWriteOncePod with other access modes on the same PersistentVolume")
+		}
+		return csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER, nil
+
+	case m[v1.ReadWriteMany]:
+		// ReadWriteMany takes precedence, regardless of what other
+		// modes are set.
+		return csi.VolumeCapability_AccessMode_MULTI_NODE_MULTI_WRITER, nil
+
+	case m[v1.ReadOnlyMany] && m[v1.ReadWriteOnce]:
+		// This is not possible in the CSI spec.
+		return csi.VolumeCapability_AccessMode_UNKNOWN, fmt.Errorf("CSI does not support ReadOnlyMany and ReadWriteOnce on the same PersistentVolume")
+
+	case m[v1.ReadOnlyMany]:
+		return csi.VolumeCapability_AccessMode_MULTI_NODE_READER_ONLY, nil
+
+	case m[v1.ReadWriteOnce]:
+		return csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER, nil
+
+	default:
+		return csi.VolumeCapability_AccessMode_UNKNOWN, fmt.Errorf("unsupported AccessMode combination: %+v", pvAccessModes)
+	}
+}
+
+// toSingleNodeMultiWriterCapableCSIAccessMode maps PersistentVolume access
+// modes in Kubernetes to CSI access modes for drivers that support the
+// SINGLE_NODE_MULTI_WRITER capability.
+//
+// +------------------+---------------------------+----------------------------------------+
+// |  K8s AccessMode  |      CSI AccessMode       |           Additional Details           |
+// +------------------+---------------------------+----------------------------------------+
+// | ReadWriteMany    | MULTI_NODE_MULTI_WRITER   |                                        |
+// | ReadOnlyMany     | MULTI_NODE_READER_ONLY    | Cannot be combined with ReadWriteOnce  |
+// | ReadWriteOnce    | SINGLE_NODE_MULTI_WRITER  | Cannot be combined with ReadOnlyMany   |
+// | ReadWriteOncePod | SINGLE_NODE_SINGLE_WRITER | Cannot be combined with any AccessMode |
+// +------------------+---------------------------+----------------------------------------+
+func toSingleNodeMultiWriterCapableCSIAccessMode(pvAccessModes []v1.PersistentVolumeAccessMode) (csi.VolumeCapability_AccessMode_Mode, error) {
+	m := uniqueAccessModes(pvAccessModes)
+
+	switch {
+	case m[v1.ReadWriteOncePod]:
+		if len(m) > 1 {
+			return csi.VolumeCapability_AccessMode_UNKNOWN, fmt.Errorf("Kubernetes does not support use of ReadWriteOncePod with other access modes on the same PersistentVolume")
+		}
+		return csi.VolumeCapability_AccessMode_SINGLE_NODE_SINGLE_WRITER, nil
+
+	case m[v1.ReadWriteMany]:
+		// ReadWriteMany trumps everything, regardless of what other
+		// modes are set.
+		return csi.VolumeCapability_AccessMode_MULTI_NODE_MULTI_WRITER, nil
+
+	case m[v1.ReadOnlyMany] && m[v1.ReadWriteOnce]:
+		// This is not possible in the CSI spec.
+		return csi.VolumeCapability_AccessMode_UNKNOWN, fmt.Errorf("CSI does not support ReadOnlyMany and ReadWriteOnce on the same PersistentVolume")
+
+	case m[v1.ReadOnlyMany]:
+		return csi.VolumeCapability_AccessMode_MULTI_NODE_READER_ONLY, nil
+
+	case m[v1.ReadWriteOnce]:
+		return csi.VolumeCapability_AccessMode_SINGLE_NODE_MULTI_WRITER, nil
+
+	default:
+		return csi.VolumeCapability_AccessMode_UNKNOWN, fmt.Errorf("unsupported AccessMode combination: %+v", pvAccessModes)
+	}
+}
+
+func uniqueAccessModes(pvAccessModes []v1.PersistentVolumeAccessMode) map[v1.PersistentVolumeAccessMode]bool {
+	m := map[v1.PersistentVolumeAccessMode]bool{}
+	for _, mode := range pvAccessModes {
+		m[mode] = true
+	}
+	return m
+}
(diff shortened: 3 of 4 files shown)


sidecar/internal/service/reclaimspace.go
```diff
@@ -61,12 +97,18 @@ func (rs *ReclaimSpaceServer) ControllerReclaimSpace(
 	pvName := req.GetPvName()
 	klog.Info(pvName)
 
-	// TODO: get following params using kubeClient and pvName
+	pv, err := rs.kubeClient.CoreV1().PersistentVolumes().Get(context.TODO(), pvName, metav1.GetOptions{})
+	volID := pv.Spec.CSI.VolumeHandle
+	volattributes := pv.Spec.CSI.VolumeAttributes
+
+	deviceSecret, err := getSecret(rs.kubeClient,
+		pv.Spec.CSI.NodeStageSecretRef.Namespace,
```
nixpanic, line 105: you will need to check if `pv.Spec.CSI.NodeStageSecretRef != nil`. Not all CSI-drivers set this. Only create the `deviceSecret` if it is non-empty.

sidecar/internal/service/reclaimspace.go
```diff
@@ -52,6 +54,40 @@ func (rs *ReclaimSpaceServer) RegisterService(server grpc.ServiceRegistrar) {
 	proto.RegisterReclaimSpaceServer(server, rs)
 }
 
+// accessModeStrToInt convert access mode type string to int32.
+// Make sure to update this function as and when there are new modes introduced.
+func accessModeStrToInt(mode v1.PersistentVolumeAccessMode) csi.VolumeCapability_AccessMode_Mode {
```
nixpanic, line 59: although the `..ToInt` name makes sense, you are returning a `csi.VolumeCapability_AccessMode_Mode` type. It might be clearer to rename the function to `k8sAccessModeToCSIAccessMode` or something like that

sidecar/internal/service/reclaimspace.go
```diff
@@ -88,14 +130,25 @@ func (rs *ReclaimSpaceServer) NodeReclaimSpace(
 	pvName := req.GetPvName()
 	klog.Info(pvName)
 
-	// TODO: get following params using kubeClient and pvName
+	pv, err := rs.kubeClient.CoreV1().PersistentVolumes().Get(context.TODO(), pvName, metav1.GetOptions{})
+	volID := pv.Spec.PersistentVolumeSource.CSI.VolumeHandle
+
+	deviceSecret, err := getSecret(rs.kubeClient,
+		pv.Spec.CSI.NodeStageSecretRef.Namespace,
```
nixpanic, line 137: same here, check for the existence of the secret options in the PV before using them

nixpanic:
Also, there is a type in the commit message (and PR title): Implemntation

sidecar/internal/service/reclaimspace.go
Madhu-1: possible to use https://github.com/kubernetes-csi/external-provisioner/blob/61e5b2d16c6bc7d3dbb0da81579cd57503f6c971/vendor/github.com/kubernetes-csi/csi-lib-utils/accessmodes/access_modes.go#L29?

sidecar/internal/service/reclaimspace.go
```diff
@@ -52,6 +54,40 @@ func (rs *ReclaimSpaceServer) RegisterService(server grpc.ServiceRegistrar) {
...
+	return csi.VolumeCapability_AccessMode_UNKNOWN
+}
+
+// getSecret get the secret details by name.
+func getSecret(kube *kubernetes.Clientset, ns, name string) (map[string]string, error) {
+	deviceSecret := make(map[string]string)
+
+	secret, err := kube.CoreV1().Secrets(ns).Get(context.TODO(), name, metav1.GetOptions{})
+	if err != nil {
+		klog.Info("get secret failed, err: %v", err)
```
Madhu-1, line 80: should be error log

sidecar/internal/service/reclaimspace.go
```diff
@@ -52,6 +54,40 @@ func (rs *ReclaimSpaceServer) RegisterService(server grpc.ServiceRegistrar) {
...
+		return csi.VolumeCapability_AccessMode_MULTI_NODE_MULTI_WRITER
+	case v1.ReadWriteOncePod:
+		return csi.VolumeCapability_AccessMode_SINGLE_NODE_SINGLE_WRITER
+	}
+
+	return csi.VolumeCapability_AccessMode_UNKNOWN
+}
+
+// getSecret get the secret details by name.
+func getSecret(kube *kubernetes.Clientset, ns, name string) (map[string]string, error) {
```
Madhu-1, line 75: kube *kubernetes.Clientset to `client *kubernetes.Clientset`

sidecar/internal/service/reclaimspace.go
```diff
@@ -61,12 +97,18 @@ func (rs *ReclaimSpaceServer) ControllerReclaimSpace(
 	pvName := req.GetPvName()
 	klog.Info(pvName)
 
-	// TODO: get following params using kubeClient and pvName
+	pv, err := rs.kubeClient.CoreV1().PersistentVolumes().Get(context.TODO(), pvName, metav1.GetOptions{})
+	volID := pv.Spec.CSI.VolumeHandle
```
Madhu-1, line 101: you need to add check for `pvc.Spec.CSI !=nil`

sidecar/internal/service/reclaimspace.go
```diff
@@ -88,14 +130,25 @@ func (rs *ReclaimSpaceServer) NodeReclaimSpace(
 	pvName := req.GetPvName()
 	klog.Info(pvName)
 
-	// TODO: get following params using kubeClient and pvName
+	pv, err := rs.kubeClient.CoreV1().PersistentVolumes().Get(context.TODO(), pvName, metav1.GetOptions{})
+	volID := pv.Spec.PersistentVolumeSource.CSI.VolumeHandle
```
Madhu-1, line 134: same here i think

sidecar/internal/service/reclaimspace.go
```diff
@@ -88,14 +130,25 @@ func (rs *ReclaimSpaceServer) NodeReclaimSpace(
...
+
+	deviceSecret, err := getSecret(rs.kubeClient,
+		pv.Spec.CSI.NodeStageSecretRef.Namespace,
+		pv.Spec.CSI.NodeStageSecretRef.Name)
+
+	var stagingPath string = "/var/lib/kubelet/plugins/kubernetes.io/csi/pv/"
```
Rakshith-R, line 140: Make this configurable through cmdline args similar to the following 
https://github.com/ceph/ceph-csi/blob/bc8ef898112169b15b8a749d1f4f6b4d84284680/deploy/rbd/kubernetes/csi-rbdplugin.yaml#L57
https://github.com/ceph/ceph-csi/blob/bc8ef898112169b15b8a749d1f4f6b4d84284680/cmd/cephcsi.go#L67

And add it as a member of ReclaimSpaceServer struct.

sidecar/internal/service/reclaimspace.go
Rakshith-R: use `kubeClient` as standard throughout the project ?? 
just client will refer to grpc client pointing towards the csi driver.

Please move internal functions to the end of file after exported/implemented functions for readability(atleast what i personally follow :sweat_smile: )

sidecar/internal/service/reclaimspace.go
```diff
@@ -107,3 +139,38 @@ func (rs *ReclaimSpaceServer) NodeReclaimSpace(
...
+		return csi.VolumeCapability_AccessMode_MULTI_NODE_MULTI_WRITER
+	case v1.ReadWriteOncePod:
+		return csi.VolumeCapability_AccessMode_SINGLE_NODE_SINGLE_WRITER
+	}
+
+	return csi.VolumeCapability_AccessMode_UNKNOWN
+}
+
+// getSecret get the secret details by name.
+func getSecret(kubeclient *kubernetes.Clientset, ns, name string) (map[string]string, error) {
```
Madhu-1, line 162: move this to new packages we need this for other controllers also like fencing

yati1998: okay

sidecar/internal/service/reclaimspace.go
```diff
@@ -61,12 +66,29 @@ func (rs *ReclaimSpaceServer) ControllerReclaimSpace(
 	pvName := req.GetPvName()
 	klog.Info(pvName)
 
-	// TODO: get following params using kubeClient and pvName
+	var deviceSecret map[string]string
+	var volID string
+	var volAttributes map[string]string
+
+	pv, err := rs.kubeClient.CoreV1().PersistentVolumes().Get(context.TODO(), pvName, metav1.GetOptions{})
```
Madhu-1, line 73: err check is missing here

Madhu-1, line 73: using existing context.

sidecar/internal/service/reclaimspace.go
```diff
@@ -61,12 +66,29 @@ func (rs *ReclaimSpaceServer) ControllerReclaimSpace(
...
+	if pv.Spec.CSI != nil {
+		volID = pv.Spec.CSI.VolumeHandle
+		volAttributes = pv.Spec.CSI.VolumeAttributes
+
+		if pv.Spec.CSI.NodeStageSecretRef != nil {
+			// Get the secrets from the k8s cluster
+			deviceSecret, err = kube.GetSecret(ctx, rs.kubeClient, pv.Spec.CSI.NodeStageSecretRef.Name, pv.Spec.CSI.NodeStageSecretRef.Namespace)
```
Madhu-1, line 80: not all drivers will have nodestage secrets. you might need to check if publish secrets exist.

sidecar/internal/service/reclaimspace.go
```diff
@@ -88,14 +110,34 @@ func (rs *ReclaimSpaceServer) NodeReclaimSpace(
...
+	if pv.Spec.CSI == nil {
+		volID = pv.Spec.CSI.VolumeHandle
+		if pv.Spec.CSI.NodeStageSecretRef != nil {
+			// Get the secrets from the k8s cluster
+			deviceSecret, err = kube.GetSecret(ctx, rs.kubeClient, pv.Spec.CSI.NodeStageSecretRef.Name, pv.Spec.CSI.NodeStageSecretRef.Namespace)
+			if err != nil {
+				klog.Errorf("Failed to get secret %v", err)
+				return nil, status.Error(codes.InvalidArgument, err.Error())
+			}
+		}
```
Madhu-1, line 125: move this to a helper function and reuse it?

sidecar/internal/service/reclaimspace.go
Madhu-1, line 80: as this is for controller use provisioner-secret-name? nodesecret might not have required access to what provisioner can do.

sidecar/internal/service/reclaimspace.go
```diff
@@ -88,14 +110,34 @@ func (rs *ReclaimSpaceServer) NodeReclaimSpace(
...
+	}
+
+	//TODO: Configure the stagingPath using cmdline args
+	var stagingPath string = "/var/lib/kubelet/plugins/kubernetes.io/csi/pv/"
```
Madhu-1, line 129: not all system will have `/var/lib/kubelet/` you should make it configurable

sidecar/internal/service/reclaimspace.go
```diff
@@ -88,14 +110,34 @@ func (rs *ReclaimSpaceServer) NodeReclaimSpace(
...
-		VolumeId:          "",
+		VolumeId:          volID,
 		VolumePath:        "",
-		StagingTargetPath: "",
-		VolumeCapability:  &csi.VolumeCapability{},
-		Secrets:           map[string]string{},
+		StagingTargetPath: stagingPath + pvName,
+		VolumeCapability: &csi.VolumeCapability{
+			AccessMode: &csi.VolumeCapability_AccessMode{
+				Mode: k8sAccessModeToCSIAccessMode(pv.Spec.AccessModes[0]),
```
Madhu-1, line 137: you should not choose one access mode here. we need something similar to https://github.com/kubernetes-csi/external-provisioner/blob/61e5b2d16c6bc7d3dbb0da81579cd57503f6c971/vendor/github.com/kubernetes-csi/csi-lib-utils/accessmodes/access_modes.go#L29

sidecar/internal/service/reclaimspace.go
yati1998: it is slightly different for both cases.  In the case of controllerreclaimspace we need to get volume attributes while for node we don't need that. Hence, in my opinion, it is fine.

sidecar/internal/service/reclaimspace.go
```diff
@@ -61,12 +66,29 @@ func (rs *ReclaimSpaceServer) ControllerReclaimSpace(
 	pvName := req.GetPvName()
 	klog.Info(pvName)
 
-	// TODO: get following params using kubeClient and pvName
+	var deviceSecret map[string]string
+	var volID string
+	var volAttributes map[string]string
+
+	pv, err := rs.kubeClient.CoreV1().PersistentVolumes().Get(context.TODO(), pvName, metav1.GetOptions{})
+	if pv.Spec.CSI != nil {
```
nixpanic, line 74: better check

```go
if pv.Spec.CSI == nil {
    return nil, status.Errorf(codes.InvalidArgument, "pv %q is not a CSI volume", pvName)
}

volID = ...
```

sidecar/internal/service/reclaimspace.go
```diff
@@ -61,12 +66,29 @@ func (rs *ReclaimSpaceServer) ControllerReclaimSpace(
...
+		}
+	}
 
 	csiReq := &csiReclaimSpace.ControllerReclaimSpaceRequest{
-		VolumeId:   "",
-		Parameters: map[string]string{},
-		Secrets:    map[string]string{},
+		VolumeId:   volID,
+		Parameters: volAttributes,
+		Secrets:    deviceSecret,
```
nixpanic, line 91: you can create `csiReq` earlier in the function, and only set `VolumeID` there. Then
```
if pv.Spec.CSI.NodeStageSecretRef != nil {
    csiReq.Secrets, err = kube.GetSecret(...)
}

res, err := rs.controllerClient.ControllerReclaimSpace(ctx, csiReq)
```

this hopefully makes the function a little simpler.

sidecar/internal/service/reclaimspace.go
```diff
@@ -88,14 +110,34 @@ func (rs *ReclaimSpaceServer) NodeReclaimSpace(
 	pvName := req.GetPvName()
 	klog.Info(pvName)
 
-	// TODO: get following params using kubeClient and pvName
+	pv, err := rs.kubeClient.CoreV1().PersistentVolumes().Get(context.TODO(), pvName, metav1.GetOptions{})
+	var deviceSecret map[string]string
+	var volID string
+	if pv.Spec.CSI == nil {
```
nixpanic, line 116: check for
```go
if pv.Spec.CSI == nil {
    return nil, status.Errorf(codes.InvalidArgument, "pv %q is not a CSI volume", pvName)
}
```
and then continue with the success case.

sidecar/internal/service/reclaimspace.go
```diff
@@ -88,14 +110,34 @@ func (rs *ReclaimSpaceServer) NodeReclaimSpace(
...
+	var stagingPath string = "/var/lib/kubelet/plugins/kubernetes.io/csi/pv/"
 
 	csiReq := &csiReclaimSpace.NodeReclaimSpaceRequest{
-		VolumeId:          "",
+		VolumeId:          volID,
 		VolumePath:        "",
-		StagingTargetPath: "",
-		VolumeCapability:  &csi.VolumeCapability{},
-		Secrets:           map[string]string{},
+		StagingTargetPath: stagingPath + pvName,
```
nixpanic, line 134: It seems that `StagingTargetPath` in the CSI `NodeStageVolume` contains `/globalmount` as well. It should be included here too.

```
I1223 12:01:43.532455  375675 utils.go:191] ID: 45 Req-ID: 0001-0011-openshift-storage-0000000000000001-96e9a7ce-63c8-11ec-89d0-0a580a83001c GRPC call: /csi.v1.Node/NodeStageVolume
I1223 12:01:43.532652  375675 utils.go:195] ID: 45 Req-ID: 0001-0011-openshift-storage-0000000000000001-96e9a7ce-63c8-11ec-89d0-0a580a83001c GRPC request: {"secrets":"***stripped***","staging_target_path":"/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-720d76c5-d994-4eb8-b8c0-d2bc5f493c17/globalmount
```

sidecar/internal/service/reclaimspace.go
```diff
@@ -88,14 +116,37 @@ func (rs *ReclaimSpaceServer) NodeReclaimSpace(
...
+	if err != nil {
+		klog.Errorf("Failed to get pv: %v", err)
+	}
 
+	if pv.Spec.CSI == nil {
+		return nil, status.Errorf(codes.InvalidArgument, "pv %q is not a CSI volume", pvName)
+	}
+	volID := pv.Spec.CSI.VolumeHandle
+	csiMode, err := accessmodes.ToCSIAccessMode(pv.Spec.AccessModes, true)
+	stPath := *rs.stagingPath + pvName + "/globalmount"
```
nixpanic, line 129: This requires `rs.stagingPath` to end with a `/`, otherwise the path becomes invalid. Consider adding slashes with `fmt.Sprintf()` or use [`filepath.Join()`](https://pkg.go.dev/path/filepath#Join).

sidecar/internal/service/reclaimspace.go
```diff
@@ -88,14 +116,37 @@ func (rs *ReclaimSpaceServer) NodeReclaimSpace(
 	pvName := req.GetPvName()
 	klog.Info(pvName)
 
-	// TODO: get following params using kubeClient and pvName
+	pv, err := rs.kubeClient.CoreV1().PersistentVolumes().Get(ctx, pvName, metav1.GetOptions{})
+	if err != nil {
+		klog.Errorf("Failed to get pv: %v", err)
```
nixpanic, line 121: return an error here, do not only log it.

sidecar/internal/service/reclaimspace.go
```diff
@@ -35,15 +40,17 @@ type ReclaimSpaceServer struct {
 	controllerClient csiReclaimSpace.ReclaimSpaceControllerClient
 	nodeClient       csiReclaimSpace.ReclaimSpaceNodeClient
 	kubeClient       *kubernetes.Clientset
+	stagingPath      *string
```
nixpanic, line 43: why a `*string`? using a non-pointer `string` will be simpler.

sidecar/internal/service/reclaimspace.go
```diff
@@ -61,13 +68,34 @@ func (rs *ReclaimSpaceServer) ControllerReclaimSpace(
 	pvName := req.GetPvName()
 	klog.Info(pvName)
 
-	// TODO: get following params using kubeClient and pvName
+	pv, err := rs.kubeClient.CoreV1().PersistentVolumes().Get(ctx, pvName, metav1.GetOptions{})
+	if err != nil {
+		klog.Errorf("Failed to get pv: %v", err)
```
nixpanic, line 73: need to return an error, not only log it

nixpanic (changes requested):
few minor changes needed, looks almost good to go IMHO

sidecar/internal/service/reclaimspace.go
yati1998: ohh yeah.

sidecar/internal/service/reclaimspace.go
yati1998: the flag.string returns *string

nixpanic: that is no an excuse to use a pointer in this struct :smiley:

sidecar/internal/service/reclaimspace.go
nixpanic: The check seems to be incorrect. You need to check for `pv.Spec.CSI == nil` as that is the error case.

sidecar/internal/service/reclaimspace.go
yati1998: I will change it.

sidecar/internal/service/reclaimspace.go
```diff
@@ -61,13 +69,35 @@ func (rs *ReclaimSpaceServer) ControllerReclaimSpace(
...
+		return nil, status.Errorf(codes.InvalidArgument, "failed to get pv")
```
nixpanic, line 75: include the name of the pv in the error, like `status.Errorf(codes.InvalidArgument, "failed to get pv %q", pvName)`. When a failure is returned and logged, it helps the person troubleshooting it a lot to have these details.

sidecar/main.go
```diff
@@ -73,7 +79,7 @@ func main() {
 
 	sidecarServer := server.NewSidecarServer(*controllerEndpoint)
 	sidecarServer.RegisterService(service.NewIdentityServer(csiClient.Client))
-	sidecarServer.RegisterService(service.NewReclaimSpaceServer(csiClient.Client, kubeClient))
+	sidecarServer.RegisterService(service.NewReclaimSpaceServer(csiClient.Client, kubeClient, stagingPath))
```
nixpanic, line 82: you can pass `*stagingPath` so that you do not need to care about the pointer indirection in the `ReclaimSpaceServer` struct

sidecar/internal/service/reclaimspace.go
nixpanic: This still needs to be fixed. Once done, I am happy to approve this.

nixpanic (changes requested):

nixpanic (approved):

sidecar/internal/service/reclaimspace.go
```diff
@@ -61,13 +69,35 @@ func (rs *ReclaimSpaceServer) ControllerReclaimSpace(
...
+	if pv.Spec.CSI.NodeStageSecretRef != nil {
+		// Get the secrets from the k8s cluster
+		csiReq.Secrets, err = kube.GetSecret(ctx, rs.kubeClient, pv.Spec.CSI.NodeStageSecretRef.Name, pv.Spec.CSI.NodeStageSecretRef.Namespace)
+		if err != nil {
+			klog.Errorf("Failed to get secret %v", err)
+			return nil, status.Error(codes.InvalidArgument, err.Error())
+		}
 	}
+
 	res, err := rs.controllerClient.ControllerReclaimSpace(ctx, csiReq)
```
Rakshith-R, line 101:
```suggestion
	res, err := rs.controllerClient.ControllerReclaimSpace(ctx, csiReq)
```

check res parameters are nil or not, other wise it throws panic in return 
```
[rakshith@fedora kubernetes-csi-addons]$ k logs csi-rbdplugin-provisioner-7bb77f549b-grpbg csi-addons-k8s-sidecar 
I1224 13:39:33.475621       1 client.go:64] Probing CSI driver for readiness
I1224 13:39:33.492878       1 server.go:89] Listening for CSI-Addons requests on address: &net.TCPAddr{IP:net.IP{0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0}, Port:6202, Zone:""}
I1224 13:40:37.915775       1 reclaimspace.go:70] pvc-eff12e5c-8bb9-4e86-b7bf-4b5d21c2c889
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x13c9565]

goroutine 75 [running]:
github.com/csi-addons/kubernetes-csi-addons/sidecar/internal/service.(*ReclaimSpaceServer).ControllerReclaimSpace(0xc00003bd40, {0x1937638, 0xc0001fe240}, 0xc000116c80)
        /workspace/go/src/github.com/csi-addons/kubernetes-csi-addons/sidecar/internal/service/reclaimspace.go:107 +0x5a5
github.com/csi-addons/kubernetes-csi-addons/internal/proto._ReclaimSpace_ControllerReclaimSpace_Handler({0x15a5fc0, 0xc00003bd40}, {0x1937638, 0xc0001fe240}, 0xc000280a20, 0x0)
        /workspace/go/src/github.com/csi-addons/kubernetes-csi-addons/internal/proto/reclaimspace_grpc.pb.go:97 +0x170
google.golang.org/grpc.(*Server).processUnaryRPC(0xc00013a700, {0x195f0e8, 0xc00028a4e0}, 0xc000196240, 0xc0001d1740, 0x25402c0, 0x0)
        /workspace/go/src/github.com/csi-addons/kubernetes-csi-addons/vendor/google.golang.org/grpc/server.go:1282 +0xccf
google.golang.org/grpc.(*Server).handleStream(0xc00013a700, {0x195f0e8, 0xc00028a4e0}, 0xc000196240, 0x0)
        /workspace/go/src/github.com/csi-addons/kubernetes-csi-addons/vendor/google.golang.org/grpc/server.go:1616 +0xa2a
google.golang.org/grpc.(*Server).serveStreams.func1.2()
        /workspace/go/src/github.com/csi-addons/kubernetes-csi-addons/vendor/google.golang.org/grpc/server.go:921 +0x98
created by google.golang.org/grpc.(*Server).serveStreams.func1
        /workspace/go/src/github.com/csi-addons/kubernetes-csi-addons/vendor/google.golang.org/grpc/server.go:919 +0x294
        ~
        ```

File: sidecar/internal/service/reclaimspace.go [hunk 26]
```diff
@@ -88,14 +118,38 @@ func (rs *ReclaimSpaceServer) NodeReclaimSpace(
...
+	if pv.Spec.CSI.NodeStageSecretRef != nil {
+		// Get the secrets from the k8s cluster
+		csiReq.Secrets, err = kube.GetSecret(ctx, rs.kubeClient, pv.Spec.CSI.NodeStageSecretRef.Name, pv.Spec.CSI.NodeStageSecretRef.Namespace)
+		if err != nil {
+			klog.Errorf("Failed to get secret %v", err)
+			return nil, status.Error(codes.InvalidArgument, err.Error())
+		}
 	}
 	res, err := rs.nodeClient.NodeReclaimSpace(ctx, csiReq)
 	if err != nil {
```
Rakshith-R, line 155: same here

Rakshith-R (changes requested):

Rakshith-R:
Since `StorageConsumption pre_usage,post_usage` is optional, cephcsi returns nil values in csi grpc response, 
This needs to be handled accordingly to avoid panic.

```
  // This field is OPTIONAL. This allows the SP to inform the CO about the
  // storage consumption before the ReclaimSpace operation was executed.
  StorageConsumption pre_usage = 1;

  // This field is OPTIONAL. This allows the SP to inform the CO about the
  // storage consumption after the ReclaimSpace operation was executed.
  StorageConsumption post_usage = 2;
```

Madhu-1:
> Since `StorageConsumption pre_usage,post_usage` is optional, cephcsi returns nil values in csi grpc response, This needs to be handled accordingly to avoid panic.
> 
> ```
>   // This field is OPTIONAL. This allows the SP to inform the CO about the
>   // storage consumption before the ReclaimSpace operation was executed.
>   StorageConsumption pre_usage = 1;
> 
>   // This field is OPTIONAL. This allows the SP to inform the CO about the
>   // storage consumption after the ReclaimSpace operation was executed.
>   StorageConsumption post_usage = 2;
> ```

the fields are optional but why return nil instead of returning empty structure with nil values?

nixpanic:
> the fields are optional but why return nil instead of returning empty structure with nil values?

Because the fields in the structure are ints. That means, an empty structure would set the fields to `0`, which has a different meaning than a `nil` structure.

sidecar/internal/service/reclaimspace.go
```diff
@@ -61,17 +69,45 @@ func (rs *ReclaimSpaceServer) ControllerReclaimSpace(
...
+	volID := pv.Spec.CSI.VolumeHandle
+	volAttributes := pv.Spec.CSI.VolumeAttributes
+	var deviceSecret map[string]string
 	csiReq := &csiReclaimSpace.ControllerReclaimSpaceRequest{
-		VolumeId:   "",
-		Parameters: map[string]string{},
-		Secrets:    map[string]string{},
+		VolumeId:   volID,
+		Parameters: volAttributes,
+		Secrets:    deviceSecret,
```
nixpanic, line 88: `Secrets` do not need to be set here, they are optional. Just drop the whole `Secrets:    deviceSecret,` line.

sidecar/internal/service/reclaimspace.go
```diff
@@ -61,17 +69,45 @@ func (rs *ReclaimSpaceServer) ControllerReclaimSpace(
...
 	}
+
+	// FIXME: use ControllerPublishSecret instead, but it is not set in the PV
+	if pv.Spec.CSI.NodeStageSecretRef != nil {
+		// Get the secrets from the k8s cluster
+		csiReq.Secrets, err = kube.GetSecret(ctx, rs.kubeClient, pv.Spec.CSI.NodeStageSecretRef.Name, pv.Spec.CSI.NodeStageSecretRef.Namespace)
```
nixpanic, line 94: This is correct and should be kept. `Secrets` only needs to be set when the PV has secrets linked.

sidecar/internal/service/reclaimspace.go
```diff
@@ -61,17 +69,45 @@ func (rs *ReclaimSpaceServer) ControllerReclaimSpace(
...
+			klog.Errorf("Failed to get secret %v", err)
+			return nil, status.Error(codes.InvalidArgument, err.Error())
+		}
+	}
+
 	res, err := rs.controllerClient.ControllerReclaimSpace(ctx, csiReq)
 	if err != nil {
 		return nil, err
 	}
+	if res.PreUsage == nil {
```
nixpanic, line 105: This is not an error or a failure, the `res.PreUsage` can be `nil` as it is an optional attribute. There is no need to log this.

sidecar/internal/service/reclaimspace.go
```diff
@@ -61,17 +69,45 @@ func (rs *ReclaimSpaceServer) ControllerReclaimSpace(
...
+		return nil, status.Error(codes.Unknown, "Empty preusage storage consumption")
+	}
+	if res.PostUsage == nil {
+		return nil, status.Errorf(codes.Unknown, "Empty postusage storage consumption")
+	}
 
 	return &proto.ReclaimSpaceResponse{
```
nixpanic, line 112: you could do this:
```go
r := &proto.ReclaimSpaceResponse{}
if res.PreUsage != nil {
    r.PreUsage = &proto.StorageConsumption{UsageBytes: res.PreUsage.UsageBytes}
}
if res.PostUsage != nil {
    r.PostUsage = &proto.StorageConsumption{UsageBytes: res.PostUsage.UsageBytes}
}

return r, nil
```

sidecar/internal/service/reclaimspace.go
```diff
@@ -61,17 +69,45 @@ func (rs *ReclaimSpaceServer) ControllerReclaimSpace(
...
+	if pv.Spec.CSI.NodeStageSecretRef != nil {
+		// Get the secrets from the k8s cluster
+		csiReq.Secrets, err = kube.GetSecret(ctx, rs.kubeClient, pv.Spec.CSI.NodeStageSecretRef.Name, pv.Spec.CSI.NodeStageSecretRef.Namespace)
+		if err != nil {
+			klog.Errorf("Failed to get secret %v", err)
+			return nil, status.Error(codes.InvalidArgument, err.Error())
+		}
+	}
+
 	res, err := rs.controllerClient.ControllerReclaimSpace(ctx, csiReq)
```
nixpanic, line 101: You use `csiReq` for the request, maybe use `csiRes` for the response? That way you can use `res` for the return value from this function.

nixpanic (changes requested):

nixpanic (dismissed):
Looks good to me, thanks!

sidecar/internal/service/reclaimspace.go
```diff
@@ -88,22 +121,49 @@ func (rs *ReclaimSpaceServer) NodeReclaimSpace(
...
+		klog.Errorf("Failed to get pv: %v", err)
+		return nil, status.Errorf(codes.InvalidArgument, "failed to get pv %q", pvName)
+	}
+
+	if pv.Spec.CSI == nil {
+		return nil, status.Errorf(codes.InvalidArgument, "pv %q is not a CSI volume", pvName)
+	}
+	volID := pv.Spec.CSI.VolumeHandle
+	csiMode, err := accessmodes.ToCSIAccessMode(pv.Spec.AccessModes, true)
+	stPath := filepath.Join(rs.stagingPath, pvName, "globalmount")
```
Rakshith-R, line 135: `err` is not handled here.

Rakshith-R (changes requested):

nixpanic (dismissed):

yati1998:
@Madhu-1 can you review the PR.

sidecar/internal/service/reclaimspace.go
```diff
@@ -61,22 +69,47 @@ func (rs *ReclaimSpaceServer) ControllerReclaimSpace(
...
+			return nil, status.Error(codes.InvalidArgument, err.Error())
+		}
+	}
+
+	csiRes, err := rs.controllerClient.ControllerReclaimSpace(ctx, csiReq)
 	if err != nil {
 		return nil, err
 	}
+	res := &proto.ReclaimSpaceResponse{}
+	if csiRes.PreUsage != nil {
```
Madhu-1, line 105: csiRes can be nil, right? Do we need a check like this?

```
if csiRes !=nil {
    if csiRes.PreUsage != nil {
              res.PreUsage = &proto.StorageConsumption{UsageBytes: csiRes.PreUsage.UsageBytes}
	}
     if csiRes.PostUsage != nil {
		res.PostUsage = &proto.StorageConsumption{UsageBytes: csiRes.PostUsage.UsageBytes}
	}
}

File: sidecar/internal/service/reclaimspace.go [hunk 34]
```diff
@@ -88,22 +121,53 @@ func (rs *ReclaimSpaceServer) NodeReclaimSpace(
...
+			return nil, status.Error(codes.InvalidArgument, err.Error())
+		}
 	}
-	res, err := rs.nodeClient.NodeReclaimSpace(ctx, csiReq)
+	csiRes, err := rs.nodeClient.NodeReclaimSpace(ctx, csiReq)
 	if err != nil {
 		return nil, err
 	}
+	res := &proto.ReclaimSpaceResponse{}
+	if csiRes.PreUsage != nil {
```
Madhu-1, line 165: same here as well.

sidecar/internal/service/reclaimspace.go
yati1998: yes. This is to avoid panic. We can instead send the empty structure. csiRes can be nil as it is optional , it will not throw any error but will get stuck in panic state as shown by @Rakshith-R . @nixpanic let me know if this is correct.

nixpanic: Indeed, if `csiRes` is `nil`, `err` should be set. Checking `csiRes == nil` in below the `if err != nil` and returning a suitable error would be good.

yati1998: Yes IMO too adding the err in case csiRes is nil will be better, as it won't get stuck in a panic state and also we can't return an empty struct.

nixpanic: In general, readability of the code is best when following a flow like:
```go
res, err := someFunction()

// handle failure cases
if err != nil {
   return fmt.Errorf("someFunction failed: %w", err)
}
if res == nil {
    return fmt.Errorf("someFunction returned and unexpected result for %q", someStringArg)
}

// continue with success case
```

Madhu-1: above works for me. res nil check is suggested here https://github.com/csi-addons/kubernetes-csi-addons/pull/32#discussion_r775018426 hence the question

sidecar/internal/service/reclaimspace.go
```diff
@@ -61,22 +70,50 @@ func (rs *ReclaimSpaceServer) ControllerReclaimSpace(
...
+			return nil, status.Error(codes.InvalidArgument, err.Error())
+		}
+	}
+
+	csiRes, err := rs.controllerClient.ControllerReclaimSpace(ctx, csiReq)
 	if err != nil {
 		return nil, err
 	}
+	if csiRes == nil {
+		return nil, fmt.Errorf("nil value returned as the response of ControllerReclaimSpace")
```
Madhu-1, line 106: Use errors.new if no formatting is required

Madhu-1 (approved):

sidecar/internal/service/reclaimspace.go
```diff
@@ -61,22 +70,50 @@ func (rs *ReclaimSpaceServer) ControllerReclaimSpace(
...
+			return nil, status.Error(codes.InvalidArgument, err.Error())
+		}
+	}
+
+	csiRes, err := rs.controllerClient.ControllerReclaimSpace(ctx, csiReq)
 	if err != nil {
 		return nil, err
 	}
+	if csiRes == nil {
+		return nil, errors.New("nil value returned as the response of ControllerReclaimSpace")
```
nixpanic, line 106: this should return a gRPC error, use `status.Error(...)` instead

sidecar/internal/service/reclaimspace.go
```diff
@@ -88,22 +125,56 @@ func (rs *ReclaimSpaceServer) NodeReclaimSpace(
...
+			klog.Errorf("Failed to get secret: %v", err)
+			return nil, status.Error(codes.InvalidArgument, err.Error())
+		}
+	}
+	csiRes, err := rs.nodeClient.NodeReclaimSpace(ctx, csiReq)
 	if err != nil {
 		return nil, err
 	}
+	if csiRes == nil {
+		return nil, errors.New("nil value returned as the response of NodeReclaimSpace")
```
nixpanic, line 169: same here, return `status.Error(...)`

nixpanic (changes requested):
minor change needed, looks good otherwise :+1:

nixpanic (approved):

Madhu-1 (approved):

PR #33, opened 2021-12-22
vendor: create the vendor/ directory and include all dependencies

nixpanic:
The `vendor/` directory can now get created. Downloading all dependencies in advance will speed up building a little bit.

A few simple CI jobs have been added as well, these should guard that the `vendor/` contents is consistent.


.github/workflows/codespell.yml
```diff
@@ -16,6 +16,6 @@ jobs:
       - name: codespell
         uses: codespell-project/actions-codespell@master
         with:
-          skip: .git
+          skip: .git,vendor
```
Rakshith-R, line 19: Can please add go.mod and go.sum too 

refer https://github.com/csi-addons/kubernetes-csi-addons/runs/4604315553?check_suite_focus=true

Rakshith-R (commented):
Do we really need vendor directory ? 
I have not seen operators/controllers using vendor directory

Rakshith-R:
> The `vendor/` directory can now get created. Downloading all dependencies in advance will speed up building a little bit.

This will need some changes in Dockerfile and ./build/Containerfile.sidecar too,  since we just copy go.mod, go.sum and download dependencies. 

https://github.com/csi-addons/kubernetes-csi-addons/blob/0daf3c935289d306346fd57ec398c1c58617eba1/Dockerfile#L6-L10
https://github.com/csi-addons/kubernetes-csi-addons/blob/0daf3c935289d306346fd57ec398c1c58617eba1/build/Containerfile.sidecar#L5-L10

nixpanic:
> Do we really need vendor directory ? I have not seen operators/controllers using vendor directory

Yes, mostly for faster building. It helps while doing an initial clone of the repository, not all dependencies need to be downloaded.

nixpanic:
> This will need some changes in Dockerfile and ./build/Containerfile.sidecar too, since we just copy go.mod, go.sum and download dependencies.

Ah, good point!

.github/workflows/codespell.yml
nixpanic: ok, done!

Rakshith-R:
add ` -mod vendor ` to go build cmd ?

nixpanic:
> add `-mod vendor` to go build cmd ?

Yes, seems to be needed. I expected that new golang version pick it up automatically, it seems not to be the case.

yati1998 (approved):

.github/workflows/test-golang.yaml
```diff
@@ -0,0 +1,41 @@
+---
+name: Run Golang tests
+
+on:
+  pull_request:
+    branches:
+      - "*"
+
+jobs:
+  go_mod_verify:
```
Madhu-1, line 10: update mergify rules for new jobs?

nixpanic: yes, but only after this is merged by Mergify

Madhu-1 (approved):

nixpanic:
@mergifyio rebase

humblec (approved):

PR #39, opened 2021-12-22
ci: build container image on one arch

Madhu-1:
To fasten the CI jobs for testing builds, build container images only on one arch and when pushing build on all arch.

### DIFF
--- a/.github/workflows/build-push.yaml
+++ b/.github/workflows/build-push.yaml
@@ -33,7 +33,7 @@ jobs:
         with:
           context: .
           file: Dockerfile
-          platforms: linux/amd64,linux/arm64,linux/ppc64le,linux/s390x,linux/arm/v7
+          platforms: linux/amd64,linux/arm64,linux/arm/v7
           push: true
           tags: quay.io/csiaddons/k8s-controller:latest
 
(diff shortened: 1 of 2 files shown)


Madhu-1:
build time from 25+ minutes is reduced to 1 min

nixpanic (changes requested):
> build time from 25+ minutes is reduced to 1 min

Sure, but that might break the build. It would prevent pushing the image after merging, and there is no obvious failure reported in case that happens.

I don't think these delays are critical.

Maybe drop s390x and maype ppc64le only? These are not architectures that are commonly available for users that consume community provided components (as opposed to enterprise products).

Madhu-1:
25+, not it's taking more than 45+ minutes and still goining. I don't think testing all PR's on all arch is required and not all images will be CI tested only build will be tested. if any failure on the pushing it can be addressed immediately  just because of that we should not slow down the entire CI

nixpanic:
> if any failure on the pushing

How are we getting informed about this? It should create a blocking issue, no other PRs may get merged until the build work again. If that can be guaranteed somehow, building for CI jobs on a single architecture is ok.

Madhu-1:
> > if any failure on the pushing
> 
> How are we getting informed about this? It should create a blocking issue, no other PRs may get merged until the build work again. If that can be guaranteed somehow, building for CI jobs on a single architecture is ok.

we can configure slack to get notifications and add some action to send mails etc.

nixpanic:
> we can configure slack to get notifications and add some action to send mails etc.

Yes, and until that is done, we should keep building on all architectures we want (s390x and ppc64le can be dropped IMHO).

Madhu-1:
okay for now s390x and ppc64le dropped

nixpanic (changes requested):
> okay for now s390x and ppc64le dropped

Drop if from the build-push workflow too. Otherwise it is still possible to break the build without noticing.

Rakshith-R (approved):

PR #40, opened 2021-12-22
controller: add networkfence skeleton

Yuggupta27:
Add networkfence controller skeleton

Signed-off-by: Yug Gupta <yuggupta27@gmail.com>

PR #42, opened 2022-01-05
ci: add a job to run "make test"

nixpanic:
"make test" is an available Makefile target that runs several checks for
Golang coding standards and code generation. In case the command fails,
or uncommitted changes are detected, the job will fail.


Yuggupta27 (approved):
LGTM! Will drop the commits in https://github.com/csi-addons/kubernetes-csi-addons/pull/40 to avoid redundancy.

Madhu-1 (approved):

PR #45, opened 2022-01-05
build: move main.go to cmd/manager/

nixpanic:
The `main.go` file in the root of the repository is a little confusing
as there are more than one `main` applications in this repository. By
moving the `main.go` to its own `cmd/manager` directory, other
executables can easily be added in the future.


yati1998 (approved):

Yuggupta27 (approved):

Rakshith-R (changes requested):
I am a little hesitant to move main.go from root directory of the project, since kubebuilder create cmd will fail. 
https://github.com/kubernetes-sigs/kubebuilder/issues/1719#issue-718392927

if this is not problem, then its fine with me too.
(probably can workaround by moving main.go back to root when creating api/webhooks)

PR #47, opened 2022-01-06
sidecar: use <POD_IP>:<PORT> for controller endpoint

Rakshith-R:
Just using port to start controller server leads to
grpc server listening on all available ip addresses.
This causes plugin pod running on host network to listen
to requests pointed towards pod ips on that particular port
which is undesirable.

Using `<POD_IP>:<PORT>` makes sure grpc server is listening
on its intended endpoint.

### DIFF
--- a/sidecar/internal/csiaddonsnode/csiaddonsnode_test.go
+++ b/sidecar/internal/csiaddonsnode/csiaddonsnode_test.go
@@ -59,12 +59,10 @@ func Test_getCSIAddonsNode(t *testing.T) {
 		podName     = "pod"
 		podNamspace = "default"
 		podUID      = "123"
-		podIP       = "110"
 	)
 	t.Setenv(podNameEnvKey, podName)
 	t.Setenv(podNamespaceEnvKey, podNamspace)
 	t.Setenv(podUIDEnvKey, podUID)
-	t.Setenv(podIPEnvKey, podIP)
 
 	type args struct {
 		driverName string
@@ -82,7 +80,7 @@ func Test_getCSIAddonsNode(t *testing.T) {
 			name: "Test 1",
 			args: args{
 				driverName: "example.com",
-				endpoint:   "6060",
+				endpoint:   "192.168.61.228:6060",
 				nodeID:     "123",
 			},
 			want: &csiaddonsv1alpha1.CSIAddonsNode{
@@ -102,7 +100,7 @@ func Test_getCSIAddonsNode(t *testing.T) {
 				Spec: csiaddonsv1alpha1.CSIAddonsNodeSpec{
 					Driver: csiaddonsv1alpha1.CSIAddonsNodeDriver{
 						Name:     "example.com",
-						EndPoint: podIP + ":" + "6060",
+						EndPoint: "192.168.61.228:6060",
 						NodeID:   "123",
 					},
 				},
@@ -113,7 +111,7 @@ func Test_getCSIAddonsNode(t *testing.T) {
 			name: "Test 2",
 			args: args{
 				driverName: "csi.example.com",
-				endpoint:   "8060",
+				endpoint:   "[2001:0db8:3c4d:0015:0000:0000:1a2f:1a2b]:8080",
 				nodeID:     "32532-1312-435354",
 			},
 			want: &csiaddonsv1alpha1.CSIAddonsNode{
@@ -133,7 +131,7 @@ func Test_getCSIAddonsNode(t *testing.T) {
 				Spec: csiaddonsv1alpha1.CSIAddonsNodeSpec{
 					Driver: csiaddonsv1alpha1.CSIAddonsNodeDriver{
 						Name:     "csi.example.com",
-						EndPoint: podIP + ":" + "8060",
+						EndPoint: "[2001:0db8:3c4d:0015:0000:0000:1a2f:1a2b]:8080",
 						NodeID:   "32532-1312-435354",
 					},
 				},
--- a/sidecar/internal/server/server.go
+++ b/sidecar/internal/server/server.go
@@ -53,7 +53,7 @@ func NewSidecarServer(endpoint string) *SidecarServer {
 	}
 
 	ss.scheme = "tcp"
-	ss.endpoint = ":" + endpoint
+	ss.endpoint = endpoint
 
 	return ss
 }
--- a/sidecar/internal/util/util.go
+++ b/sidecar/internal/util/util.go
@@ -1,0 +1,43 @@
+/*
+Copyright 2022 The Kubernetes-CSI-Addons Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	"fmt"
+	"net"
+	"strconv"
+)
+
+// ValidateControllerEndpoint validates given ip address and port,
+// returning valid endpoint containing both ip address and port.
+func ValidateControllerEndpoint(rawIP, rawPort string) (string, error) {
+	ip := net.ParseIP(rawIP)
+	if ip == nil {
+		return "", fmt.Errorf("invalid controller ip address %q", rawIP)
+	}
+
+	port, err := strconv.ParseUint(rawPort, 10, 16)
+	if err != nil {
+		return "", fmt.Errorf("invalid controller port %q: %w", rawPort, err)
+	}
+
+	if ip.To4() != nil {
+		return fmt.Sprintf("%s:%d", ip.String(), port), nil
+	}
+
+	return fmt.Sprintf("[%s]:%d", ip.String(), port), nil
+}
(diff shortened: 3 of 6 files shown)


sidecar/internal/server/server.go
```diff
@@ -53,7 +53,7 @@ func NewSidecarServer(endpoint string) *SidecarServer {
 	}
 
 	ss.scheme = "tcp"
-	ss.endpoint = ":" + endpoint
+	ss.endpoint = endpoint
```
Madhu-1, line 56: why can't get the podIP env and use it here?

Rakshith-R: > why can't get the podIP env and use it here?

It can be done but I thought 
- It will be better to be explicit that `<pod_ip>:<port>` needs to passed and grpc server will be started to listen this address.
- prevent fetching of podIP env twice.

Madhu-1: >prevent fetching of podIP env twice.
 
fetch once and use it as it's a  single service.

anything is fine. I have seen in most of the places it is passed as the env. you also need to make sure the endpoint contains IP:port format

nixpanic: some validation of the parameters would be good, do that immediately after `flag.Parse()`.

Rakshith-R: added validations, please take a look

sidecar/internal/util/util.go
```diff
@@ -0,0 +1,48 @@
+/*
+Copyright 2021 The Kubernetes-CSI-Addons Authors.
```
Madhu-1, line 2: 2022

sidecar/internal/util/util.go
```diff
@@ -0,0 +1,48 @@
...
+
+	ip := net.ParseIP(rawIP)
+	if ip == nil {
+		return fmt.Errorf("invalid ip address %q in endpoint %q", rawIP, *endpoint)
+	}
+
+	_, err := strconv.ParseUint(splits[1], 10, 16)
+	if err != nil {
+		return fmt.Errorf("invalid port %q in endpoint %q: %w", rawPort, *endpoint, err)
+	}
```
Madhu-1, line 45: possible to use https://pkg.go.dev/inet.af/netaddr#ParseIPPort?

sidecar/internal/util/util_test.go
```diff
@@ -0,0 +1,75 @@
+/*
+Copyright 2021 The Kubernetes-CSI-Addons Authors.
```
Madhu-1, line 2: 2022

sidecar/internal/util/util_test.go
```diff
@@ -0,0 +1,75 @@
...
+			name: "valid case 1",
+			args: args{
+				endpoint: "192.168.61.228:8080",
+			},
+			wantErr: false,
+		},
+		{
+			name: "valid case 2",
+			args: args{
+				endpoint: "172.0.61.22:8000",
```
nixpanic, line 42: What is the difference with the 1st case?

Do not use 172.0.x.x as example, use an address from the ranges listed at https://datatracker.ietf.org/doc/html/rfc1918#section-3

sidecar/internal/util/util.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"fmt"
+	"net"
+	"strconv"
+	"strings"
+)
+
+// ValidateControllerEndpoint validate endpoint to make sure it contains
+// valid ip address and port number.
+func ValidateControllerEndpoint(endpoint *string) error {
```
nixpanic, line 28: no need to pass a pointer to the string, Go commonly passes strings directly.

nixpanic (changes requested):
Just a few nits, looks good otherwise.

sidecar/internal/util/util.go
Rakshith-R: > possible to use https://pkg.go.dev/inet.af/netaddr#ParseIPPort?

https://github.com/inetaf/netaddr

I was not sure, since they use a BSD-3 and the project is in very early stages. 
If that's fine, then I'll change it to use this function.

Madhu-1: >This package is mature, optimized, and used heavily in production at Tailscale. However, API stability is not yet guaranteed.

it's not in the early stage but stability is not guaranteed. am fine with this change for now.

sidecar/internal/util/util.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	rawIP, rawPort := splits[0], splits[1]
+
+	ip := net.ParseIP(rawIP)
+	if ip == nil {
+		return fmt.Errorf("invalid ip address %q in endpoint %q", rawIP, endpoint)
+	}
+
+	_, err := strconv.ParseUint(splits[1], 10, 16)
```
Madhu-1, line 42: max port number validation is required?

sidecar/internal/util/util.go
```diff
@@ -0,0 +1,48 @@
...
+func ValidateControllerEndpoint(endpoint string) error {
+	separator := ":"
+	splits := strings.Split(endpoint, separator)
+	if len(splits) != 2 {
+		return fmt.Errorf("invalid endpoint %q, expected format '<ip>:<port>'", endpoint)
+	}
+
+	rawIP, rawPort := splits[0], splits[1]
+
+	ip := net.ParseIP(rawIP)
```
Madhu-1, line 37: no need for Ip variable just check `net.ParseIP(rawIP) == nil`?

sidecar/internal/util/util.go
Rakshith-R: > max port number validation is required?

I was not sure about that, but while creating pod and exposing a port, the port is validated, so I guess we don't need to have that check here.

sidecar/internal/util/util.go
```diff
@@ -0,0 +1,47 @@
...
+	"fmt"
+	"net"
+	"strconv"
+	"strings"
+)
+
+// ValidateControllerEndpoint validate endpoint to make sure it contains
+// valid ip address and port number.
+func ValidateControllerEndpoint(endpoint string) error {
+	separator := ":"
```
nixpanic, line 29: Actually, this will fail on IPv6 addresses. Those should definitely be supported!

nixpanic (changes requested):

Madhu-1 (approved):

sidecar/internal/csiaddonsnode/csiaddonsnode_test.go
```diff
@@ -133,7 +131,7 @@ func Test_getCSIAddonsNode(t *testing.T) {
 				Spec: csiaddonsv1alpha1.CSIAddonsNodeSpec{
 					Driver: csiaddonsv1alpha1.CSIAddonsNodeDriver{
 						Name:     "csi.example.com",
-						EndPoint: podIP + ":" + "8060",
+						EndPoint: "192.168.61.228:8060",
```
nixpanic, line 134: please add a test with an IPv6 address

sidecar/internal/util/util.go
Rakshith-R: > Actually, this will fail on IPv6 addresses. Those should definitely be supported!

modified to handle IPv6 too, and added some tests, PTAL @nixpanic @Madhu-1

sidecar/internal/util/util_test.go
```diff
@@ -0,0 +1,103 @@
...
+			name: "IPv4 valid case",
+			args: args{
+				endpoint: "192.168.61.228:8080",
+			},
+			wantErr: false,
+		},
+		{
+			name: "IPv6 valid case",
+			args: args{
+				endpoint: "2001:0db8:3c4d:0015:0000:0000:1a2f:1a2b:8080",
```
nixpanic, line 42: Also include the notation that is is like `[2001:0db8::1]:8080`, see https://en.wikipedia.org/wiki/IPv6_address#Literal_IPv6_addresses_in_network_resource_identifiers

nixpanic (changes requested):

sidecar/internal/util/util_test.go
Rakshith-R: > Also include the notation that is is like `[2001:0db8::1]:8080`, see https://en.wikipedia.org/wiki/IPv6_address#Literal_IPv6_addresses_in_network_resource_identifiers

This will fail, but i expect PodIP to be of the form `"2001:0db8:3c4d:0015:0000:0000:1a2f:1a2b"`, so we should be fine

nixpanic: If the port is included, the IPv6 address should be within `[...]` ? I do not think there is a notation without `[...]` that can include the port.

sidecar/internal/csiaddonsnode/csiaddonsnode_test.go
```diff
@@ -111,9 +109,40 @@ func Test_getCSIAddonsNode(t *testing.T) {
 		},
 		{
 			name: "Test 2",
+			args: args{
+				driverName: "example.com",
+				endpoint:   "2001:0db8:3c4d:0015:0000:0000:1a2f:1a2b:8080",
```
nixpanic, line 114: I do not think this is a valid IPv6+port notation

sidecar/internal/util/util_test.go
nixpanic: Maybe it is required to have separate arguments for IP-address and port?

Or the endpoint needs to be a full URI, like `tcp://<IP>:<PORT>`, where `<IP>` is a valid IPv4 or IPv6 address? That way standard `net` utilities can parse it.

But with the full URI syntax, it may be difficult to create the URI in a deployment/daemonset?

nixpanic (approved):
thanks!

yati1998 (approved):
LGTM

Issue #48, opened 2022-01-06
In case the CSIAddonsNode CRD is not available, the sidcar should not abort but retry


nixpanic:
When the sidecar fails to create the CSIAddonsNode, it fails, and causes a deployment/deamonset of the CSI-driver to fail.

It is better to report the error of the missing CRD and the failure, and backoff-retry until the CRD becomes available (might be never). This will allow deploying the sidecar by default, even when users do not deploy the controller and the CRDs.

Madhu-1:
This is applicable to the controllers also. if any of the CRD is missing the operator should not crash instead of that the operator should be try with back-off

Rakshith-R:
> When the sidecar fails to create the CSIAddonsNode, it fails, and causes a deployment/deamonset of the CSI-driver to fail.


This was partially intentional, since the user may not notice missing CRDs until they look inside sidecar logs but sidecar being crashloop backoff would make a bit more prominent. Note: only sidecar will be crashloopbackoff.

> It is better to report the error of the missing CRD and the failure, and backoff-retry until the CRD becomes available (might be never). This will allow deploying the sidecar by default, even when users do not deploy the controller and the CRDs.
 
 I am okay with backoff-retry method too. 
 
 
---
> This is applicable to the controllers also.

I have not seen any of the controllers wait explicitly for core CRDs which it is supposed to be watching with backoff-retry.
(volrep being exception, I inferred so since it is part of cephcsi deployment and unique)

>if any of the CRD is missing the operator should not crash instead of that the operator should be try with back-off

controller will not crash but will internally retry 
example  with only reclaimspacejob CRD missing
```
W0107 05:20:10.003375       1 reflector.go:324] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:250: failed to list *v1alpha1.ReclaimSpaceJob: the server could not find the requested resource (get reclaimspacejobs.csiaddons.openshift.io)
E0107 05:20:10.003409       1 reflector.go:138] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:250: Failed to watch *v1alpha1.ReclaimSpaceJob: failed to list *v1alpha1.ReclaimSpaceJob: the server could not find the requested resource (get reclaimspacejobs.csiaddons.openshift.io)
[...]

PR #51, opened 2022-01-07
sidecar: retry creating CSIAddonsNode on failure

nixpanic:
In case creating the CSIAddonsNode for a Pod fails, a retry is done.
There are several causes that can cause failures, these seem to be the
most common ones:

 - User cannot create resource "csiaddonsnodes" in API group "csiaddons.openshift.io"
 - no kind "CSIAddonsNode" is registered for the internal version of group "csiaddons.openshift.io"

In the above cases, either the RBAC is limiting access to the API, or
the CRD is not available in the Kubernetes cluster.

Updates: #48

### DIFF
--- a/sidecar/internal/csiaddonsnode/csiaddonsnode.go
+++ b/sidecar/internal/csiaddonsnode/csiaddonsnode.go
@@ -20,11 +20,11 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"strings"
 	"time"
 
 	csiaddonsv1alpha1 "github.com/csi-addons/kubernetes-csi-addons/api/v1alpha1"
 
+	"k8s.io/apimachinery/pkg/api/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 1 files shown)


sidecar/internal/csiaddonsnode/csiaddonsnode.go
```diff
@@ -21,31 +21,63 @@ import (
...
+
+	// nodeCreationRetry is the delay for calling newCSIAddonsNode after a
+	// failure.
+	nodeCreationRetry = time.Minute * 5
 )
 
 // Deploy creates CSIAddonsNode custom resource with all required information.
+// When information to create the CSIAddonsNode is missing, an error will be
+// returned immediately. If creating the CSIAddonsNode in the Kubernetes
+// cluster fails (missing CRD, object already exists, ...), an error will be
```
Rakshith-R, line 51: if `object already exists,` , then we should just continue without error or retries ?

nixpanic: If the object exists, it may not be correct (different IP address?), so the container should probably be blocked from starting completely until that is resolved.

Maybe there needs to be a check, that in case the object exists, it is updated instead of created? But that can be a different PR.

Rakshith-R: No, ip address will be unique per pod and will not change during pod lifetime.

If object already exists, then I don't think any parameter will have changed so we just ignore this error and continue.

And in this case, I don't think retrying will help since object will continue to exist until pod is deleted or someone manually deletes the object.

nixpanic: Ah, there is a check for that already. I'll clean that up and improve the comment.

Rakshith-R (approved):

Madhu-1 (approved):

PR #55, opened 2022-01-10
controller: add CSIAddonsNode controller

Rakshith-R:
-    controller: add CSIAddonsNode controller
    
    This commit adds code for CSIAddonsNode controller
    which watches CSIAddonsNode CR, establishes connections,
    maintains connection pool and andles deletion.
    
    
-    util: add util ContainsInSlice & RemoveFromSlice funcs

### DIFF
--- a/controllers/csiaddonsnode_controller.go
+++ b/controllers/csiaddonsnode_controller.go
@@ -18,19 +18,30 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 
+	csiaddonsv1alpha1 "github.com/csi-addons/kubernetes-csi-addons/api/v1alpha1"
+	"github.com/csi-addons/kubernetes-csi-addons/internal/connection"
+	"github.com/csi-addons/kubernetes-csi-addons/internal/util"
+
+	"google.golang.org/grpc/status"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
 
-	csiaddonsv1alpha1 "github.com/csi-addons/kubernetes-csi-addons/api/v1alpha1"
+const (
+	csiAddonsNodeFinalizer = "csiaddons.openshift.io"
 )
 
 // CSIAddonsNodeReconciler reconciles a CSIAddonsNode object
 type CSIAddonsNodeReconciler struct {
 	client.Client
-	Scheme *runtime.Scheme
+	Scheme   *runtime.Scheme
+	ConnPool *connection.ConnectionPool
 }
 
 //+kubebuilder:rbac:groups=csiaddons.openshift.io,resources=csiaddonsnodes,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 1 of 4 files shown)


internal/util/strings.go
```diff
@@ -0,0 +1,40 @@
+/*
+Copyright 2021 The Kubernetes-CSI-Addons Authors.
```
Madhu-1, line 2: 2022

internal/util/strings.go
```diff
@@ -0,0 +1,40 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+// ConstainsInSlice find given string in given slice
```
Madhu-1, line 19: end the comment with `.` and add unit testing

controllers/csiaddonsnode_controller.go
```diff
@@ -39,17 +50,98 @@ type CSIAddonsNodeReconciler struct {
...
+		ctx,
+		endPoint,
+		nodeID,
+		driverName)
+	if err != nil {
+		s, _ := status.FromError(err)
+		logger.Error(s.Err(), "Failed to establish connection with sidecar")
+
+		instance.Status.State = csiaddonsv1alpha1.CSIAddonsNodeStateFailed
+		instance.Status.Message = fmt.Sprintf("Failed to establish connection with sidecar: %v", s.Message())
```
Madhu-1, line 123: s.Message exists only if `status.FromError` can convert the error to GRPC. can you check the bool value from `status.FromError` and access s.Message?

controllers/csiaddonsnode_controller.go
```diff
@@ -39,17 +50,98 @@ type CSIAddonsNodeReconciler struct {
...
+		if statusErr := r.Client.Status().Update(ctx, instance); statusErr != nil {
+			logger.Error(err, "Failed to update status")
```
Madhu-1, line 125: log statusErr?

controllers/csiaddonsnode_controller.go
```diff
@@ -58,5 +150,11 @@ func (r *CSIAddonsNodeReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 func (r *CSIAddonsNodeReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&csiaddonsv1alpha1.CSIAddonsNode{}).
+		WithEventFilter(predicate.GenerationChangedPredicate{}).
 		Complete(r)
 }
+
+// createKey returns "namespace/name" string to be used as key for connection pool.
+func (r *CSIAddonsNodeReconciler) creatKey(namespace, name string) string {
```
Madhu-1, line 158: `creatKey` to `createKey`

controllers/csiaddonsnode_controller.go
```diff
@@ -39,17 +50,98 @@ type CSIAddonsNodeReconciler struct {
...
+
+	logger.Info("Connecting to sidecar")
+	newCon, err := connection.NewConnection(
+		ctx,
+		endPoint,
+		nodeID,
+		driverName)
+	if err != nil {
+		s, _ := status.FromError(err)
+		logger.Error(s.Err(), "Failed to establish connection with sidecar")
```
Madhu-1, line 120: can you check the bool value from status.FromError and access s.Err()

controllers/csiaddonsnode_controller.go
```diff
@@ -39,17 +50,98 @@ type CSIAddonsNodeReconciler struct {
...
-// perform operations to make the cluster state reflect the state specified by
-// the user.
-//
-// For more details, check Reconcile and its Result here:
-// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.10.0/pkg/reconcile
 func (r *CSIAddonsNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	_ = log.FromContext(ctx)
+	logger := log.FromContext(ctx)
+
+	logger.Info("Starting Reconcile")
```
Madhu-1, line 56: not a useful log remove it?

controllers/csiaddonsnode_controller.go
```diff
@@ -39,17 +50,98 @@ type CSIAddonsNodeReconciler struct {
...
+	err := r.Client.Get(ctx, req.NamespacedName, instance)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// Request object not found, could have been deleted after reconcile request.
+			logger.Info("CSIAddonsNode resource not found")
+
+			return ctrl.Result{}, nil
+		}
+
+		// Error reading the object - requeue the request.
```
Madhu-1, line 68: generic comment not required here

controllers/csiaddonsnode_controller.go
```diff
@@ -39,17 +50,98 @@ type CSIAddonsNodeReconciler struct {
...
+
+	logger.Info("Starting Reconcile")
+	// Fetch CSIAddonsNode instance
+	instance := &csiaddonsv1alpha1.CSIAddonsNode{}
+	err := r.Client.Get(ctx, req.NamespacedName, instance)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// Request object not found, could have been deleted after reconcile request.
+			logger.Info("CSIAddonsNode resource not found")
+
```
Madhu-1, line 64: if the object doesn't exist check if the connection exists in the pool and remove it?

controllers/csiaddonsnode_controller.go
```diff
@@ -39,17 +50,98 @@ type CSIAddonsNodeReconciler struct {
...
+	logger.Info("Added connection to connection pool")
+
+	instance.Status.State = csiaddonsv1alpha1.CSIAddonsNodeStateConnected
+	instance.Status.Message = "Successfully established connection with sidecar"
+	if err = r.Client.Status().Update(ctx, instance); err != nil {
+		logger.Error(err, "Failed to update status")
+
+		return ctrl.Result{}, err
+	}
+	logger.Info("Exiting Reconcile")
```
Madhu-1, line 144: not a useful log remove it?

controllers/csiaddonsnode_controller.go
```diff
@@ -39,17 +50,98 @@ type CSIAddonsNodeReconciler struct {
...
+		"Driver Name", driverName)
+
+	if !instance.DeletionTimestamp.IsZero() {
+		logger.Info("Deleting connection")
+		r.ConnPool.Delete(key)
+
+		if util.ContainsInSlice(instance.Finalizers, csiAddonsNodeFinalizer) {
+			logger.Info("Removing finalizer")
+
+			instance.Finalizers = util.RemoveFromSlice(instance.Finalizers, csiAddonsNodeFinalizer)
```
Madhu-1, line 89: how this CR will be deleted? have we documented creation/deletion of the CSiAddonsNode CR workflow?

controllers/csiaddonsnode_controller.go
Rakshith-R: > if the object doesn't exist check if the connection exists in the pool and remove it?

The object cannot get deleted until finalizer is removed, so we should be fine here.

controllers/csiaddonsnode_controller.go
Rakshith-R: > how this CR will be deleted?

CSIAddonsNode CR created by sidecar will have ownerRef set to the pod itself, so when the pod is deleted, CR will get deleted too.
https://github.com/csi-addons/kubernetes-csi-addons/blob/main/sidecar/internal/csiaddonsnode/csiaddonsnode.go#L107-L114

> have we documented creation/deletion of the CSiAddonsNode CR workflow?

I guess it's not, https://github.com/csi-addons/kubernetes-csi-addons/blob/main/docs/csiaddonsnode.md
@yati1998 , can update the docs with more details ? 
(possibly when adding deployment guide)

internal/util/strings.go
```diff
@@ -0,0 +1,40 @@
...
+		if item == s {
+			return true
+		}
+	}
+
+	return false
+}
+
+// RemoveFromSlice removes given string from given slice
+func RemoveFromSlice(slice []string, s string) (result []string) {
```
nixpanic, line 31: please do not use these implicit return variables, we don't use it anywhere else either. declare `result` and return it.

controllers/csiaddonsnode_controller.go
```diff
@@ -39,17 +50,98 @@ type CSIAddonsNodeReconciler struct {
...
-// TODO(user): Modify the Reconcile function to compare the state specified by
-// the CSIAddonsNode object against the actual cluster state, and then
-// perform operations to make the cluster state reflect the state specified by
-// the user.
-//
-// For more details, check Reconcile and its Result here:
-// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.10.0/pkg/reconcile
 func (r *CSIAddonsNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	_ = log.FromContext(ctx)
+	logger := log.FromContext(ctx)
```
nixpanic, line 54: this function now takes ~100 lines, consider making it simpler by adding helper functions

Rakshith-R: > this function now takes ~100 lines, consider making it simpler by adding helper functions

its about 80 lines now, 
creating helper functions will need a lot of parameters(logger,instance,client...) to be passed to it, which makes it more ugly actually. 
(had tried it :sweat: )

On another hand, single flow will make it more readable

nixpanic: lines 51 to 147... and that is without a lot of comments. I expect this function to grow over time, so some cleanup would be good.

It should be possible to handle different events in separate functions. If there are many parameters that need passing, consider adding a type like `CSIAddonsNodeOperation` containing the logger, client and other members. It can then have methods like `removeFinalizer()`,  `connect()` and so on.

Rakshith-R:
opened https://github.com/csi-addons/kubernetes-csi-addons/pull/58 
Was not able to push to this pr since mergifyio deleted branch while experimenting here https://github.com/Rakshith-R/kubernetes-csi-addons/pull/5 
:disappointed:

PR #58, opened 2022-01-10
controller: add CSIAddonsNode controller

Rakshith-R:
-    controller: add CSIAddonsNode controller
    
    This commit adds code for CSIAddonsNode controller
    which watches CSIAddonsNode CR, establishes connections,
    maintains connection pool and andles deletion.

-    util: add util ContainsInSlice & RemoveFromSlice funcs

-    util: add GetErrorMessage() helper function
    
    This commit adds GetErrorMessage helper function
    to get error message without grpc error codes.


controllers/csiaddonsnode_controller.go
```diff
@@ -39,17 +49,85 @@ type CSIAddonsNodeReconciler struct {
...
 
-	// TODO(user): your logic here
+		return ctrl.Result{}, nil
+	}
+
+	if !util.ContainsInSlice(instance.Finalizers, csiAddonsNodeFinalizer) {
+		logger.Info("Adding finalizer")
+
+		instance.Finalizers = append(instance.Finalizers, csiAddonsNodeFinalizer)
+		if err = r.Client.Update(ctx, instance); err != nil {
```
nixpanic, line 97: I think using `Patch()` is better for most cases?

no need to place the assignment and check on the same line, split it like done elsewhere and stick to the same coding convention everywhere

controllers/csiaddonsnode_controller.go
```diff
@@ -18,19 +18,29 @@ package controllers
...
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
 
-	csiaddonsv1alpha1 "github.com/csi-addons/kubernetes-csi-addons/api/v1alpha1"
+const (
+	csiAddonsNodeFinalizer = "csiaddons.openshift.io"
```
nixpanic, line 36: The contents looks generic, can is be shared with other controllers in this package? Place it in a file containing other shared constants and functions maybe?

Yuggupta27, line 36: Name the finalizer `csiaddons.openshift.io/operation`?
For example for networkfence it can be `csiaddons.openshift.io/network-fence`
Wdyt?

controllers/csiaddonsnode_controller.go
Rakshith-R: > I think using `Patch()` is better for most cases?

The examples I have seen use Update
https://github.com/csi-addons/volume-replication-operator/blob/05f25d5bffac00533e6acfa12709991d43336337/controllers/finalizers.go#L40
https://github.com/red-hat-storage/odf-operator/blob/0d5dd3ad5248c1dc3df62a9425bf375bddc77774/controllers/storagesystem_controller.go#L126
https://github.com/red-hat-storage/ocs-operator/blob/7438d205e74a63de1ccf52c2902d38923c5e3f62/controllers/storagecluster/reconcile.go#L309

> no need to place the assignment and check on the same line, split it like done elsewhere and stick to the same coding convention everywhere

I believe it is cleaner to user shorthand when only error is being returned, looks cleaner easier to read for me :sweat_smile:

internal/util/strings_test.go
```diff
@@ -0,0 +1,110 @@
...
+			args: args{
+				slice: []string{"hello", "hi", "hey"},
+				s:     "bye",
+			},
+			wantResult: []string{"hello", "hi", "hey"},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if gotResult := RemoveFromSlice(tt.args.slice, tt.args.s); !reflect.DeepEqual(gotResult, tt.wantResult) {
```
nixpanic, line 105: you could use [testify/assert](https://pkg.go.dev/github.com/stretchr/testify/assert) to simplify the code:

```go
r := RemoveFromSlice(tt.args.slice, tt.args.s)
assert.Equal(t, tt.wantResult, r)
```

cmd/manager/main.go
```diff
@@ -78,13 +79,17 @@ func main() {
 		os.Exit(1)
 	}
 
+	ConnPool := connection.NewConnectionPool()
```
nixpanic, line 82: use a lowecase variable name

controllers/csiaddonsnode_controller.go
```diff
@@ -39,17 +50,65 @@ type CSIAddonsNodeReconciler struct {
...
+	// Fetch CSIAddonsNode instance
+	instance := &csiaddonsv1alpha1.CSIAddonsNode{}
+	if err := r.Client.Get(ctx, req.NamespacedName, instance); err != nil {
+		if errors.IsNotFound(err) {
+			// Request object not found, could have been deleted after reconcile request.
+			logger.Info("CSIAddonsNode resource not found")
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
```
nixpanic, line 65: add an empty line after closing `}`

controllers/csiaddonsnode_controller.go
```diff
@@ -58,5 +117,50 @@ func (r *CSIAddonsNodeReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
 		For(&csiaddonsv1alpha1.CSIAddonsNode{}).
+		WithEventFilter(predicate.GenerationChangedPredicate{}).
 		Complete(r)
 }
+
+// addFinalizer adds finalizer to object instance if it is not present.
+func (r *CSIAddonsNodeReconciler) addFinalizer(
+	ctx context.Context,
+	instance *csiaddonsv1alpha1.CSIAddonsNode,
+	logger *logr.Logger) error {
```
nixpanic, line 128: context and logger are common arguments for functions of CSIAddonsNodeReconciler, use those as 1st parameters in the list. Arguments that change depending on the function (like CSIAddonsNode), pass those afterwards.

I'm not very fond of a name like `instance` or `object`. Better use `node` or something that is a little more descriptive.

controllers/csiaddonsnode_controller.go
```diff
@@ -39,17 +50,65 @@ type CSIAddonsNodeReconciler struct {
...
 func (r *CSIAddonsNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	_ = log.FromContext(ctx)
+	logger := log.FromContext(ctx)
+
+	// Fetch CSIAddonsNode instance
+	instance := &csiaddonsv1alpha1.CSIAddonsNode{}
+	if err := r.Client.Get(ctx, req.NamespacedName, instance); err != nil {
+		if errors.IsNotFound(err) {
+			// Request object not found, could have been deleted after reconcile request.
+			logger.Info("CSIAddonsNode resource not found")
```
nixpanic, line 61: include details about the CSIAddonsNode that could not be found, like namespace/name

controllers/csiaddonsnode_controller.go
```diff
@@ -39,17 +50,65 @@ type CSIAddonsNodeReconciler struct {
...
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+	nodeID := instance.Spec.Driver.NodeID
```
nixpanic, line 66: Consider a helper like `newLogger(instance)`. Add validation of the used fields/structs (`instance.Spec != nil` and so on) there.

controllers/csiaddonsnode_controller.go
```diff
@@ -39,17 +50,65 @@ type CSIAddonsNodeReconciler struct {
...
+	driverName := instance.Spec.Driver.Name
+	endPoint := instance.Spec.Driver.EndPoint
+	key := r.createKey(instance.Namespace, instance.Name)
+	logger = logger.WithValues("NodeID", nodeID, "EndPoint", endPoint, "Driver Name", driverName)
+
+	if !instance.DeletionTimestamp.IsZero() {
```
nixpanic, line 72: a comment like `if deletion timestap is set, the CSIAddonsNode is getting deleted, cleanup resources` (specifying the resources, or calling a helper with suitable comments) would be nice

controllers/csiaddonsnode_controller.go
```diff
@@ -39,17 +50,65 @@ type CSIAddonsNodeReconciler struct {
...
+		return ctrl.Result{}, err
+	}
+
+	logger.Info("Successfully connected to sidecar")
+	r.ConnPool.Put(key, newCon)
+	logger.Info("Added connection to connection pool")
+
+	instance.Status.State = csiaddonsv1alpha1.CSIAddonsNodeStateConnected
+	instance.Status.Message = "Successfully established connection with sidecar"
+	if err = r.Client.Status().Update(ctx, instance); err != nil {
```
nixpanic, line 107: use `Patch()` instead of `Update()`?

controllers/csiaddonsnode_controller.go
```diff
@@ -58,5 +117,50 @@ func (r *CSIAddonsNodeReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+			logger.Error(err, "Failed to remove finalizer")
+
+			return err
+		}
+	}
+	return nil
+}
+
+// createKey returns "namespace/name" string to be used as key for connection pool.
+func (r *CSIAddonsNodeReconciler) createKey(namespace, name string) string {
```
nixpanic, line 164: consider using CSIAddonsNode as a single argument

nixpanic (commented):
just a few minor changes for improving maintainability

controllers/csiaddonsnode_controller.go
Rakshith-R: `logger := log.FromContext(ctx)`
This will provide the name namespace into the log.

controllers/csiaddonsnode_controller.go
Rakshith-R: > Consider a helper like `newLogger(instance)`. 

`logger = logger.WithValues("NodeID", nodeID, "EndPoint", endPoint, "Driver Name", driverName)`
just a line, other variables will be used in following lines.

>Add validation of the used fields/structs (`instance.Spec != nil` and so on) there.

I really think validation should be a part of webhooks/CRD validation. I've added a TODO for now.

controllers/csiaddonsnode_controller.go
nixpanic: I do not think it is easier to read. The check `err != nil` (or something else) is less obvious as it is moved completely to the far right if the line.

internal/util/strings_test.go
```diff
@@ -0,0 +1,109 @@
...
+			args: args{
+				slice: []string{"hello", "hi", "hey"},
+				s:     "bye",
+			},
+			wantResult: []string{"hello", "hi", "hey"},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			assert.Equal(t, RemoveFromSlice(tt.args.slice, tt.args.s), tt.wantResult)
```
nixpanic, line 106: nit: the format is `Equal(t, <expected>, <result>)`, so the other way around

internal/util/strings_test.go
```diff
@@ -0,0 +1,109 @@
...
+				slice: []string{},
+				s:     "hello",
+			},
+			want: false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := ContainsInSlice(tt.args.slice, tt.args.s); got != tt.want {
+				t.Errorf("ContainsInSlice() = %v, want %v", got, tt.want)
```
nixpanic, line 71: you cleaned up the other test, you can use `assert.Equal()` here too.

cmd/manager/main.go
```diff
@@ -33,6 +33,7 @@ import (
 
 	csiaddonsv1alpha1 "github.com/csi-addons/kubernetes-csi-addons/api/v1alpha1"
 	"github.com/csi-addons/kubernetes-csi-addons/controllers"
+	"github.com/csi-addons/kubernetes-csi-addons/internal/connection"
```
nixpanic, line 36: the commit message contains a typo (andles)

controllers/csiaddonsnode_controller.go
```diff
@@ -39,17 +50,69 @@ type CSIAddonsNodeReconciler struct {
...
+		}
+		return ctrl.Result{}, nil
+	}
+
+	if err := r.addFinalizer(ctx, &logger, csiAddonsNode); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	logger.Info("Connecting to sidecar")
+	newCon, err := connection.NewConnection(ctx, endPoint, nodeID, driverName)
```
nixpanic, line 90: try to be consistent with naming variables `connPool` (2x 'n'), `newCon` (single 'n')

controllers/csiaddonsnode_controller.go
```diff
@@ -39,17 +50,69 @@ type CSIAddonsNodeReconciler struct {
...
+	if err != nil {
+		logger.Error(err, "Failed to establish connection with sidecar")
+
+		errMessage := util.GetErrorMessage(err)
+		csiAddonsNode.Status.State = csiaddonsv1alpha1.CSIAddonsNodeStateFailed
+		csiAddonsNode.Status.Message = fmt.Sprintf("Failed to establish connection with sidecar: %v", errMessage)
+		if statusErr := r.Client.Status().Update(ctx, csiAddonsNode); statusErr != nil {
```
nixpanic, line 97: Still wondering about the `Update()` vs `Patch()` call

controllers/csiaddonsnode_controller.go
```diff
@@ -58,5 +121,50 @@ func (r *CSIAddonsNodeReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+	if util.ContainsInSlice(csiAddonsNode.Finalizers, csiAddonsNodeFinalizer) {
+		logger.Info("Removing finalizer")
+
+		csiAddonsNode.Finalizers = util.RemoveFromSlice(csiAddonsNode.Finalizers, csiAddonsNodeFinalizer)
+		if err := r.Client.Update(ctx, csiAddonsNode); err != nil {
+			logger.Error(err, "Failed to remove finalizer")
+
+			return err
+		}
+	}
```
nixpanic, line 163: need an empty line here?

controllers/csiaddonsnode_controller.go
```diff
@@ -58,5 +121,50 @@ func (r *CSIAddonsNodeReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 func (r *CSIAddonsNodeReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&csiaddonsv1alpha1.CSIAddonsNode{}).
+		WithEventFilter(predicate.GenerationChangedPredicate{}).
 		Complete(r)
 }
+
+// addFinalizer adds finalizer to csiAddonsNode if it is not present.
+func (r *CSIAddonsNodeReconciler) addFinalizer(
```
Yuggupta27, line 129: With this, each controller will need its own `addFinalizer` function?

- If yes, we can leave it as it is and define separate functions in each controller.

- But if we want a generic one, better to keep in a new `finalizers.go`

Thoughts? @Rakshith-R @nixpanic @Madhu-1

controllers/csiaddonsnode_controller.go
Rakshith-R: > Still wondering about the `Update()` vs `Patch()` call

again most of the operators and controllers user `Update()`, I don't see any advantage of explicitly specifying what to patch when you can just update the entire object.
https://book.kubebuilder.io/cronjob-tutorial/controller-implementation.html#2-list-all-active-jobs-and-update-the-status

I didn't find any such examples of Patch() being used by controller/operator.

Madhu-1: common practice is to use Update for resource update ex:- 
https://github.com/operator-framework/operator-sdk-samples/blob/caf90347ff2e40ae9511d651d6fef4bb42c891af/go/memcached-operator/controllers/memcached_controller.go#L114-L118

controllers/csiaddonsnode_controller.go
```diff
@@ -18,19 +18,30 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 
+	csiaddonsv1alpha1 "github.com/csi-addons/kubernetes-csi-addons/api/v1alpha1"
+	"github.com/csi-addons/kubernetes-csi-addons/internal/connection"
+	"github.com/csi-addons/kubernetes-csi-addons/internal/util"
+	"github.com/go-logr/logr"
```
Madhu-1, line 26: logr is third party library separation is required here with new line

controllers/csiaddonsnode_controller.go
```diff
@@ -18,19 +18,30 @@ package controllers
...
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
 
-	csiaddonsv1alpha1 "github.com/csi-addons/kubernetes-csi-addons/api/v1alpha1"
+const (
+	csiAddonsNodeFinalizer = "csiaddons.openshift.io/csiaddonsnode"
```
Madhu-1, line 37: use `GroupVersion.Group+"csiaddonsNode"` to avoid duplicate of `csiaddons.openshift.io` if something changes in future it we don't need to changes things in many places.

controllers/csiaddonsnode_controller.go
```diff
@@ -39,17 +50,69 @@ type CSIAddonsNodeReconciler struct {
...
+
+	if !csiAddonsNode.DeletionTimestamp.IsZero() {
+		// if deletion timestap is set, the CSIAddonsNode is getting deleted,
+		// delete connections and remove finalizer.
+		logger.Info("Deleting connection")
+		r.ConnPool.Delete(key)
+		if err := r.removeFinalizer(ctx, &logger, csiAddonsNode); err != nil {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{}, nil
```
Madhu-1, line 82: as we are not logging anything err check is required here? just return `ctrl.Result{},err`?

controllers/csiaddonsnode_controller.go
```diff
@@ -58,5 +121,50 @@ func (r *CSIAddonsNodeReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+			logger.Error(err, "Failed to remove finalizer")
+
+			return err
+		}
+	}
+	return nil
+}
+
+// createKey returns "namespace/name" string to be used as key for connection pool.
+func (r *CSIAddonsNodeReconciler) createKey(csiAddonsNode *csiaddonsv1alpha1.CSIAddonsNode) string {
```
Madhu-1, line 168: this can be a function not a method as `CSIAddonsNodeReconciler` is not used.

controllers/csiaddonsnode_controller.go
nixpanic: input needs to be validated _everywhere_

controllers/csiaddonsnode_controller.go
Rakshith-R: > this can be a function not a method as `CSIAddonsNodeReconciler` is not used.

removed as it was only used in one place

config/samples/csiaddons_v1alpha1_reclaimspacejob.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: csiaddons.openshift.io/v1alpha1
 kind: ReclaimSpaceJob
 metadata:
-  name: reclaimspacejob-sample
+  name: reclaimspacejob-sample1
```
Yuggupta27, line 4: Unintentional?

Yuggupta27 (changes requested):

config/samples/csiaddons_v1alpha1_reclaimspacejob.yaml
Rakshith-R: > Unintentional?

yup, fixed now

controllers/csiaddonsnode_controller.go
```diff
@@ -18,19 +18,31 @@ package controllers
...
+	apiErrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
 
-	csiaddonsv1alpha1 "github.com/csi-addons/kubernetes-csi-addons/api/v1alpha1"
+var (
```
Madhu-1, line 37: const instead of var?

controllers/csiaddonsnode_controller.go
```diff
@@ -39,17 +51,87 @@ type CSIAddonsNodeReconciler struct {
...
+
+		// invalid parameter, do not requeue
+		return ctrl.Result{}, nil
+	}
+
+	nodeID := csiAddonsNode.Spec.Driver.NodeID
+	driverName := csiAddonsNode.Spec.Driver.Name
+	endPoint := csiAddonsNode.Spec.Driver.EndPoint
+	key := csiAddonsNode.Namespace + "/" + csiAddonsNode.Name
+	logger = logger.WithValues("NodeID", nodeID, "EndPoint", endPoint, "Driver Name", driverName)
```
Madhu-1, line 90: `Driver Name` to `DriverName`

controllers/csiaddonsnode_controller.go
```diff
@@ -58,5 +140,58 @@ func (r *CSIAddonsNodeReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+		}
+	}
+
+	return nil
+}
+
+// validateCSIAddonsNodeSpec validates if Name and Endpoint are not empty.
+func validateCSIAddonsNodeSpec(csiaddonsnode *csiaddonsv1alpha1.CSIAddonsNode) error {
+	if csiaddonsnode.Spec.Driver.Name == "" {
+		return errors.New("required parameter 'Name' missing in CSIAddonsNode.Spec.Driver empty")
```
Madhu-1, line 190: empty or object? the error message is not meaningful

controllers/csiaddonsnode_controller.go
```diff
@@ -18,19 +18,31 @@ package controllers
...
+var (
+	csiAddonsNodeFinalizer = csiaddonsv1alpha1.GroupVersion.Group + "/csiaddonsnode"
 )
 
 // CSIAddonsNodeReconciler reconciles a CSIAddonsNode object
 type CSIAddonsNodeReconciler struct {
 	client.Client
-	Scheme *runtime.Scheme
+	Scheme   *runtime.Scheme
+	ConnPool *connection.ConnectionPool
```
Madhu-1, line 45: all connections are stored in this CSIAddonsNodeReconciler struct how consumers will use it as it's not stored in some global variable?

controllers/csiaddonsnode_controller.go
Rakshith-R: > const instead of var?

not possible, since we are using another variable to infer the value.

controllers/csiaddonsnode_controller.go
Madhu-1: got this one as it's defined in main.go. Thanks @Rakshith-R

Madhu-1 (dismissed):

controllers/csiaddonsnode_controller.go
Rakshith-R: > all connections are stored in this CSIAddonsNodeReconciler struct how consumers will use it as it's not stored in some global variable?

variable will be created in main.go and pointer will be passed to each controller as part of their struct(this change will be in respective controller prs).

controllers/csiaddonsnode_controller.go
```diff
@@ -18,19 +18,31 @@ package controllers
...
 	"context"
+	"errors"
+	"fmt"
 
+	csiaddonsv1alpha1 "github.com/csi-addons/kubernetes-csi-addons/api/v1alpha1"
+	"github.com/csi-addons/kubernetes-csi-addons/internal/connection"
+	"github.com/csi-addons/kubernetes-csi-addons/internal/util"
+
+	"github.com/go-logr/logr"
+	apiErrors "k8s.io/apimachinery/pkg/api/errors"
```
nixpanic, line 29: nit: convention is to use lowercase for imports

nixpanic (approved):

Yuggupta27 (dismissed):

controllers/csiaddonsnode_controller.go
Rakshith-R: fixed this too, PTAL

Madhu-1 (approved):

Yuggupta27 (approved):

nixpanic (approved):

controllers/csiaddonsnode_controller.go
```diff
@@ -39,17 +51,87 @@ type CSIAddonsNodeReconciler struct {
...
+		csiAddonsNode.Status.Message = fmt.Sprintf("Failed to validate CSIAddonsNode parameters: %v", err)
+		statusErr := r.Client.Status().Update(ctx, csiAddonsNode)
+		if statusErr != nil {
+			logger.Error(statusErr, "Failed to update status")
+
+			return ctrl.Result{}, statusErr
+		}
+
+		// invalid parameter, do not requeue
+		return ctrl.Result{}, nil
```
Yuggupta27, line 83: In case parameters are invalid/empty... do we not want to return an error?

suggested change `return ctrl.Result{}, err`
Thoughts?

Yuggupta27 (changes requested):

controllers/csiaddonsnode_controller.go
Rakshith-R: > In case parameters are invalid/empty... do we not want to return an error?
> 
> suggested change `return ctrl.Result{}, err` Thoughts?

it is intentional, we're logging the error and dequeue it. 
No use in reconciling invalid parameters again and again. 

If there is a change in parameter from outside, it'll be requeued for reconcile

Yuggupta27: Cool :+1:

Yuggupta27 (approved):

PR #59, opened 2022-01-10
sidecar: prevent failure and listen immediately

nixpanic:
When the sidecar starts, the following fatal error is logged:

    failed to create csiaddonsnode object: no kind "CSIAddonsNode" is registered for the internal version of group "csiaddons.openshift.io" in scheme "k8s.io/client-go/kubernetes/scheme/register.go:74"

After that, the sidecar restarts.

Even though the error is handled as fatal, the new CSIAddonsNode object
is created successfully. After the restart, the new CSIAddonsNode is
found, and the sidecar continues to startup successfully.

It seems that placing the result `Into(nil)` causes the failure, as
opposed to the documentation. Getting the error from the `Do()` response
is sufficient here, and prevents the failure of calling `Into()`.


yati1998 (approved):

Rakshith-R (approved):

Rakshith-R:
Thanks !

PR #61, opened 2022-01-11
tools: build `kustomize` from the `vendor/` directory

nixpanic:
By installing `kustomize` from the vendor directory, versions of the
tools can automatically be updated, just like other dependencies. It
also makes things easier to install the tools in environments that have
slow networking or are completely disconnected.


tools.go
```diff
@@ -0,0 +1,29 @@
+//go:build tools
+// +build tools
+
+/*
+Copyright 2021 The Kubernetes-CSI-Addons Authors.
```
Madhu-1, line 5: 2022

Makefile
```diff
@@ -43,11 +43,15 @@ help: ## Display this help.
...
-	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./cmd/... ./controllers/... ./sidecar/..." output:crd:artifacts:config=config/crd/bases
 
 .PHONY: generate
 generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
-	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
+	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./api/..."
+
+.PHONY: bundle
+bundle: kustomize operator-sdk
```
Madhu-1, line 53: can we have CI to validate kustomize/manifests/generate commands to make sure it's working and won't break in the future?

tools.go
```diff
@@ -0,0 +1,29 @@
...
+
+// These imports are used to maintain a copy of the used tools in the vendor/
+// directory. Versions of the tools can automatically be updated, just like
+// other dependencies. It also makes things easier to install the tools in
+// environments that have slow networking or are completely disconnected.
+
+package main
+
+import (
+	_ "sigs.k8s.io/kustomize/kustomize/v4"
```
Madhu-1, line 28: if kustomize is updated to v5 we need changes here also. will dependent bot takes care of it?

deepshikhaaa:
+1, It will be helpful for downstream builds too.

Makefile
nixpanic: I hope so, with this change it should be easier to write such jobs.

Actually, the `bundle` target is not intentionally included, I am only experimenting with that now.

tools.go
nixpanic: right!

tools.go
nixpanic: no, I do not think so. But The way it is currently done (version in the `Makefile`) will not even see minor updates.

Madhu-1 (approved):

Rakshith-R (approved):

PR #62, opened 2022-01-12
controller: modify ConnectionPool.Get() to return a copy

Rakshith-R:
This commit modifies ConnectionPool.Get() to take in
driverName and optional nodeID as parameters and now
returns a filtered copy of connections map.
This is done to avoid any deadlock issue that can occur
when unlock function may not be called in previous iteration
of the function.
Unit tests have been added and it now uses assert package.

- Further filtering using Capabilities seems a bit tricky, It can be done as a future enhancement.

### DIFF
--- a/internal/connection/connection_pool.go
+++ b/internal/connection/connection_pool.go
@@ -62,11 +62,21 @@ func (cp *ConnectionPool) Delete(key string) {
 	delete(cp.pool, key)
 }
 
-// Get returns map of connections and unlock function to be called
-// after parsing the connections.
-func (cp *ConnectionPool) Get() (*map[string]*Connection, func()) {
-	rlock := cp.rwlock.RLocker()
-	rlock.Lock()
+// GetByNodeID returns map of connections, filtered with given driverName and optional nodeID.
+func (cp *ConnectionPool) GetByNodeID(driverName, nodeID string) map[string]*Connection {
+	cp.rwlock.RLock()
+	defer cp.rwlock.RUnlock()
 
-	return &cp.pool, rlock.Unlock
+	newPool := make(map[string]*Connection)
+	for k, v := range cp.pool {
+		if v.DriverName != driverName {
+			continue
+		}
+		if nodeID != "" && v.NodeID != nodeID {
+			continue
+		}
+		newPool[k] = v
+	}
+
+	return newPool
 }
--- a/internal/connection/connection_pool.go
+++ b/internal/connection/connection_pool.go
@@ -62,21 +62,34 @@ func (cp *ConnectionPool) Delete(key string) {
 	delete(cp.pool, key)
 }
 
-// GetByNodeID returns map of connections, filtered with given driverName and optional nodeID.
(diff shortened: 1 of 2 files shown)


internal/connection/connection_pool.go
```diff
@@ -62,11 +62,21 @@ func (cp *ConnectionPool) Delete(key string) {
 	delete(cp.pool, key)
 }
 
-// Get returns map of connections and unlock function to be called
-// after parsing the connections.
-func (cp *ConnectionPool) Get() (*map[string]*Connection, func()) {
-	rlock := cp.rwlock.RLocker()
-	rlock.Lock()
+// Get returns map of connections, filtered with given driverName and optional nodeID.
+func (cp *ConnectionPool) Get(driverName, nodeID string) map[string]*Connection {
```
Madhu-1, line 66: if the connection is not found return error?

internal/connection/connection_pool.go
```diff
@@ -62,11 +62,21 @@ func (cp *ConnectionPool) Delete(key string) {
...
+	cp.rwlock.RLock()
+	defer cp.rwlock.RUnlock()
 
-	return &cp.pool, rlock.Unlock
+	newPool := make(map[string]*Connection)
+	for k, v := range cp.pool {
+		if v.DriverName != driverName {
+			continue
+		}
+		if nodeID != "" && v.NodeID != nodeID {
```
Madhu-1, line 75: add comment why nodeID empty check is required here.

internal/connection/connection_pool.go
nixpanic, line 66: Assuming we need a drivername for all operations, there should probably be a `getByDrivername(drivername)` (internal only?) and separate `GetByNodeID(drivername, nodeID)` functions. In the future a `GetByCapability(drivername, capability)` can then easily be added without the need to modify (much) of the callers.

nixpanic (changes requested):

internal/connection/connection_pool.go
Rakshith-R: > if the connection is not found return error?

I think it's better not to return an error here, since connectionPool is just doing basic Get,Put,Delete operation with none of it returning an error. 

controller may need to still return a `not found` error if matching caps is not found even when map returned here is not empty.

internal/connection/connection_pool.go
```diff
@@ -62,21 +62,34 @@ func (cp *ConnectionPool) Delete(key string) {
...
 			continue
 		}
+		newPool[k] = v
+	}
+
+	return newPool
+}
+
+// Get returns map of connections, filtered with given driverName and optional nodeID.
+func (cp *ConnectionPool) Get(driverName, nodeID string) map[string]*Connection {
```
nixpanic, line 80: Rename this to `GetByNodeID()`? How would a `Get()` with capabilities instead of nodeID be added instead?

Rakshith-R: missed that, renamed it now.

Madhu-1 (approved):

nixpanic (approved):

Rakshith-R:
@mergifyio rebase

PR #63, opened 2022-01-12
build: use vendored controller-gen

nixpanic:
By installing `controller-gen` from the vendor directory, versions of the
tools can automatically be updated, just like other dependencies. It
also makes things easier to install the tools in environments that have
slow networking or are completely disconnected.


Makefile
```diff
@@ -115,10 +115,11 @@ deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in
...
 	$(KUSTOMIZE) build config/default | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
 
-CONTROLLER_GEN = $(shell pwd)/bin/controller-gen
+# controller-gen gets installed from the vendor/ directory.
+CONTROLLER_GEN = $(shell go env GOPATH)/bin/controller-gen
 .PHONY: controller-gen
-controller-gen: ## Download controller-gen locally if necessary.
-	$(call go-get-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.7.0)
+controller-gen:
+	cd vendor && go install ./$(shell grep controller-gen tools.go | sed 's/.*_ "//;s/"//')
```
Madhu-1, line 122: install it in some temp location as this is specific to the csi-addons repo?

Madhu-1 (commented):
@nixpanic left a question

Makefile
nixpanic: I had a quick look on how to do that, but `go install` does not make that easy.

Using `go build ...` it seems to work, so I'll post that in a bit. Will also update the location where `kustomize` gets installed.

Madhu-1 (approved):

Rakshith-R (approved):

PR #69, opened 2022-01-13
build: publish versioned container-images on pushing a git tag

nixpanic:
When a release is tagged, the GitHub workflow will build the container
images and push them to Quay.io with the `:<git-tag>` version.

A test-run (without publishing) is available in my [personal fork](https://github.com/nixpanic/kubernetes-csi-addons/runs/4801220182?check_suite_focus=true).

### DIFF
--- a/.github/workflows/tag-release.yaml
+++ b/.github/workflows/tag-release.yaml
@@ -1,0 +1,100 @@
+---
+name: Build and release versioned container images
+
+on:
+  create
+
+jobs:
+  tag_bundle:
+    name: Build and release the bundle container-image
+    if: github.repository == 'csi-addons/kubernetes-csi-addons' && github.ref_type  == 'tag'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+
+      - name: Install Go 1.17
+        uses: actions/setup-go@v2
+        with:
+          go-version: '1.17'
+
+      - name: Generate the bundle contents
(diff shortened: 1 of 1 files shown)


yati1998 (dismissed):
LGTM

.github/workflows/tag-release.yaml
```diff
@@ -0,0 +1,93 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+
+      - name: Install Go 1.17
+        uses: actions/setup-go@v2
+        with:
+          go-version: '1.17'
+
```
Rakshith-R, line 20:
```suggestion

      - name: Login to quay.io
        uses: docker/login-action@v1
        with:
          registry: quay.io
          username: ${{ secrets.QUAY_USERNAME }}
          password: ${{ secrets.QUAY_PASSWORD }}
          
```
We need to login to quay here too?
seems like this step for missed @nixpanic 
https://github.com/csi-addons/kubernetes-csi-addons/blob/main/.github/workflows/build-push.yaml#L26-L31

Rakshith-R (changes requested):

.github/workflows/tag-release.yaml
nixpanic: right, will update it!

Rakshith-R (approved):
thanks

yati1998 (approved):

PR #71, opened 2022-01-13
build: use the tagged version in the release bundle

nixpanic:
The bundle container-image always includes `:latest` as the tag for the
controller image. This is not really appropriate, as some bundles will
be generated for a tagged release. When a bundle is tagged with a
release, it should refer to the same version of the controller
container-image.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,8 +3,11 @@
 IMG ?= quay.io/csiaddons/k8s-controller:latest
 SIDECAR_IMG ?= quay.io/csiaddons/k8s-sidecar
 BUNDLE_IMG ?= quay.io/csiaddons/k8s-bundle
-TAG?= latest
 
+# set TAG to a release for consumption in the bundle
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -3,8 +3,11 @@
 IMG ?= quay.io/csiaddons/k8s-controller:latest
 SIDECAR_IMG ?= quay.io/csiaddons/k8s-sidecar
 BUNDLE_IMG ?= quay.io/csiaddons/k8s-bundle
-TAG?= latest
 
+# set TAG to a release for consumption in the bundle
+TAG ?= latest
```
Rakshith-R, line 8: Tag is also used  as sidecar image tag

nixpanic: yes, that is only a good thing?

Rakshith-R (approved):

yati1998 (approved):

PR #84, opened 2022-01-18
controller: add ReclaimSpaceCronJob controller

Rakshith-R:
This commit adds ReclaimSpaceCronJob controller code
and unit tests for the same.

ReclaimSpaceCronJob controller watches for ReclaimSpaceCronJob,
validates resource, parses given schedule, create reclaimspacejob
as per shedule, deletes old failed/successful reclaimspacejob
and updates the status.

Inspired from
- https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/ 
- https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller.go

### DIFF
--- a/controllers/reclaimspacecronjob_controller.go
+++ b/controllers/reclaimspacecronjob_controller.go
@@ -1,0 +1,395 @@
+/*
+Copyright 2022 The Kubernetes-CSI-Addons Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"time"
+
+	csiaddonsv1alpha1 "github.com/csi-addons/kubernetes-csi-addons/api/v1alpha1"
+
+	"github.com/go-logr/logr"
+	"github.com/robfig/cron/v3"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	ref "k8s.io/client-go/tools/reference"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+// ReclaimSpaceCronJobReconciler reconciles a ReclaimSpaceCronJob object
+type ReclaimSpaceCronJobReconciler struct {
+	client.Client
+	Scheme *runtime.Scheme
+}
+
+const (
+	jobOwnerKey = ".metadata.controller"
+	// default values for Spec parameters.
+	defaultFailedJobsHistoryLimit     int32 = 1
+	defaultSuccessfulJobsHistoryLimit int32 = 3
+)
+
+var (
+	apiGVStr                = csiaddonsv1alpha1.GroupVersion.String()
+	scheduledTimeAnnotation = csiaddonsv1alpha1.GroupVersion.Group + "/scheduled-at"
+)
+
+//+kubebuilder:rbac:groups=csiaddons.openshift.io,resources=reclaimspacecronjobs,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=csiaddons.openshift.io,resources=reclaimspacecronjobs/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=csiaddons.openshift.io,resources=reclaimspacecronjobs/finalizers,verbs=update
+//+kubebuilder:rbac:groups=csiaddons.openshift.io,resources=reclaimspacejobs,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=csiaddons.openshift.io,resources=reclaimspacejobs/status,verbs=get
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+// Reconcile will be called when there is any event related to reclaimSpaceCronJob and also when
+// there is an event related to child reclaimSpaceJobs due to controllerOwnerRef.
+func (r *ReclaimSpaceCronJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	_ = log.FromContext(ctx)
+	logger := log.FromContext(ctx)
+
+	// Fetch ReclaimSpaceCronJob instance
+	rsCronJob := &csiaddonsv1alpha1.ReclaimSpaceCronJob{}
+	err := r.Client.Get(ctx, req.NamespacedName, rsCronJob)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			// Request object not found, could have been deleted after reconcile request.
+			logger.Info("ReclaimSpaceCronJob resource not found")
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
+	// set history limit defaults, if not specified.
+	if rsCronJob.Spec.FailedJobsHistoryLimit == nil {
+		*rsCronJob.Spec.FailedJobsHistoryLimit = defaultFailedJobsHistoryLimit
+	}
+	if rsCronJob.Spec.SuccessfulJobsHistoryLimit == nil {
+		*rsCronJob.Spec.SuccessfulJobsHistoryLimit = defaultSuccessfulJobsHistoryLimit
+	}
+
+	var childJobs csiaddonsv1alpha1.ReclaimSpaceJobList
+	err = r.List(ctx, &childJobs, client.InNamespace(req.Namespace), client.MatchingFields{jobOwnerKey: req.Name})
+	if err != nil {
+		logger.Error(err, "Failed to list child ReclaimSpaceJobs")
+		return ctrl.Result{}, err
+	}
+
+	// find the active,failed and successful list of jobs and mostRecent,lastSuccessful time.
+	childJobsInfo := parseJobList(&logger, &childJobs)
+
+	if childJobsInfo.mostRecentTime != nil {
+		rsCronJob.Status.LastScheduleTime = &metav1.Time{Time: *childJobsInfo.mostRecentTime}
+	} else {
+		rsCronJob.Status.LastScheduleTime = nil
+	}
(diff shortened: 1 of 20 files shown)


controllers/reclaimspacecronjob_controller.go
```diff
@@ -0,0 +1,370 @@
...
+package controllers
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"time"
+
+	csiaddonsv1alpha1 "github.com/csi-addons/kubernetes-csi-addons/api/v1alpha1"
+	"github.com/go-logr/logr"
```
nixpanic, line 26: group this import with the other non-local imports

controllers/reclaimspacecronjob_controller.go
```diff
@@ -0,0 +1,370 @@
...
+
+//+kubebuilder:rbac:groups=csiaddons.openshift.io,resources=reclaimspacecronjobs,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=csiaddons.openshift.io,resources=reclaimspacecronjobs/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=csiaddons.openshift.io,resources=reclaimspacecronjobs/finalizers,verbs=update
+//+kubebuilder:rbac:groups=csiaddons.openshift.io,resources=reclaimspacejobs,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=csiaddons.openshift.io,resources=reclaimspacejobs/status,verbs=get
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+func (r *ReclaimSpaceCronJobReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
nixpanic, line 64: maybe explain when this is called? I think it gets called when a child ReclaimSpaceJob has been updated too?

controllers/reclaimspacecronjob_controller.go
```diff
@@ -0,0 +1,370 @@
...
+		logger.Error(err, "Failed to list child ReclaimSpaceJobs")
+		return ctrl.Result{}, err
+	}
+
+	// find the active,failed and successful list of jobs
+	var activeJob *csiaddonsv1alpha1.ReclaimSpaceJob
+	var successfulJobs, failedJobs []*csiaddonsv1alpha1.ReclaimSpaceJob
+	var mostRecentTime, lastSuccessfulTime *time.Time // find the last run so we can update the status
+
+	for i, job := range childJobs.Items {
```
Madhu-1, line 101: move this to a helper function and return required objects?

controllers/reclaimspacecronjob_controller.go
```diff
@@ -0,0 +1,370 @@
...
+	}
+	rsCronJob.Status.Active = nil
+	if activeJob != nil {
+		jobRef, err := ref.GetReference(r.Scheme, activeJob)
+		if err != nil {
+			logger.Error(err, "Failed to make reference to active job", "job", activeJob)
+		}
+		rsCronJob.Status.Active = jobRef
+	}
+	logger.Info("ReclaimSpaceJob count", "successfulJobs", len(successfulJobs), "failedJobs", len(failedJobs))
```
Madhu-1, line 148: job successful and failed job names?

controllers/reclaimspacecronjob_controller.go
```diff
@@ -0,0 +1,372 @@
...
+	}
+
+	// make sure we're not too late to start the run
+	logger = logger.WithValues("currentRun", missedRun)
+	tooLate := false
+	if rsCronJob.Spec.StartingDeadlineSeconds != nil {
+		tooLate = missedRun.Add(time.Duration(*rsCronJob.Spec.StartingDeadlineSeconds) * time.Second).Before(r.Now())
+	}
+	if tooLate {
+		logger.Info("Missed starting deadline for last run, sleeping till next")
```
Madhu-1, line 189: we are not sleeping here? improve log

controllers/reclaimspacecronjob_controller.go
```diff
@@ -0,0 +1,372 @@
...
+// SetupWithManager sets up the controller with the Manager.
+func (r *ReclaimSpaceCronJobReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &csiaddonsv1alpha1.ReclaimSpaceJob{}, jobOwnerKey, func(rawObj client.Object) []string {
+		// extract the owner from job object.
+		job := rawObj.(*csiaddonsv1alpha1.ReclaimSpaceJob)
+		owner := metav1.GetControllerOf(job)
+		if owner == nil {
+			return nil
+		}
+		if owner.APIVersion != apiGVStr || owner.Kind != "ReclaimSpaceCronJob" {
```
Madhu-1, line 234: don't we already have kind defined in types.go for ReclaimSpaceCronJob , if yes please use it

controllers/reclaimspacecronjob_controller.go
```diff
@@ -0,0 +1,372 @@
...
+	// Reconcile will be triggered if job starts or finishes, cronjob is modified, etc.
+	// Requeue till next run.
+	return scheduledResult, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *ReclaimSpaceCronJobReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &csiaddonsv1alpha1.ReclaimSpaceJob{}, jobOwnerKey, func(rawObj client.Object) []string {
+		// extract the owner from job object.
+		job := rawObj.(*csiaddonsv1alpha1.ReclaimSpaceJob)
```
Madhu-1, line 229: this can cause panic better to do typecasting check here?

controllers/reclaimspacecronjob_controller.go
```diff
@@ -0,0 +1,372 @@
...
+	}
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&csiaddonsv1alpha1.ReclaimSpaceCronJob{}).
+		Owns(&csiaddonsv1alpha1.ReclaimSpaceJob{}).
+		Complete(r)
+}
+
+// Now returns the current local time.
+func (r *ReclaimSpaceCronJobReconciler) Now() time.Time {
```
Madhu-1, line 250: this can be a function not a method

controllers/reclaimspacecronjob_controller.go
```diff
@@ -0,0 +1,372 @@
...
+	return time.Now()
+}
+
+// constructRSJobForCronJob constructs reclaimspacejob.
+func (r *ReclaimSpaceCronJobReconciler) constructRSJobForCronJob(
```
Madhu-1, line 255: keep it as a function not as a method as `r` is not used?

controllers/reclaimspacecronjob_controller.go
```diff
@@ -0,0 +1,372 @@
...
+	}
+	return &timeParsed, nil
+}
+
+// TODO: add unit test for getNextSchedule
+// getNextSchedule returns lastMissed and next time after parsing the schedule.
+// This function returns error if there are more than 100 missed start times.
+func getNextSchedule(
+	rsCronJob *csiaddonsv1alpha1.ReclaimSpaceCronJob,
+	now time.Time) (lastMissed time.Time, next time.Time, err error) {
```
Madhu-1, line 336: `now time.Time) (lastMissed time.Time, next time.Time, err error) {` to `now time.Time) (ltime.Time,  time.Time,  error) {`

controllers/reclaimspacecronjob_controller.go
```diff
@@ -0,0 +1,372 @@
...
+	}
+
+	starts := 0
+	for t := sched.Next(earliestTime); !t.After(now); t = sched.Next(t) {
+		lastMissed = t
+		starts++
+		if starts > 100 {
+			// We can't get the most recent times so just return an empty slice
+			return time.Time{}, time.Time{},
+				fmt.Errorf("Too many missed start times (> 100). Set or decrease" +
```
Madhu-1, line 367: start error msg with small letter.

controllers/reclaimspacecronjob_controller_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+package controllers
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	csiaddonsv1alpha1 "github.com/csi-addons/kubernetes-csi-addons/api/v1alpha1"
+	"github.com/stretchr/testify/assert"
```
Madhu-1, line 25: new line is required

controllers/reclaimspacecronjob_controller.go
Rakshith-R: > move this to a helper function and return required objects?

It would need to return 5 variables, IMO better to keep it inline in the same function flow.

controllers/reclaimspacecronjob_controller.go
Rakshith-R: > job successful and failed job names?

The number of job maybe large, 
the child jobs have name = `<cronjob-name>-<time-hash>`
so it is easy to identify child jobs
```
[rakshith@fedora kubernetes-csi-addons]$ k get reclaimspacejobs.csiaddons.openshift.io 
NAME                                    NAMESPACE   AGE    RETRIES   RESULT
reclaimspacecronjob-sample-1642503300   rook-ceph   109m             Succeeded
reclaimspacecronjob-sample-1642503360   rook-ceph   108m             Succeeded
reclaimspacecronjob-sample-1642503420   rook-ceph   107m             Succeeded
```
I don't think printing names is required

controllers/reclaimspacecronjob_controller.go
Madhu-1: as we make more changes this becomes difficult to maintain. if you are concerned with returns variables create a  struct and return it

controllers/reclaimspacecronjob_controller.go
Madhu-1: in that case, IMO logging just numbers is not useful at all. as it won't provide information like which all are failed and succeed.

controllers/reclaimspacecronjob_controller.go
Rakshith-R: > don't we already have kind defined in types.go for ReclaimSpaceCronJob , if yes please use it

I don't think so , was not able to find it.

controllers/reclaimspacecronjob_controller.go
Rakshith-R:
```
	if err := ctrl.SetControllerReference(rsCronJob, job, r.Scheme); err != nil {
		return nil, err
	}
```
it is used.

controllers/reclaimspacecronjob_controller.go
Rakshith-R: removing it

controllers/reclaimspacecronjob_controller.go
Rakshith-R: > as we make more changes this becomes difficult to maintain. if you are concerned with returns variables create a struct and return it

done,ptal

nixpanic (approved):

PR #91, opened 2022-01-19
build: enable inclusion of differently versioned container-images in the bundle

nixpanic:
In case the *_IMG variables can contain a full qualified container-image
resource (includes a ":"), the container-images should not use the TAG
valued. The :TAG option will only be added if no predefined version is
passed.


yati1998 (approved):
LGTM

Rakshith-R (approved):

Madhu-1 (approved):

PR #94, opened 2022-01-19
docs: healer controller design proposal

pkalever:
Signed-off-by: Prasanna Kumar Kalever \<prasanna.kalever@redhat.com\>

### DIFF
--- a/docs/design/proposals/healer-controller.md
+++ b/docs/design/proposals/healer-controller.md
@@ -1,0 +1,42 @@
+# Introduction
+There is [external health monitoring](https://github.com/kubernetes-csi/external-health-monitor/) which does mostly similar stuff, but actions are not in the scope of it, all it can do is log the health condition of the mounts.
+
+# Volume Healer Controller
+
+**Now the obvious question we expect is "what do we achieve with the new controller?"**
+* Achieve a way to monitor the health of volumes.
+* We currently do not have a way to act on failed volumes. Once the controller detects that the volume is not healthy, it will have an opportunity to act on it. For example userspace mounters such as NBD has a process running per volume, this controller should ideally help restart such processes in case if they are not-started or crashed.
+* This will be a generic controller design for any mounters added in the future, for userspace mounters it will offer healing in addition, like restarting the respective processes if the volume goes abnormal. Some example userspace mounters would be rbd-nbd, cephfs-fuse, etc.
+
+# Details
+
+                    .-------------------.
+                    |     CSI-Addons    |
+                    | Healer Controller |
+                    '-------------------'
+                              |
+                              | gRPC
+                              |
+            .---------+------------------------------.
+            |         |                              |
+            |  .------------.        .------------.  |
+            |  | CSI-Addons |  gRPC  |    CSI     |  |
+            |  |  side-car  |--------| NodePlugin |  |
+            |  '------------'        '------------'  |
+            | CSI-driver Pod                         |
+            '----------------------------------------'
+
+## CSI-Addons Healer Controller
+* The healer controller will be implemented as part of the CSI-Addons controller. It will trigger controller RPC to check the health condition of the CSI volumes.
+* Watch for the `CSIAddonsNode` object, if the object is created/modified by the side-car do below operations
+    * Once the connection is established, fetch the capabilities list and check for nodes that supports `VOLUME_HEALER` capability
+    * Fetch all the VolumeAttachment` list
+    * Filter the `VolumeAttachment` list through matching driver name, node name and status attached
+    * For each `VolumeAttachment` get the respective PersistentVolume information and check the criteria of PersistentVolume Bound (also we can have configuration options like mounter type)
+    * Send the RPC request for side-car to check if the volume is healthy along with the PersistentVolume details and may attempt to heal the volume if it is found abnormal
+    * Log the final response returned from the side-car
+
+## CSI-Addons side-car
+* When a nodeplugin pod is restarted the side-car will create [CSIAddonsNode](https://github.com/csi-addons/kubernetes-csi-addons/blob/main/api/v1alpha1/csiaddonsnode_types.go) object.
+* The side-car works as relay, upon receiving a RPC call from healer controller, the CSI-Addons side-car will then send a VolumeMonitor request to CSI driver. The request contains some internal details like secrets, staging path, volume-id etc.
+* Upon receiving the VolumeMonitor request, the CSI driver check volume's mounting conditions, in case if any volume condition is reported abnormal, it may act on it by performing required healing and finally returns back the response to the controller.


pkalever:
As discussed, had opened https://github.com/ceph/ceph-csi/issues/2794 for the `VOLUME_CONDITION` capability.
Please comment below if we can split this into more granular parts.

docs/design/proposals/healer-controller.md
```diff
@@ -0,0 +1,120 @@
...
+
+## CSI-Addons Healer Controller
+* The healer controller will be implemented as part of the CSI-Addons controller driver. It is the one who triggers controller RPC to check the health condition of the CSI volumes.
+* Watch for the `CSIAddonsNode` object, if the object is created/modified by the side-car do below operations
+    * Fetch all the volume attachments (VA) list
+    * Filter the volume attachments list through matching driver name and status attached
+    * For each volume attachment get the respective PV information and check the criteria of PV Bound (also we can have configuration options like mounter type)
+    * Send the RPC request for side-car to check if the volume is healthy along with the PV details details
+    * Log the final response returned from the side-car
+* The healer controller periodically triggers the health status checks of the CSI volumes via the gRPC calls.
```
Rakshith-R, line 39:
```suggestion
* The healer controller periodically triggers the health status checks of the CSI volumes via the gRPC calls.
```
we should combine this functionality in the same controller and it requires a separate docs with explanation of which volumes should be watched, how and so on.

docs/design/proposals/healer-controller.md
```diff
@@ -0,0 +1,120 @@
...
+            |  |  side-car  |--------| NodePlugin |  |
+            |  '------------'        '------------'  |
+            | CSI-driver Pod                         |
+            '----------------------------------------'
+
+## CSI-Addons Healer Controller
+* The healer controller will be implemented as part of the CSI-Addons controller driver. It is the one who triggers controller RPC to check the health condition of the CSI volumes.
+* Watch for the `CSIAddonsNode` object, if the object is created/modified by the side-car do below operations
+    * Fetch all the volume attachments (VA) list
+    * Filter the volume attachments list through matching driver name and status attached
```
Rakshith-R, line 35:
```suggestion
    * Filter the volume attachments list through matching driver name, node name and status attached
    * Wait for status to be in connected state.
```
Only after csiaddonsnode is in connected state, connection object will be available in pool to be used.

docs/design/proposals/healer-controller.md
```diff
@@ -0,0 +1,120 @@
...
+    * Send the RPC request for side-car to check if the volume is healthy along with the PV details details
+    * Log the final response returned from the side-car
+* The healer controller periodically triggers the health status checks of the CSI volumes via the gRPC calls.
+
+## CSI-Addons side-car
+* The side-car registers itself by creating a `CSIAddonsNode` CR that the CSI-Addons Controller can use to connect to the side-car and execute operations.
+* When a nodeplugin container is restarted the side-car will update/create(if not present) [CSIAddonsNode](https://github.com/csi-addons/kubernetes-csi-addons/blob/main/api/v1alpha1/csiaddonsnode_types.go) object.
+* The side-car works as relay, upon receiving a gRPC call from healer controller, the CSI-Addons side-car will then check volume's mounting conditions collected via the CSI NodeGetVolumeStats RPC.
+* In case if any volume condition is reported abnormal, the CSI-Addons side-car will act on it, it does some internal operation like fetching secrets, constructing staging path and finally sends NodeStageVolume request to the CSI driver and finally returns back the response to the controller.
```
Rakshith-R, line 45: Just sent a heal call to csidriver and let it decide whether node stage req is required or not.

docs/design/proposals/healer-controller.md
```diff
@@ -0,0 +1,120 @@
...
+
+PS: Some modification might be required in side-car to update the `CSIAddonsNode` object if its already present and its minimal change.
```
Rakshith-R, line 47: discard above comment regarding the same :sweat_smile:

docs/design/proposals/healer-controller.md
pkalever: I know it looks simple but, I think this needs more cycles for fetching secrets, formatting the staging path, and capturing/collecting more details needed to make the NodeStageCall(), for almost all the volume even if they are healthy.

docs/design/proposals/healer-controller.md
pkalever: Why should we care about node names?

docs/design/proposals/healer-controller.md
pkalever: This can be considered as a task that can be split, but I don't see a reason why this needs a separate doc, maybe we can open an issue and discuss more details there!

docs/design/proposals/healer-controller.md
Rakshith-R: > I know it looks simple but, I think this needs more cycles for fetching secrets, formatting the staging path, and capturing/collecting more details needed to make the NodeStageCall(), for almost all the volume even if they are healthy.

sidecar is just a relay, 
its better to have more api calls in single request flow from controller to sidecar to csidriver
than to have back and forth between sidecar and csi-driver while controller sits waiting.

docs/design/proposals/healer-controller.md
Rakshith-R: > Why should we care about node names?

you want to heals volume from that particular node where sidecar resides and got restarted ?

docs/design/proposals/healer-controller.md
Rakshith-R: > This can be considered as a task that can be split, but I don't see a reason why this needs a separate doc, maybe we can open an issue and discuss more details there!

No, same controller should not watch for an event and also do periodic checks for another event.

pkalever: 1) If the event is currently being handled, just cancel the periodic check.
2) If the event comes while periodic check is ongoing, just abort the periodic check and restart the healer check again!

docs/design/proposals/healer-controller.md
pkalever: Do we need to filter it? I think the event watcher lists the `CSIAddonsNode` object for Add/modify already, that should take care I guess?

docs/design/proposals/healer-controller.md
pkalever: I agree that this is a bit debatable topic on its own.
1. For restart of the nodeplugin case, it will mostly need NodeStageVolume() for 100% of PVs
2. For periodic checks, I'm optimistic that most of the time the health of the volume is in a good state, only the corner cases are process restarts (considering userspace mounters). Hence I think sending NodeStageVolume() all the time is a bit costly.

So maybe a hybrid model is needed to satisfy and save some jiffies here, like
* For restart case send a NodeStageVolume()
* For periodic health checks scenarios, send a NodeGetVolumeStats() first and then send a NodeStageVolume() if the status of the PV is set to abnormal?

docs/design/proposals/healer-controller.md
```diff
@@ -0,0 +1,120 @@
+# Introduction
+There is [external health monitoring](https://github.com/kubernetes-csi/external-health-monitor/) which does mostly similar stuff, but actions are not in the scope of it, all it can do is log the health condition of the mounts.
+
+# Volume Healer Controller
+Before we begin it is noteworthy to know that the existing healer logic remains the same.
+
+**Now the obvious question we expect is "what do we achieve with the new controller?"**
+* Moving the k8s specific code out of ceph-csi, to make ceph-csi orchestrator independent.
```
nixpanic, line 8: Use names of projects correctly, use Kubenetes, Ceph-CSI.

docs/design/proposals/healer-controller.md
```diff
@@ -0,0 +1,120 @@
+# Introduction
+There is [external health monitoring](https://github.com/kubernetes-csi/external-health-monitor/) which does mostly similar stuff, but actions are not in the scope of it, all it can do is log the health condition of the mounts.
+
+# Volume Healer Controller
+Before we begin it is noteworthy to know that the existing healer logic remains the same.
```
nixpanic, line 5: what existing healer logic? there is none in this project

docs/design/proposals/healer-controller.md
```diff
@@ -0,0 +1,120 @@
...
+
+**Now the obvious question we expect is "what do we achieve with the new controller?"**
+* Moving the k8s specific code out of ceph-csi, to make ceph-csi orchestrator independent.
+* Achieve a way to monitor the health of NBD volumes.
+* We currently do not have a way to act on failed/crashed NBD volumes with the existing healer logic. Once the controller detects that the volume is not healthy, it will have an opportunity to act on it.
```
nixpanic, line 10: Do not focus on NBD, it is just one example. Why would NBD be different from other block devices or mounters?

docs/design/proposals/healer-controller.md
```diff
@@ -0,0 +1,120 @@
...
+* This will be a generic framework for any userspace mounters added in the future, example ceph-fuse
```
nixpanic, line 11: csi-addons is the framework! Phrase it differently, explain what the framework offers, and only after that describe an example with Ceph-CSI and rbd-nbb (and maybe mention that the healing is optional, e.g. krbd would not need it?)

docs/design/proposals/healer-controller.md
```diff
@@ -0,0 +1,120 @@
...
+            |         |                              |
+            |  .------------.        .------------.  |
+            |  | CSI-Addons |  gRPC  |    CSI     |  |
+            |  |  side-car  |--------| NodePlugin |  |
+            |  '------------'        '------------'  |
+            | CSI-driver Pod                         |
+            '----------------------------------------'
+
+## CSI-Addons Healer Controller
+* The healer controller will be implemented as part of the CSI-Addons controller driver. It is the one who triggers controller RPC to check the health condition of the CSI volumes.
```
nixpanic, line 32: just call it "CSI-Addons Controller".

"It is the new one" reads a little weird, obviously it is new if it will be implemented

docs/design/proposals/healer-controller.md
```diff
@@ -0,0 +1,120 @@
...
+* Watch for the `CSIAddonsNode` object, if the object is created/modified by the side-car do below operations
```
nixpanic, line 33: why `CSIAddonsNode` and not `VolumeAttachment`? ideally use the same  formatting/capitalization for all Kuberenets objects

docs/design/proposals/healer-controller.md
```diff
@@ -0,0 +1,120 @@
...
+    * Fetch all the volume attachments (VA) list
+    * Filter the volume attachments list through matching driver name and status attached
+    * For each volume attachment get the respective PV information and check the criteria of PV Bound (also we can have configuration options like mounter type)
```
nixpanic, line 36: PV -> PersistentVolume, try to reduce abbreviations to make it better readable

What is "mounter type" here? The StorageClass, or CSI-driver internal parameters that CSI-Addons should not inspect?

docs/design/proposals/healer-controller.md
```diff
@@ -0,0 +1,120 @@
...
+    * Send the RPC request for side-car to check if the volume is healthy along with the PV details details
```
nixpanic, line 37: not only check, but heal as well?

docs/design/proposals/healer-controller.md
nixpanic, line 39: I do not think a regular check is needed, you refer to https://github.com/kubernetes-csi/external-health-monitor/ for that in the top of the document

docs/design/proposals/healer-controller.md
```diff
@@ -0,0 +1,120 @@
...
+    * Log the final response returned from the side-car
+* The healer controller periodically triggers the health status checks of the CSI volumes via the gRPC calls.
+
+## CSI-Addons side-car
+* The side-car registers itself by creating a `CSIAddonsNode` CR that the CSI-Addons Controller can use to connect to the side-car and execute operations.
```
nixpanic, line 42: no need to go through the existing steps of CSIAddonsNode, that was a different feature, just refer to the doc

docs/design/proposals/healer-controller.md
nixpanic: CSI-Addons can not send CSI commands, those are reserved for the CO.

CSI-drivers can decide on their own implementation of healing, maybe they want to call their own NodeStageVolume procedure. Don't mix standard CSI procedures with CSI-Addons procedures.

There is a difference between regular health checks, and an informational message "you restarted, and you should make sure the volume is still available". The last part is expected when a CSIAddonsNode is detected, the 1st is for https://github.com/kubernetes-csi/external-health-monitor/

docs/design/proposals/healer-controller.md
nixpanic, line 47: not sure what you mean by this?

nixpanic (changes requested):
We still need to have a description of the procedures that a CSI-driver should implement for this functionality. That description needs to get reviewed and merged in the csi-addons/spec repository before a Kubernetes implementation can be proposed.

docs/design/proposals/healer-controller.md
pkalever: You read it wrong, I worded it as "It is the one" not "It is the new one"

docs/design/proposals/healer-controller.md
pkalever: `VolumeAttachment` is specific to a single volume, which is not our interest here. We should be watching for events that are at node level i.e. `CSIAddonsNode`

Volume level abnormalities should be taken care by the periodic healing checks.

docs/design/proposals/healer-controller.md
pkalever: We want a way for users to provide global options/filters, I don't know yet what would be the right place to capture such specifications from users.

Still thinking about this part.

docs/design/proposals/healer-controller.md
pkalever: Soon after the sidecar is restarted, if it sees `CSIAddonsNode` exist, we need to implement a way to update the already existing `CSIAddonsNode` object.

pkalever:
> We still need to have a description of the procedures that a CSI-driver should implement for this functionality. That description needs to get reviewed and merged in the csi-addons/spec repository before a Kubernetes implementation can be proposed.

I had already opened https://github.com/ceph/ceph-csi/issues/2794, and will update the implementation details there soon.

I will also send the spec changes after some more clarity on the implementation side details(more on the internal structures). Maybe in parallel to the implementation PR.

This is just the overall design idea, for everyone's reference.

> CSI-Addons can not send CSI commands, those are reserved for the CO.

I'm a bit surprised to hear this @nixpanic , is it restricted anywhere? or at least documented?
I definitely thought the sidecar can call NodeStageVolume() RPC in the nodeplugin pod.

> CSI-drivers can decide on their own implementation of healing, maybe they want to call their own NodeStageVolume procedure. Don't mix standard CSI procedures with CSI-Addons procedures.

Yes, definitely the CSI-Addons can implement its own RPC, just wanted to avoid another copy-paste.

docs/design/proposals/healer-controller.md
nixpanic: that should be handled by the CSIAddonsNode feature, it is not relevant enough to mention here

I think you are concerned that the CSI-driver container in the pod restarted, and not the pod was re-created. Only in that case, the CSIAddonsNode object would exist. It is not clear to me if this is something that can happen in reality.

nixpanic:
> > CSI-Addons can not send CSI commands, those are reserved for the CO.
> 
> I'm a bit surprised to hear this @nixpanic , is it restricted anywhere? or at least documented? I definitely thought the sidecar can call NodeStageVolume() RPC in the nodeplugin pod.

Drivers (at least Ceph-CSI) offer different gRPC servers for different protocols, so it is technically prevented. Mixing protocols does not look like a good practice, or very clean.

If both the CO and CSI-Addons Controller send CSI operations, we break the spec https://github.com/container-storage-interface/spec/blob/master/spec.md#concurrency

A CSI-driver will need to take care that healing does not interfere with actions the CO does. The only practical way to do so, is by using different operations for CSI and CSI-Addons that can coordinate (block/abort/..) concurrent access.

docs/design/proposals/healer-controller.md
pkalever: Ah! I realized what you meant now.

docs/design/proposals/healer-controller.md
pkalever: As discussed lets work on one at a time, may be will target the periodic healer triggers later

docs/design/proposals/healer-controller.md
pkalever: For now, as we are focusing on restarting node plugin case, we might want to fill in all details in one shot and send. As discussed for periodic healing we might need some kind of a hybrid model to reduce the cycles involved in collecting details which might be unneeded in most of the times

pkalever: > CSI-Addons can not send CSI commands, those are reserved for the CO.

@nixpanic We wanted to send NodeGetVolumeStats() to check the volume condition, why not NodeStageVolume() ?

pkalever: I was thinking we can implement a function like HealVolume() that can be called from sidecar which does NodeGetVolumeStats() and  Heal() in one go, as @Rakshith-R pointed above

pkalever:
> > > CSI-Addons can not send CSI commands, those are reserved for the CO.
> > 
> > 
> > I'm a bit surprised to hear this @nixpanic , is it restricted anywhere? or at least documented? I definitely thought the sidecar can call NodeStageVolume() RPC in the nodeplugin pod.
> 
> Drivers (at least Ceph-CSI) offer different gRPC servers for different protocols, so it is technically prevented. Mixing protocols does not look like a good practice, or very clean.
> 
> If both the CO and CSI-Addons Controller send CSI operations, we break the spec https://github.com/container-storage-interface/spec/blob/master/spec.md#concurrency
> 
> A CSI-driver will need to take care that healing does not interfere with actions the CO does. The only practical way to do so, is by using different operations for CSI and CSI-Addons that can coordinate (block/abort/..) concurrent access.

@nixpanic IIUC, this means even sidecars calling NodeGetVolumeStats is bad?

docs/design/proposals/healer-controller.md
```diff
@@ -0,0 +1,115 @@
...
+            |  | CSI-Addons |  gRPC  |    CSI     |  |
+            |  |  side-car  |--------| NodePlugin |  |
+            |  '------------'        '------------'  |
+            | CSI-driver Pod                         |
+            '----------------------------------------'
+
+## CSI-Addons Healer Controller
+* The healer controller will be implemented as part of the CSI-Addons controller. It will trigger controller RPC to check the health condition of the CSI volumes.
+* Watch for the `CSIAddonsNode` object, if the object is created/modified by the side-car do below operations
+    * Fetch all the VolumeAttachment` list
```
Rakshith-R, line 33:
```suggestion
    * Wait for connection to be established and determine if `VolumeHeal` capability is present. (steps similar to this)
    * Fetch all the VolumeAttachment` list
```
After CSIAddons controller connects to the driver, It'll fetch and stores its capabilities. 
We can use that capabilities list to determine whether `volumeheal` cap is present or not and then proceed further.

docs/design/proposals/healer-controller.md
```diff
@@ -0,0 +1,115 @@
...
+    * Filter the `VolumeAttachment` list through matching driver name, node name and status attached
+    * For each `VolumeAttachment` get the respective PersistentVolume information and check the criteria of PersistentVolume Bound (also we can have configuration options like mounter type)
+    * Send the RPC request for side-car to check if the volume is healthy along with the PersistentVolume details and may attempt to heal the volume if it is found abnormal
+    * Log the final response returned from the side-car
+
+## CSI-Addons side-car
+* When a nodeplugin container is restarted the side-car will update/create(if not present) [CSIAddonsNode](https://github.com/csi-addons/kubernetes-csi-addons/blob/main/api/v1alpha1/csiaddonsnode_types.go) object.
```
Rakshith-R, line 40:
```suggestion
* When a nodeplugin pod is restarted the side-car will create [CSIAddonsNode](https://github.com/csi-addons/kubernetes-csi-addons/blob/main/api/v1alpha1/csiaddonsnode_types.go) object.
```

When a pod is restarted, it essentially is deleted (which also deletes corresponding CSIAddonsNode object) and
a new pod is created (with new CSIAddonsNode object).

docs/design/proposals/healer-controller.md
```diff
@@ -0,0 +1,115 @@
...
+            |  '------------'        '------------'  |
+            | CSI-driver Pod                         |
+            '----------------------------------------'
+
+## CSI-Addons Healer Controller
+* The healer controller will be implemented as part of the CSI-Addons controller. It will trigger controller RPC to check the health condition of the CSI volumes.
+* Watch for the `CSIAddonsNode` object, if the object is created/modified by the side-car do below operations
+    * Fetch all the VolumeAttachment` list
+    * Filter the `VolumeAttachment` list through matching driver name, node name and status attached
+    * For each `VolumeAttachment` get the respective PersistentVolume information and check the criteria of PersistentVolume Bound (also we can have configuration options like mounter type)
```
Rakshith-R, line 35: Send request irrespective of any particular PV config option ?
Or
let the driver include more information in its capabilities response, so this decision can be made with the help of the extra info provided by the driver itself ?

docs/design/proposals/healer-controller.md
```diff
@@ -0,0 +1,115 @@
...
+    * Send the RPC request for side-car to check if the volume is healthy along with the PersistentVolume details and may attempt to heal the volume if it is found abnormal
+    * Log the final response returned from the side-car
+
+## CSI-Addons side-car
+* When a nodeplugin container is restarted the side-car will update/create(if not present) [CSIAddonsNode](https://github.com/csi-addons/kubernetes-csi-addons/blob/main/api/v1alpha1/csiaddonsnode_types.go) object.
+* The side-car works as relay, upon receiving a gRPC call from healer controller, the CSI-Addons side-car will then check volume's mounting conditions collected via the CSI NodeGetVolumeStats RPC.
+* In case if any volume condition is reported abnormal, the CSI-Addons side-car will act on it, it does some internal operation like fetching secrets, constructing staging path and finally sends StageVolume request and finally returns back the response to the controller.
```
Rakshith-R, line 42: We agreed it will be a single call?
and let the driver decide and heal on its own?

Rakshith-R (changes requested):
It is expected that CSI-Driver will run two different servers one for CSI Spec and one for CSI-Addons spec. 

I think, from CSIAddons sidecar perspective, we should only call CSI-Addons spec rpc calls and not mix both of them.

docs/design/proposals/healer-controller.md
pkalever: Yes, this is right. I will capture it, I just took this for granted.

docs/design/proposals/healer-controller.md
pkalever: Agree

docs/design/proposals/healer-controller.md
pkalever: My idea is similar to the latter one, somehow get the choice like, what kind of mounter types need what services.
Example:
```
{
  rbd-nbd: healing
  krbd: monitoring
  xyz: disable
}
```

docs/design/proposals/healer-controller.md
pkalever: Yes, for the restart case only, send everything in one go.

docs/design/proposals/healer-controller.md
pkalever: I would like to leave this more for an implementation-specific detail.

pkalever:
@Madhu-1 @nixpanic @Rakshith-R Can we finalize this, please? Let me know if we are missing any major calrifications in the design. If not, I hope we can discuss and address any nitty-gritty details in the implementation PR.

Thanks!

docs/design/proposals/healer-controller.md
nixpanic: A function like `HealVolume()` would need implementation in the CSI-driver (called from the sidecar). CSI-drivers needs to handle healing differently, I expect.

`NodeGetVolumeStats()` is a CSI call, so can not be called from CSI-Addons operations. You need to design an operation that the CSI-Addons sidecar can call on the CSI-driver. The design for `Heal()` and a `HealthCheck()` function need to go to csi-addons/spec.

nixpanic:
> @nixpanic IIUC, this means even sidecars calling NodeGetVolumeStats is bad?

Yes. CSI-Addons should not mix calling CSI operations. A CSI-driver will need to provide dedicated CSI-Addons operations that can be called instead. The CSI-driver is then in charge of making sure no conflicting calls from the CO are being processed.

docs/design/proposals/healer-controller.md
nixpanic: CSI-Addons capabilities from a CSI-driver should include if it supports healing a volume. If the CSI-driver does not need healing, it can omit the capability.
In case the CSI-driver needs to perform heal checks on a volume (depending on the volume options?), it can either return things like `HEALTH_OK` (to recheck later), `HEALTH_UNSUPPORTED` (no need to check, this volume does not support healing), `HEALTH_BAD` (needs healing) and so on. But this is a discussion for csi-addons/spec.

nixpanic: CSI-Addons should not need to know about different _mounter types_, that would require extending the list of types and the implementation for each CSI-driver. Better have a generic approach that works for all CSI-drivers, so that any CSI-driver can opt-in without changing CSI-Addons itself.

docs/design/proposals/healer-controller.md
pkalever: > A function like HealVolume() would need implementation in the CSI-driver (called from the sidecar). CSI-drivers needs to handle healing differently, I expect.

Yes

>NodeGetVolumeStats() is a CSI call, so can not be called from CSI-Addons operations. You need to design an operation that the CSI-Addons sidecar can call on the CSI-driver. The design for Heal() and a HealthCheck() function need to go to csi-addons/spec.

@nixpanic 
 oh! sorry if that confused you, I now understand why we do not want to use NodeGetVolumeStats() served by the CSI driver which was for CSI Spec. I mean NodeGetVolumeStats() RPC is also served by CSI-addons server. Do we see any reason why we do not want to use the same RPC signature logically? do you think it just creates some additional confusion? if so we can just name it NodeGetVolumeStatus().

pkalever: Or maybe I will just have the new RPC naming it as NodeGetVolumeStatus() which is mostly a copy of NodeGetVolumeStats()

pkalever: > The design for Heal() and a HealthCheck() function need to go to csi-addons/spec.

@nixpanic I expect I will have more clarity on Heal() and a HealthCheck() functions while I work on this RFE, so I hope to start on the draft implementation first and then send a PR to csi-addons/spec, only after the spec gets merged I will send the actual controller PR.

Hope that won't hurt.

docs/design/proposals/healer-controller.md
pkalever: This sounds reasonable.

`HEALTH_OK`, `HEALTH_UNSUPPORTED`, `HEALTH_BAD` are some kind of volume attributes that CSI driver sets on the persistent volume?

In my previous comment, I mean something like:
* capability response returns some string that holds
```
{
  Mounter1: NEED_HEALING
  Mounter2: NEED_MONITORING
  Mounter3: UNKNOWN
}
```
* If the PV attribute mounter matches with `Mounter1`, attempt to heal it.
* If it is of type `Mounter2` only check the health,
* If it is of type `Mounter3` do not attempt anything from the healer controller.

docs/design/proposals/healer-controller.md
```diff
@@ -0,0 +1,116 @@
...
+    * For each `VolumeAttachment` get the respective PersistentVolume information and check the criteria of PersistentVolume Bound (also we can have configuration options like mounter type)
+    * Send the RPC request for side-car to check if the volume is healthy along with the PersistentVolume details and may attempt to heal the volume if it is found abnormal
+    * Log the final response returned from the side-car
+
+## CSI-Addons side-car
+* When a nodeplugin pod is restarted the side-car will create [CSIAddonsNode](https://github.com/csi-addons/kubernetes-csi-addons/blob/main/api/v1alpha1/csiaddonsnode_types.go) object.
+* The side-car works as relay, upon receiving a gRPC call from healer controller, the CSI-Addons side-car will then send a VolumeMonitor request to CSI driver. The request contains some internal details like secrets, staging path, volume-id etc.
+* Upon receiving the VolumeMonitor request, the CSI driver check volume's mounting conditions collected via the NodeGetVolumeStats RPC and in case if any volume condition is reported abnormal, it will act on it by performing required healing and finally returns back the response to the controller.
```
Rakshith-R, line 43: This is still a two step process ? 

The CSI-driver can itself simply determine if volume is healthy and just return ALL_OK or heal it if it is not.

docs/design/proposals/healer-controller.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Add Controller Changes
```
Rakshith-R, line 45: I guess we can discuss rpc changes later on too.

Rakshith-R (changes requested):
Just one more question and I think we do without gRPC description in the doc ? 
Everything else look good to me.

Rakshith-R (approved):

docs/design/proposals/healer-controller.md
```diff
@@ -0,0 +1,43 @@
+# Introduction
+There is [external health monitoring](https://github.com/kubernetes-csi/external-health-monitor/) which does mostly similar stuff, but actions are not in the scope of it, all it can do is log the health condition of the mounts.
+
+# Volume Healer Controller
+
+**Now the obvious question we expect is "what do we achieve with the new controller?"**
+* Achieve a way to monitor the health of volumes.
+* Ceph-CSI has healer component which take cares of restarting the userspace processes when the node plugin is restarted, but in many places it uses Kubernetes specific code, with this controller model we are moving out the Kubernetes specific code out, and help make Ceph-CSI orchestrator independent.
```
Madhu-1, line 8: cephcsi name is not required here as its a general component.

docs/design/proposals/healer-controller.md
```diff
@@ -0,0 +1,43 @@
...
+* Watch for the `CSIAddonsNode` object, if the object is created/modified by the side-car do below operations
+    * Once the connection is established, fetch the capabilities list and check for nodes that supports `VOLUME_HEALER` capability
+    * Fetch all the VolumeAttachment` list
+    * Filter the `VolumeAttachment` list through matching driver name, node name and status attached
+    * For each `VolumeAttachment` get the respective PersistentVolume information and check the criteria of PersistentVolume Bound (also we can have configuration options like mounter type)
+    * Send the RPC request for side-car to check if the volume is healthy along with the PersistentVolume details and may attempt to heal the volume if it is found abnormal
+    * Log the final response returned from the side-car
+
+## CSI-Addons side-car
+* When a nodeplugin pod is restarted the side-car will create [CSIAddonsNode](https://github.com/csi-addons/kubernetes-csi-addons/blob/main/api/v1alpha1/csiaddonsnode_types.go) object.
```
Madhu-1, line 41: will create or update the object as per https://github.com/csi-addons/kubernetes-csi-addons/blob/57ed4f6211c0847eecead1d6f7baaee15f8e356d/sidecar/internal/csiaddonsnode/csiaddonsnode.go#L135-L138?

docs/design/proposals/healer-controller.md
```diff
@@ -0,0 +1,43 @@
...
+* The side-car works as relay, upon receiving a gRPC call from healer controller, the CSI-Addons side-car will then send a VolumeMonitor request to CSI driver. The request contains some internal details like secrets, staging path, volume-id etc.
```
Madhu-1, line 42: `gRPC` to `RPC` as we have used RPC in many places?

docs/design/proposals/healer-controller.md
pkalever: I'm convinced with a comment by Niels earlier:

```
I think you are concerned that the CSI-driver container in the pod restarted, and not the pod was re-created.
Only in that case, the CSIAddonsNode object would exist. It is not clear to me if this is something that can happen in reality.
``

File: docs/design/proposals/healer-controller.md (hunk 21 above)
Comment (pkalever) [reply docs/design/proposals/healer-controller.md:41]: I'm convinced with a comment by Niels earlier:

```
I think you are concerned that the CSI-driver container in the pod restarted, and not the pod was re-created.
Only in that case, the CSIAddonsNode object would exist. It is not clear to me if this is something that can happen in reality.
```

Madhu-1: if the objects already exist there should be an update operation to let the controller know that sidecar is restarted. am missing this point.

pkalever: @Madhu-1 
As far as I understand from Niels previous comments, the object will be existing only when the container is restarted and it will be absent when the whole pod is restarted. If that is the case, when do we expect the container to restart, practically? Hence is it worth mentioning it here?

Madhu-1: Yes, correct i don't have a practical answer for this one either. i was thinking object update might help which is not present currently.

pkalever: let's take it later? maybe based on future discussions?

Madhu-1 (approved):

Rakshith-R (approved):

docs/design/proposals/healer-controller.md
nixpanic: Interactions between CSI-Addons and the CSI-driver need to be documented in a [csi-addons/spec](https://github.com/csi-addons/spec). The protocol then describes what a CSI-driver needs to implement for supporting the healer operations. This will probably include a description for capabilities of the CSI-driver. Certain operations on a volume can return attributes that healing is not supported for the volume, but other volumes from the same driver might support more features.

I do not think CSI-Addons should know about internals of the CSI-driver, so _mounters_ would not be needed,

nixpanic (changes requested):
I would love to see this in combination with a proposal in csi-addons/spec, with that it is much clearer how an interface between CSI-Addons and the CSI-driver looks like.

Madhu-1:
closing this as stale, feel free to reopen when required.

PR #104, opened 2022-01-27
controller/reclaimspacejob: skip making node req if nodeID is empty

Rakshith-R:
Node reclaimspace request should not be made
when nodeID is empty indicating the pvc is
not mounted to any pod and no volumeattachment
object for that PVC was found.

### DIFF
--- a/controllers/reclaimspacejob_controller.go
+++ b/controllers/reclaimspacejob_controller.go
@@ -173,6 +173,13 @@ type targetDetails struct {
 	nodeID     string
 }
 
+// canNodeReclaimSpace returns true if nodeID is not empty,
+// indicating volume is mounted to a pod and node reclaimspace
+// request can be sent.
+func (td *targetDetails) canNodeReclaimSpace() bool {
+	return td.nodeID != ""
+}
+
 // reconcile performs time based validation, fetches required details and makes
 // grpc request for controller and node reclaim space operation.
 func (r *ReclaimSpaceJobReconciler) reconcile(
@@ -211,15 +218,22 @@ func (r *ReclaimSpaceJobReconciler) reconcile(
 		return err
 	}
 
-	nodeFound, nodeReclaimedSpace, err := r.nodeReclaimSpace(ctx, logger, target)
-	if err != nil {
-		logger.Error(err, "Failed to make node request")
-		setFailedCondition(
-			&rsJob.Status.Conditions,
-			fmt.Sprintf("Failed to make node request: %v", util.GetErrorMessage(err)),
-			rsJob.Generation)
+	var (
+		nodeFound          = false
+		nodeReclaimedSpace *int64
+	)
+	if target.canNodeReclaimSpace() {
+		nodeFound = true
+		nodeReclaimedSpace, err = r.nodeReclaimSpace(ctx, logger, target)
+		if err != nil {
+			logger.Error(err, "Failed to make node request")
+			setFailedCondition(
+				&rsJob.Status.Conditions,
+				fmt.Sprintf("Failed to make node request: %v", util.GetErrorMessage(err)),
+				rsJob.Generation)
 
-		return err
+			return err
+		}
 	}
 
 	controllerFound, controllerReclaimedSpace, err := r.controllerReclaimSpace(ctx, logger, target)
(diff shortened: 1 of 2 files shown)


Madhu-1:
@Rakshith-R can you please provide more tails when the node  ID will be empty?

controllers/reclaimspacejob_controller.go
```diff
@@ -376,6 +376,11 @@ func (r *ReclaimSpaceJobReconciler) nodeReclaimSpace(
 	ctx context.Context,
 	logger *logr.Logger,
 	target *targetDetails) (bool, *int64, error) {
+	if target.nodeID == "" {
```
nixpanic, line 379: it is cleaner to skip calling this function altogether. do the check where `nodeReclaimSpace()` is called?

Madhu-1: is this means VA object is created but no NodeID is added to it?  explaining in what case it can happen will be helpful.

Rakshith-R:
> 


> @Rakshith-R can you please provide more tails when the node ID will be empty?

When pvc is  not mounted to any pod, no volumeattachments will be found, hence nodeID will be empty

controllers/reclaimspacejob_controller.go
nixpanic: maybe add a helper `func (td *targetDetails) canNodeReclaimSpace() bool {}` so that the check for `target.nodeID` does not need to be placed in the execution flow of `reconcile()`. You might be able to drop the 1st return value from `nodeReclaimSpace()` that way too?

Rakshith-R: > is this means VA object is created but no NodeID is added to it? explaining in what case it can happen will be helpful.

It means no VA object exists for that pvc, I'll add comments




> maybe add a helper `func (td *targetDetails) canNodeReclaimSpace() bool {}` so that the check for `target.nodeID` does not need to be placed in the execution flow of `reconcile()`. You might be able to drop the 1st return value from `nodeReclaimSpace()` that way too?

okay :+1:

controllers/reclaimspacejob_controller.go
```diff
@@ -369,20 +387,19 @@ func (r *ReclaimSpaceJobReconciler) controllerReclaimSpace(
...
+	target *targetDetails) (*int64, error) {
 	clientName, nodeClient := r.getRSClientWithCap(
 		target.driverName,
 		target.nodeID,
 		identity.Capability_ReclaimSpace_ONLINE)
 	if nodeClient == nil {
-		logger.Info("Node Client not found")
-		return false, nil, nil
+		err := fmt.Errorf("Node Client not found")
+		return nil, err
```
Madhu-1, line 402:
``` suggestion
return nil, errors.New("Node Client not found")
```

controllers/reclaimspacejob_controller_test.go
```diff
@@ -176,3 +176,42 @@ func TestCalculateReclaimedSpace(t *testing.T) {
 		})
 	}
 }
+
+func Test_targetDetails_canNodeReclaimSpace(t *testing.T) {
```
Madhu-1, line 180:
```suggestion
TestcanNodeReclaimSpace
```

controllers/reclaimspacejob_controller_test.go
```diff
@@ -176,3 +176,42 @@ func TestCalculateReclaimedSpace(t *testing.T) {
...
+	type fields struct {
+		driverName string
+		pvName     string
```
Madhu-1, line 183: pv name and drivername are not required?

Rakshith-R: > pv name and drivername are not required?

optional, added them now to keep it consistent in the tests.

controllers/reclaimspacejob_controller.go
```diff
@@ -369,20 +387,18 @@ func (r *ReclaimSpaceJobReconciler) controllerReclaimSpace(
...
-	target *targetDetails) (bool, *int64, error) {
+	target *targetDetails) (*int64, error) {
 	clientName, nodeClient := r.getRSClientWithCap(
 		target.driverName,
 		target.nodeID,
 		identity.Capability_ReclaimSpace_ONLINE)
 	if nodeClient == nil {
-		logger.Info("Node Client not found")
-		return false, nil, nil
+		return nil, errors.New("Node Client not found")
```
Madhu-1, line 401: Error should not start with capital letters

Madhu-1 (approved):

Madhu-1 (approved):

nixpanic (commented):
just a few nits for your consideration

controllers/reclaimspacejob_controller.go
```diff
@@ -211,15 +218,26 @@ func (r *ReclaimSpaceJobReconciler) reconcile(
...
-
-		return err
+	var (
+		nodeFound          = false
+		nodeReclaimedSpace *int64
+	)
+	// PVC may not be mounted to a pod and volumeattachment for PVC
+	// may not be found due to which nodeID will be empty.
+	// Make node reclaimspace request only if
+	// nodeID is not empty.
```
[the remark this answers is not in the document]
Rakshith-R: > the comment is not really needed to be this verbose. Only the comment for the function should explain about `nodeID`, the check may change in the future.

I agree. The function comment should be sufficient.

controllers/reclaimspacejob_controller_test.go
```diff
@@ -176,3 +176,46 @@ func TestCalculateReclaimedSpace(t *testing.T) {
...
+	}
+	tests := []struct {
+		name   string
+		fields fields
+		want   bool
+	}{
+		{
+			name: "empty nodeID",
+			fields: fields{
+				driverName: "example.csi.com",
```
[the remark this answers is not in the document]
Rakshith-R: > csi.com exists, use `csi.example.com` instead

done

controllers/reclaimspacejob_controller_test.go
```diff
@@ -176,3 +176,46 @@ func TestCalculateReclaimedSpace(t *testing.T) {
 		})
 	}
 }
+
+func TestCanNodeReclaimSpace(t *testing.T) {
+	type fields struct {
```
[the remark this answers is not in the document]
Rakshith-R: > why introduce `fields`? just use `targetDetails` in the `tests` struct

auto generated test, It uses targetDetails directly now.

Madhu-1 (approved):

yati1998 (commented):
LGTM now.

nixpanic (approved):

Issue #105, opened 2022-01-27
Generate the manifest files for crds and controller manager


yati1998:
We need to generate the raw yamls required to install and deploy the controller so that those yamls can be directly used by others(basically rook for now) to deploy the controller and run the operation.

PR #106, opened 2022-01-28
controller: adds yaml to install and deploy controller

yati1998:
this commit adds raw yaml generated using kustomize
in deploy/crds.yaml which can be used by others to
directly deploy the controller

Fixes: #105

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,8 +83,10 @@ help: ## Display this help.
 ##@ Development
 
 .PHONY: manifests
-manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
+manifests: controller-gen kustomize ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
 	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="{./api/...,./cmd/...,./controllers/...,./sidecar/...}" output:crd:artifacts:config=config/crd/bases
+	cd config/manager && $(KUSTOMIZE) edit set image controller=${CONTROLLER_IMG} $(KUSTOMIZE_RBAC_PROXY)
+	$(KUSTOMIZE) build config/default > deploy/controller/setup-controller.yaml
 
 # generate the <package-name>.clusterserviceversion.yaml
 config/manifests/bases/$(PACKAGE_NAME).clusterserviceversion.yaml: config/manifests/bases/clusterserviceversion.yaml.in
--- a/deploy/controller/setup-controller.yaml
+++ b/deploy/controller/setup-controller.yaml
@@ -1,0 +1,1050 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    control-plane: controller-manager
+  name: csi-addons-system
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.8.0
+  creationTimestamp: null
+  name: csiaddonsnodes.csiaddons.openshift.io
+spec:
+  group: csiaddons.openshift.io
+  names:
+    kind: CSIAddonsNode
+    listKind: CSIAddonsNodeList
+    plural: csiaddonsnodes
+    singular: csiaddonsnode
+  scope: Namespaced
+  versions:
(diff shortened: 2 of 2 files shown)


yati1998:
@nixpanic @Rakshith-R please do add your reviews. Thanks, Rakshith for helping out with makefile.

Makefile
```diff
@@ -83,8 +83,10 @@ help: ## Display this help.
 ##@ Development
 
 .PHONY: manifests
-manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
+manifests: controller-gen kustomize## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
 	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="{./api/...,./cmd/...,./controllers/...,./sidecar/...}" output:crd:artifacts:config=config/crd/bases
+	cd config/manager && $(KUSTOMIZE) edit set image controller=${CONTROLLER_IMG} $(KUSTOMIZE_RBAC_PROXY)
+	$(KUSTOMIZE) build config/default > deploy/controller/setup-controller.yaml
```
Rakshith-R, line 89: With this, yaml will be auto-generated whenever manifests is run and can be verified in the ci https://github.com/csi-addons/kubernetes-csi-addons/runs/4977203144?check_suite_focus=true#step:4:10

Rakshith-R (dismissed):

Makefile
Madhu-1, line 89: we should do `kubectl dry-run`  to verify the template's syntax is correct.

deploy/controller/setup-controller.yaml
```diff
@@ -0,0 +1,1050 @@
+apiVersion: v1
```
Madhu-1, line 1: is it possible to have 3 files, one for CRD, one for RBAC, and one for deployment? `1050` lines in a single file seem heavy for me.

deploy/controller/setup-controller.yaml
```diff
@@ -0,0 +1,1050 @@
...
+        - containerPort: 8443
+          name: https
+          protocol: TCP
+      - args:
+        - --health-probe-bind-address=:8081
+        - --metrics-bind-address=127.0.0.1:8080
+        - --leader-elect
+        command:
+        - /manager
+        image: quay.io/csiaddons/k8s-controller:latest
```
Madhu-1, line 1024: hope this will get replaces when we do a release and release branch will contain the proper tagged image.

yati1998: So, its fine we do edit this in makefile. I created the file with this image. We can update it with latest build.

deploy/controller/setup-controller.yaml
yati1998: Yeah, we can, but here we are just building the config/default which generates all these at a place. I would prefer to have it as a TODO as it required more modifications.

yati1998:
@Madhu-1 @nixpanic @Rakshith-R Let's go with one file for now. This file is auto-generated via Kustomize and splitting it breaks some or the other things. So I guess we can add a TODO in Makefile itself so that we can have the rook implementation done. Please do share your thoughts on this and will update the PR accordingly.

Makefile
```diff
@@ -83,8 +83,10 @@ help: ## Display this help.
 ##@ Development
 
 .PHONY: manifests
-manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
+manifests: controller-gen kustomize## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
```
nixpanic, line 86: use a space between `kustomize##`, not sure if this does what is intended

deploy/controller/setup-controller.yaml
nixpanic: This should probably become an artifact when we do a release. Creating a tag in git can then generate the yaml, and post it as part of the release (somehow). It requires adding one or more jobs to the `build-push` workflow we already have.

nixpanic (changes requested):
nit about the makefile comment after `kustomize`, looks good to me otherwise

Madhu-1 (approved):

Rakshith-R (approved):

Yuggupta27 (approved):

nixpanic (approved):

PR #122, opened 2022-02-28
use a unique label for csi-addons resources

Madhu-1:
csi-addon operator is using the default label and labelSelector provided by operator-sdk `control-plane: controller-manager` for its resources, this is a generic label and it can cause problems when multiple resources use the same labels changing the label to unique one.

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -4,18 +4,18 @@ metadata:
   name: controller-manager
   namespace: system
   labels:
-    control-plane: controller-manager
+    app.kubernetes.io/name: csi-addons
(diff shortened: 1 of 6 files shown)


Madhu-1:
cc @iamniting

iamniting (approved):

nixpanic (approved):
The used `app.kubernetes.io/name` seems to be mentioned in the [recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/), so this makes sense to me. Thanks!

Rakshith-R (approved):

PR #152, opened 2022-04-27
NetworkFence: Explicitly mentioned Fenced and Unfenced state

raghavendrabhat:
Currently, the assumption is that, the resource by default
strives to ensure that the CIDRs mentioned in the spec are
fenced. Deletion of the resource makes those CIDRs to have
been unfenced.

This has some issues when fencing is consumed for operations
such as Disaster Recovery. Hence, changed the CR to have
explicit state as either fence or unfence.

Co-authored-by: Raghavendra Talur <raghavendra.talur@gmail.com>
Signed-off-by: Raghavendra M <raghavendra@redhat.com>


api/v1alpha1/networkfence_types.go
```diff
@@ -20,6 +20,16 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+type FenceState string
+
+const (
+	// PrimaryState represents the Primary replication state
```
yati1998, line 26: I guess the comments need to be changed, it does not match with the variables

api/v1alpha1/networkfence_types.go
```diff
@@ -20,6 +20,16 @@ import (
...
+	Fenced FenceState = "Fenced"
+
+	// SecondaryState represents the Secondary replication state
```
yati1998, line 29: Same here

yati1998 (changes requested):

api/v1alpha1/networkfence_types.go
```diff
@@ -46,6 +56,11 @@ type NetworkFenceSpec struct {
 	// +kubebuilder:validation:Required
 	Driver string `json:"driver"`
 
+	// FenceState contains the desired state for the CIDRs
+	// mentioned in the Spec. i.e. fenced or unfenced
+	// +kubebuilder:validation:Required
```
Rakshith-R, line 61:
```suggestion
	// FenceState contains the desired state for the CIDRs
	// mentioned in the Spec. i.e. fenced or unfenced
	// +kubebuilder:validation:Required
```

config/crd/bases/csiaddons.openshift.io_networkfences.yaml
```diff
@@ -57,6 +57,10 @@ spec:
               driver:
                 description: Driver contains  the name of CSI driver.
                 type: string
+              fenceState:
+                description: FenceState contains the desired state for the CIDRs mentioned
+                  in the Spec. i.e. fenced or unfenced
```
Rakshith-R, line 62:
```suggestion
                  in the Spec. i.e. Fenced or Unfenced
```

controllers/networkfence_controller.go
```diff
@@ -113,37 +118,66 @@ func (r *NetworkFenceReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 	// check if the networkfence object is getting deleted, if yes, then send
 	// UnfenceCluster request to the driver.
 	// After the UnfenceCluster request is processed, the finalizer will be removed
 	// so that the networkfence object can be deleted gracefully.
-	if !nwFence.GetDeletionTimestamp().IsZero() {
+	if !nf.instance.GetDeletionTimestamp().IsZero() {
 		if util.ContainsInSlice(nwFence.GetFinalizers(), networkFenceFinalizer) {
 
-			err := r.unfenceClusterNetwork(ctx, nwFence, client, logger)
+			err := nf.deleteNetworkFenceResource()
```
Rakshith-R, line 136: Call removefinalizer directly ? 

No need to Unfence here?

controllers/networkfence_controller.go
```diff
@@ -113,37 +118,66 @@ func (r *NetworkFenceReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 		logger.Info("NetworkFence object is terminated, skipping reconciliation")
 		return ctrl.Result{}, nil
 	}
 
-	err = r.fenceClusterNetwork(ctx, nwFence, client, logger)
+	err = nf.processFencing()
 	if err != nil {
 		logger.Error(err, "failed to fence cluster network")
+		err = nf.updateStatus(csiaddonsv1alpha1.FencingOperationResultFailed, "fencing operation failed")
 		return ctrl.Result{}, err
```
Rakshith-R, line 151:
```suggestion
		UpdateErr = nf.updateStatus(csiaddonsv1alpha1.FencingOperationResultFailed, "fencing operation failed")
		if UpdateErr!=nil{
			logger.Error(UpdateErr, "failed to update status")
		}
		return ctrl.Result{}, err
```

The current format will not retrigger reconcile

controllers/networkfence_controller.go
```diff
@@ -153,91 +187,108 @@ func (r *NetworkFenceReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 	}
-	logger.Info("UnfenceClusterNetwork Request Succeeded")
 
+	nf.logger.Info("UnFenceClusterNetwork Request Succeeded")
+
+	return nil
+}
+
+// unfenceClusterNetwork sends a UnfenceClusterNetwork request and removes the finalizer on success.
+func (nf *NetworkFenceInstance) deleteNetworkFenceResource() error {
```
Rakshith-R, line 256: No need for this function at all? 
Call removeFinalizerFromNF directly ?

Rakshith-R (changes requested):

api/v1alpha1/networkfence_types.go
raghavendrabhat: ACK

api/v1alpha1/networkfence_types.go
raghavendrabhat: ACK

config/crd/bases/csiaddons.openshift.io_networkfences.yaml
raghavendrabhat: ACK

controllers/networkfence_controller.go
raghavendrabhat: Yes, no need to unfence here. Only reason to do it in a separate function to ensure that everything related to deletion of the resource is contained in a function. As of today with this PR, we are only removing the finalizer in the deletion handling. In future if something more needs to be done (cant think of any as of now though), then the function deleteNetworkFenceResource would be the one doing the additional things.

Let me know your thoughts.

controllers/networkfence_controller.go
nixpanic: I would suggest to add `deleteNetworkFenceResource()` or similar in the future when there is a need for it

nixpanic: Or, alternatively, if it is clearer, rename `removeFinalizerFromNF` to `deleteNetworkFenceResource`?

Rakshith-R: I don't think delete as function name makes sense,  maybe handleNetworkFenceDeletion() ?

api/v1alpha1/networkfence_types.go
```diff
@@ -46,6 +56,12 @@ type NetworkFenceSpec struct {
...
+	// +kubebuilder:validation:Enum=Fence;UnFenced
```
nixpanic, line 62: `UnFenced` -> `Unfenced` ?

config/crd/bases/csiaddons.openshift.io_networkfences.yaml
```diff
@@ -57,6 +57,13 @@ spec:
...
+                enum:
+                - Fence
+                - UnFenced
```
nixpanic, line 65: This string is different from `Unfenced FenceState`

controllers/networkfence_controller.go
```diff
@@ -113,37 +118,68 @@ func (r *NetworkFenceReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+
+	if err = nf.updateStatus(csiaddonsv1alpha1.FencingOperationResultSucceeded, "fencing operation successful"); err != nil {
+		logger.Error(err, "failed to update status")
 		return ctrl.Result{}, err
 	}
 
 	return ctrl.Result{}, nil
 }
 
+type NetworkFenceInstance struct {
```
nixpanic, line 163: Missing documentation for this type and its attributes.

Storing `context.Context` in a struct is not common, it usually gets passed as an argument to functions.

raghavendrabhat: Will be adding the doc for this type.

This struct is used here as a kind of class where the attributes are the members of it and there are methods (such as fenceClusterNetwork, unfenceClusterNetwork) that need the attributes from the struct "NetworkFenceInstance". And this helped in keeping the functions simple. Also, the struct NetworkFenceInstace has not been passed as an argument to any function. 

This is based on some of the controller implementations in the ramen project which simplified the implementation.

https://github.com/RamenDR/ramen/blob/main/controllers/drplacementcontrol.go#L43
https://github.com/RamenDR/ramen/blob/main/controllers/volumereplicationgroup_controller.go#L316

However, I am fine with removing the attribute "ctx context.Context" from the struct "NetworkFenceInstance" and instead pass "ctx context.Context" as the argument to the functions where necessary.

Please let me know your thoughts.

api/v1alpha1/networkfence_types.go
raghavendrabhat: Yes. Will correct it.

controllers/networkfence_controller.go
```diff
@@ -113,37 +118,68 @@ func (r *NetworkFenceReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+		instance:         nwFence,
+		controllerClient: client,
+	}
+
+	// check if the networkfence object is getting deleted and handle it.
+	if !nf.instance.GetDeletionTimestamp().IsZero() {
 		if util.ContainsInSlice(nwFence.GetFinalizers(), networkFenceFinalizer) {
 
-			err := r.unfenceClusterNetwork(ctx, nwFence, client, logger)
+			err := nf.removeFinalizerFromNetworkFence()
```
ShyamsundarR, line 133: (consider) Delete should ensure state is as desired before deletion. IOW, a fence resource request that did not succeed and is being deleted, may leave the system fenced or unfenced (and vice versa).

For example, if the resource status for the current generation being deleted reports desired state (due to a prior reconcile) then proceed with deletion, else retry the operation as we could have timed out in the past, and the system would be in an inconsistent state (some parts maybe fenced/unfenced as per the spec.FenceState).

This is not as critical as a CreateVolume or snapshot request that may leak resources if deleted prematurely (IOW, not ensuring existing system state), we may still get into corner cases where this maybe leave the system inconsistent.

raghavendrabhat: Yes. Perhaps it can be done in a separate PR? The current way of handling is similar to how it was being handled before before this PR. With current changes, it may not be exactly same as how it was before this PR as before for delete request, the reconciler would attempt unfencing. However, even with that, the issue you mentioned would exist as it did not check the current state of the world before attempting unefnce operation as part of deletion.

api/v1alpha1/networkfence_types.go
Yuggupta27: Correction : "validation:Enum=Fence;UnFenced" -> " validation:Enum=**Fenced**;UnFenced"

controllers/networkfence_controller.go
```diff
@@ -113,37 +118,71 @@ func (r *NetworkFenceReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+
+	// check if the networkfence object is getting deleted and handle it.
+	if !nf.instance.GetDeletionTimestamp().IsZero() {
 		if util.ContainsInSlice(nwFence.GetFinalizers(), networkFenceFinalizer) {
 
-			err := r.unfenceClusterNetwork(ctx, nwFence, client, logger)
+			err := nf.removeFinalizerFromNetworkFence(ctx)
 			if err != nil {
-				logger.Error(err, "failed to unfence cluster network")
+				logger.Error(err, "failed to remove finalizer on network fence resource")
```
Yuggupta27, line 134:
```suggestion
				logger.Error(err, "failed to remove finalizer on NetworkFence resource")
```

deploy/controller/crds.yaml
```diff
@@ -164,6 +164,13 @@ spec:
               driver:
                 description: Driver contains  the name of CSI driver.
                 type: string
+              fenceState:
+                description: FenceState contains the desired state for the CIDRs mentioned
+                  in the Spec. i.e. fenced or unfenced
```
Yuggupta27, line 169:
```suggestion
                  in the Spec. i.e. Fenced or Unfenced
```

Yuggupta27 (changes requested):

api/v1alpha1/networkfence_types.go
raghavendrabhat: ACK

controllers/networkfence_controller.go
nixpanic: Ok, I think that is fine. Thanks for the explanation.

deploy/controller/crds.yaml
```diff
@@ -164,6 +164,13 @@ spec:
               driver:
                 description: Driver contains  the name of CSI driver.
                 type: string
+              fenceState:
```
nixpanic, line 167: Should this be included in the `additionalPrinterColumns`?

I guess it also is required to update the version?

raghavendrabhat: ACK to "additionalPrinterColumns"

Question about the version. Since the version is still v1alpha1 for the api and ramen being the only consumer of NetworkFence resource as of now, do you thing it is necessary to update the version?

controllers/networkfence_controller.go
```diff
@@ -63,6 +63,11 @@ func validateNetworkFenceSpec(nwFence *csiaddonsv1alpha1.NetworkFence) error {
 	if nwFence.Spec.Cidrs == nil {
 		return errors.New("required parameter cidrs is not specified")
 	}
+
+	if nwFence.Spec.FenceState != csiaddonsv1alpha1.Fenced && nwFence.Spec.FenceState != csiaddonsv1alpha1.Unfenced {
+		return errors.New("invalid fence state. fence state can only be either of Fenced or Unfenced")
+	}
```
Madhu-1, line 69: Will this will be hit as we already have validation at the CR state?

controllers/networkfence_controller.go
```diff
@@ -113,37 +118,71 @@ func (r *NetworkFenceReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+
 		logger.Info("NetworkFence object is terminated, skipping reconciliation")
 		return ctrl.Result{}, nil
 	}
 
-	err = r.fenceClusterNetwork(ctx, nwFence, client, logger)
+	err = nf.processFencing(ctx)
 	if err != nil {
 		logger.Error(err, "failed to fence cluster network")
+		updateStatusErr := nf.updateStatus(ctx, csiaddonsv1alpha1.FencingOperationResultFailed, "fencing operation failed")
```
Madhu-1, line 146: message need to be derived from the fencing operation as it supports both fencing and unfencing.

controllers/networkfence_controller.go
```diff
@@ -113,37 +118,71 @@ func (r *NetworkFenceReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+		instance:         nwFence,
+		controllerClient: client,
+	}
+
+	// check if the networkfence object is getting deleted and handle it.
+	if !nf.instance.GetDeletionTimestamp().IsZero() {
 		if util.ContainsInSlice(nwFence.GetFinalizers(), networkFenceFinalizer) {
 
-			err := r.unfenceClusterNetwork(ctx, nwFence, client, logger)
+			err := nf.removeFinalizerFromNetworkFence(ctx)
```
Madhu-1, line 132: This seems wrong the finalizers should be removed only after unfencing as we need to support the fencing state also? or else in case of an error we might end up deleting the CR but unfencing is not done.

controllers/networkfence_controller.go
```diff
@@ -152,92 +191,101 @@ func (r *NetworkFenceReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+// fence or unfence based on the spec.
+func (nf *NetworkFenceInstance) processFencingRequest(ctx context.Context) error {
 	// send FenceClusterNetwork request.
-	logger.Info("FenceClusterNetwork Request")
-	timeoutContext, cancel := context.WithTimeout(ctx, r.Timeout)
+	if nf.instance.Spec.FenceState == csiaddonsv1alpha1.Fenced {
+		nf.logger.Info("FenceClusterNetwork Request")
+	} else {
+		nf.logger.Info("UnFenceClusterNetwork Request")
+	}
```
Madhu-1, line 215: this is not helpful logging. log in at Reconcile function with namespaced name.

api/v1alpha1/networkfence_types.go
```diff
@@ -46,6 +56,12 @@ type NetworkFenceSpec struct {
 	// +kubebuilder:validation:Required
 	Driver string `json:"driver"`
 
+	// FenceState contains the desired state for the CIDRs
+	// mentioned in the Spec. i.e. Fenced or Unfenced
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Enum=Fenced;Unfenced
```
Madhu-1, line 62: won't this be a backward compatibility issue for existing CRs?

deploy/controller/crds.yaml
nixpanic: I don't know... That is probably something @Yuggupta27 can say a little more about.

controllers/networkfence_controller.go
nixpanic: maybe if a user had the previous version of the CRD?

Madhu-1: it breaks the backward compatibility, isn't it? I assume with this change upgrades will also fail as the current spec will not have these fields. @umangachapagain I saw somewhere this recently happened in ocs. can you confirm?

controllers/networkfence_controller.go
raghavendrabhat: Yes, if a fencing CR is deleted after setting its state to "Fenced" and having the set of CIDRs fenced, then the set of CIDRs would be left in a fenced state.

It can be interpreted in 2 ways I think. One is to ensure that the cluster has been unfenced before the resource deletion. Another interpretation can be to say that all it has to do is to ensure its DSoW matches with ASoW and the consumer of the resource is aware of what they are doing with this resource.

From my perspective, I think it would be better to make the reconciler attempt to match DSoW with ASoW instead of assuming that when a resource is deleted, the cluster should be unfenced. 

However, we can introduce an option to say "unfence on delete" which upon being true, will lead to the behavior you suggested. i.e. upon deletion of a NetworkFence CR, the reconciler will first unfence the CIDRs from spec and then deletes the resource. This way, the consumer of this resource would have more control over how they can use the resource.

Let me know your thoughts.

controllers/networkfence_controller.go
raghavendrabhat: ACK

controllers/networkfence_controller.go
raghavendrabhat: ACK.

controllers/networkfence_controller.go
raghavendrabhat: Based on the comment [1]  as both NetworkFence CRD and ramen (the consumer of NetworkFence CR as of now) are in v1alpha1 versions, does it require any changes from a compatibility perspective at this point in time?

[1] https://github.com/csi-addons/kubernetes-csi-addons/pull/152#discussion_r864862022

controllers/networkfence_controller.go
Madhu-1: >It can be interpreted in 2 ways I think. One is to ensure that the cluster has been unfenced before the resource deletion. Another interpretation can be to say that all it has to do is to ensure its DSoW matches with ASoW and the consumer of the resource is aware of what they are doing with this resource.

am not aware of what is DSoW and ASoW can be please explain.

> From my perspective, I think it would be better to make the reconciler attempt to match DSoW with ASoW instead of assuming that when a resource is deleted, the cluster should be unfenced.

If the resource is deleted why does someone wants to keep the cluster in a fenced state?

>However, we can introduce an option to say "unfence on delete" which upon being true, will lead to the behavior you suggested. i.e. upon deletion of a NetworkFence CR, the reconciler will first unfence the CIDRs from spec and then deletes the resource. This way, the consumer of this resource would have more control over how they can use the resource.

IMO We are complicating it by introducing more options to the users. i would like to do unfence when the CR is deleted and the state is fenced to keep it backward compatible and IMO it's an easy way to do fence/unfence.

controllers/networkfence_controller.go
Madhu-1: add a default state to the CRD if the value is not set so it will be backward compatible

nixpanic: I am still not sure how this is expected to work. My knowledge on CRD versioning is very minimal, but I would like to know if the following causes issues:

1. current version of the CRD is deployed, with the current kubernetes-csi-addons controller (without this PR)
2. Ramen gets deployed in a weeks time, expecting the new `FenceState` attribute in the CRD

As the new attribute does not exist in the CRD, Ramen would not be able to access it? How can Ramen require the new version of the CRD/API, or let the user know to update the `NetworkFence` CRD?

controllers/networkfence_controller.go
raghavendrabhat: > > It can be interpreted in 2 ways I think. One is to ensure that the cluster has been unfenced before the resource deletion. Another interpretation can be to say that all it has to do is to ensure its DSoW matches with ASoW and the consumer of the resource is aware of what they are doing with this resource.
> 
> am not aware of what is DSoW and ASoW can be please explain.

Oops. I should have used the complete form.

DSoW: Desired State of World
ASoW: Actual State of World

> 
> > From my perspective, I think it would be better to make the reconciler attempt to match DSoW with ASoW instead of assuming that when a resource is deleted, the cluster should be unfenced.
> 
> If the resource is deleted why does someone wants to keep the cluster in a fenced state?

This is mainly assuming the declarative model of doing things where the reconciler tries to match the Actual State of World (ASOW) with the Desired State of World(DSoW). 

I assume one of the feedbacks given to this PR at [1] was to ensure DSoW meets ASoW before deleting the resource ensure that it has met the DSoW mentioned in the spec. 

I think this example where this behavior of unfencing upon deletion can cause problems.

Lets assume we have 3 kube clusters A, B, C sharing the same storage cluster. Now, A goes down and it is fenced off via B. Now before we can clean up A lets say B has to be decommissioned. And this is where ideally the fencing resource from B has to be migrated to the surviving cluster C along with other things. Here the fencing resource is deleted as part of cleanup of B. Hence even though A should be in a fenced state, it can get unfenced if the deletion from B happens after the migration of the fencing CR to C. And this will be the order of execution (i.e. migrate to C and then clean up B) to ensure smooth migration causing A to be silently get unfenced.

Let me know if I am missing something.

[1] https://github.com/csi-addons/kubernetes-csi-addons/pull/152#discussion_r861995122



> 
> > However, we can introduce an option to say "unfence on delete" which upon being true, will lead to the behavior you suggested. i.e. upon deletion of a NetworkFence CR, the reconciler will first unfence the CIDRs from spec and then deletes the resource. This way, the consumer of this resource would have more control over how they can use the resource.
> 
> IMO We are complicating it by introducing more options to the users. i would like to do unfence when the CR is deleted and the state is fenced to keep it backward compatible and IMO it's an easy way to do fence/unfence.

raghavendra-talur: Like I said in a different thread, we should not be taking backward compatibility into consideration here. The CRD has been kept at alpha version so that we can fix the expectations now.

api/v1alpha1/networkfence_types.go
Madhu-1: set `kubebuilder:default:Fenced` so that we dont have any issues for the existing CR's.

controllers/networkfence_controller.go
Madhu-1: >Lets assume we have 3 kube clusters A, B, C sharing the same storage cluster. Now, A goes down and it is fenced off via B. Now before we can clean up A lets say B has to be decommissioned. And this is where ideally the fencing resource from B has to be migrated to the surviving cluster C along with other things. Here the fencing resource is deleted as part of cleanup of B. Hence even though A should be in a fenced state, it can get unfenced if the deletion from B happens after the migration of the fencing CR to C. And this will be the order of execution (i.e. migrate to C and then clean up B) to ensure smooth migration causing A to be silently get unfenced.

For this case, it sounds good :+1: Thanks for explaining it.

I think it's taking some other direction. this is all that I meant, I don't like the idea of asking users to create and delete 2 CRs to fence and unfenced the operation. this is what am looking at

>However, we can introduce an option to say "unfence on delete" which upon being true, will lead to the behavior you suggested. i.e. upon deletion of a NetworkFence CR, the reconciler will first unfence the CIDRs from spec and then deletes the resource. This way, the consumer of this resource would have more control over how they can use the resource.

This is the best option for now to avoid more CR creation and deletion for some users. (we can take this in a followup PR)

controllers/networkfence_controller.go
raghavendrabhat: This is a valid concern. Perhaps it can be incremented to v1alpha2. 

There can be 2 scenarios how this PR can have impacts if the version is not incremented.

1) Installed CRD Version is alpha1 (before this PR) and controller code is new (as of this PR)
    Older specs continue to work i.e. people can submit yamls with older specs. But submitting a yaml that has current changes (fence and unfence states) will not be accepted.

   However, since the current controller refers the fence and unfence state from spec, the controller will report errors related to CR.

2) Installed  CRD is new (as of this PR) and controller is older (before this PR)
    In this scenario, due to the recommendation made in the comment [1], both the older specs  of alpha1 and newer specs should be accepted as no fence state means it will be considered as fenced state by default. Also, from the controller perspective, it does not even look at the spec.State, so it would not report errors related to CR. However, if someone explicitly wants to unfence a cluster by mentioning Unfence state in the spec, the controller would still go ahead and fence them which can be a problem. 

Currently not aware of any consumers who are using the current version of this CR or the controller. So, not sure if this PR would have an impact on any existing deployments or if it is worth the effort to make those changes to increment the version to v1alpha2.

Currently submitting the changes with comments mentioned in [1]. 


[1] https://github.com/csi-addons/kubernetes-csi-addons/pull/152#discussion_r866512971

api/v1alpha1/networkfence_types.go
raghavendrabhat: ACK

controllers/networkfence_controller.go
raghavendrabhat: ACK

controllers/networkfence_controller.go
Madhu-1: Yes, we dont need to worry about this one now. as we already have a default and the validation is already done in the kube builder. if someone wants new functionality they need to update the controller. keeping that in mind can we remove this check. except for this one all LGTM.

controllers/networkfence_controller.go
```diff
@@ -63,6 +63,11 @@ func validateNetworkFenceSpec(nwFence *csiaddonsv1alpha1.NetworkFence) error {
 	if nwFence.Spec.Cidrs == nil {
 		return errors.New("required parameter cidrs is not specified")
 	}
+
+	if nwFence.Spec.FenceState != csiaddonsv1alpha1.Fenced && nwFence.Spec.FenceState != csiaddonsv1alpha1.Unfenced {
+		return errors.New("invalid fence state. fence state can only be either of Fenced or Unfenced")
```
Yuggupta27, line 68: Should we log fence state in the error message?

deploy/controller/crds.yaml
Yuggupta27: Updating the version depends on if we expect more major changes to the CRD. If we expect more such changes in the recent future, better to stay with `v1alpha1` for now. What do you think?

Yuggupta27 (commented):
Minor comments. Else LGTM!

controllers/networkfence_controller.go
raghavendrabhat: As suggested in the comment [1], will be removing this check.


[1] https://github.com/csi-addons/kubernetes-csi-addons/pull/152#discussion_r867643651

controllers/networkfence_controller.go
raghavendrabhat: ACK. Will be removing this check from the CR validation function.

controllers/networkfence_controller.go
```diff
@@ -113,37 +114,78 @@ func (r *NetworkFenceReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 		logger.Info("NetworkFence object is terminated, skipping reconciliation")
 		return ctrl.Result{}, nil
 	}
 
-	err = r.fenceClusterNetwork(ctx, nwFence, client, logger)
+	if nwFence.Spec.FenceState == csiaddonsv1alpha1.Fenced {
+		nf.logger.Info("FenceClusterNetwork Request")
+	} else {
+		nf.logger.Info("UnFenceClusterNetwork Request")
+	}
```
Madhu-1, line 143: log namespaced name (CR name and namespace name) which helps

controllers/networkfence_controller.go
```diff
@@ -113,37 +114,78 @@ func (r *NetworkFenceReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+
+	err = nf.processFencing(ctx)
 	if err != nil {
 		logger.Error(err, "failed to fence cluster network")
+		msg := fmt.Sprintf("%v", err.Error())
+		updateStatusErr := nf.updateStatus(ctx, csiaddonsv1alpha1.FencingOperationResultFailed, msg)
```
Madhu-1, line 149: use pass err.Error() directly as it's a string?

Madhu-1 (approved):
nits, LGTM

controllers/networkfence_controller.go
raghavendrabhat: ACK

controllers/networkfence_controller.go
raghavendrabhat: ACK

Madhu-1 (approved):

Yuggupta27 (approved):
LGTM!

nixpanic (approved):
thanks for the corrections and explanations!

Rakshith-R (approved):

Rakshith-R:
@yati1998 , this pr is waiting for your review, ptal

yati1998 (approved):

PR #186, opened 2022-07-20
sidecar: delete pre-existing csiaddonsnode object & recreate

Rakshith-R:
There have been instances of pod ip being changed, while
the pod name remains the same.
This causes csiaddonsnode object to have the wrong endpoint.
With this commit, on each restart delete any csiaddonsnode
object if it pre-exists and create new one.

### DIFF
--- a/sidecar/internal/csiaddonsnode/csiaddonsnode.go
+++ b/sidecar/internal/csiaddonsnode/csiaddonsnode.go
@@ -86,7 +86,7 @@ func (mgr *Manager) Deploy() error {
 	}
 
 	// loop until the CSIAddonsNode has been created
-	wait.PollImmediateInfinite(nodeCreationRetry, func() (bool, error) {
+	_ = wait.PollImmediateInfinite(nodeCreationRetry, func() (bool, error) {
 		err := mgr.newCSIAddonsNode(object)
 		if err != nil {
 			klog.Errorf("failed to create CSIAddonsNode %s/%s: %v",
@@ -96,6 +96,8 @@ func (mgr *Manager) Deploy() error {
 			return false, nil
 		}
 
+		klog.Infof("CSIAddonsNode '%s/%s' successfully created",
+			object.Namespace, object.Name)
 		// no error, so the CSIAddonsNode has been created
 		return true, nil
 	})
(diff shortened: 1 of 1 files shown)


Madhu-1 (commented):
changes Look Good did you get a chance to test it?

Rakshith-R:
> changes Look Good did you get a chance to test it?

Yes, 
I changed the container image
(same image, different tags :sweat_smile: )
This leads to only container restart. 

The csiaddons node was deleted and recreated.

Madhu-1 (approved):

nixpanic (changes requested):
Instead of deleting the CR and re-creating it with a different IP-address, would it not be easier to use the hostname of the Pod, instead of IP-addresses?

Rakshith-R:
> Instead of deleting the CR and re-creating it with a different IP-address, would it not be easier to use the hostname of the Pod, instead of IP-addresses?

I am not sure how we can leverage hostname of a Pod for csi-addons controller to connect to. 
@nixpanic Can you please add any links/references which demonstrate how this can be done ?

nixpanic:
> @nixpanic Can you please add any links/references which demonstrate how this can be done ?

I think you need to set `setHostnameAsFQDN: true` as described in https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-sethostnameasfqdn-field

nixpanic:
Using Pod hostnames instead of their IP-address seems not trivial. I got it to work for some Pods, but not the Ceph-CSI DaemonSets.

Instead of hard-coding IP-addresses in the CSIAddonsNode CR as endpoint, I am considering to add an alternative format for the endpoint. Using a URL format like `pod://csi-rbdplugin-9n9jf:9070` to describe the name of the Pod (to fetch the IP through the k8s API) and the port. `pod://csi-rbdplugin-9n9jf.rook-ceph:9070` could be used to include the namespace as well.

Madhu-1:
> Using Pod hostnames instead of their IP-address seems not trivial. I got it to work for some Pods, but not the Ceph-CSI DaemonSets.
> 
> Instead of hard-coding IP-addresses in the CSIAddonsNode CR as endpoint, I am considering to add an alternative format for the endpoint. Using a URL format like `pod://csi-rbdplugin-9n9jf:9070` to describe the name of the Pod (to fetch the IP through the k8s API) and the port. `pod://csi-rbdplugin-9n9jf.rook-ceph:9070` could be used to include the namespace as well.

You mean without cache, won't this become a costly operation for each request we need to do get call for the API-server to featch the IP addresses.

nixpanic:
> You mean without cache, won't this become a costly operation for each request we need to do get call for the API-server to featch the IP addresses.

Once a connection is established, it will be re-used for subsequent operations. So the IP-lookup is not done constantly. It should be done after a connection failure, as it is possible the IP-address of the Pod changed. Also note that there is a cache in the k8s API client, I assume that works for getting details from a Pod by name just fine.

Rakshith-R:
> > You mean without cache, won't this become a costly operation for each request we need to do get call for the API-server to featch the IP addresses.
> 
> Once a connection is established, it will be re-used for subsequent operations. So the IP-lookup is not done constantly. It should be done after a connection failure, as it is possible the IP-address of the Pod changed. Also note that there is a cache in the k8s API client, I assume that works for getting details from a Pod by name just fine.

This approach sounds good to me.
pod name and namespace is used as csiaddonsnode object name and namespace. 
I guess we could query pod ip from this combination and use it to connect.
And have a mechanism to detect connection failure to redo the pod-ip query if needed.

https://github.com/csi-addons/kubernetes-csi-addons/blob/1e8178f1431e28540e88497fd78d39db88cdd8ce/sidecar/internal/csiaddonsnode/csiaddonsnode.go#L169-L172

Rakshith-R:
Closing in favour of https://github.com/csi-addons/kubernetes-csi-addons/pull/190

PR #190, opened 2022-07-28
Use `pod://` URL formatting for CSIAddonsNode endpoints

nixpanic:
When there is no `controller-ip` parameter set, the endpoint in the
CSIAddonsNode CR will be set to `pod://<name>.<namespace>:<port>`. The
Controller can then get the PodIP through the Kubernetes API. This makes
sure that the CSIAddonsNode does not need to be updated when the
IP-address of the Pod hosting the sidecar changes.

Closes: #186

### DIFF
--- a/controllers/csiaddons/csiaddonsnode_controller.go
+++ b/controllers/csiaddons/csiaddonsnode_controller.go
@@ -20,12 +20,15 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/url"
+	"strings"
 
 	csiaddonsv1alpha1 "github.com/csi-addons/kubernetes-csi-addons/apis/csiaddons/v1alpha1"
 	"github.com/csi-addons/kubernetes-csi-addons/internal/connection"
 	"github.com/csi-addons/kubernetes-csi-addons/internal/util"
 
 	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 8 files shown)


nixpanic:
This is currently still a **Draft** as it is unclear how Pods can change their IP-address but keep their name.

Madhu-1 (commented):
change LGTM. lets test this one and see its working as expected?

nixpanic:
@mergifyio rebase

controllers/csiaddons/csiaddonsnode_controller.go
```diff
@@ -184,6 +196,59 @@ func (r *CSIAddonsNodeReconciler) removeFinalizer(
...
+	namespace, podname, port, err := parseEndpoint(rawURL)
+	if err != nil && errors.Is(err, errLegacyEndpoint) {
+		return rawURL, nil
+	} else if err != nil {
+		return "", err
+	}
+
+	pod := &corev1.Pod{}
+	err = r.Client.Get(ctx, client.ObjectKey{
+		Namespace: namespace,
```
Madhu-1, line 211: parseEndpoint can return empty namespace, if the namespace is `""` empty the `r.Client.Get` will fail. we need to set the namespace to `default` if not set

controllers/csiaddons/csiaddonsnode_controller.go
```diff
@@ -184,6 +196,59 @@ func (r *CSIAddonsNodeReconciler) removeFinalizer(
...
+		Name:      podname,
+	}, pod)
+	if err != nil {
+		return "", fmt.Errorf("failed to get pod %s/%s: %w", namespace, podname, err)
+	}
+
```
Madhu-1, line 217: add a check for `pod.Status.PodIP` here to ensure the pod IP is set?

controllers/csiaddons/csiaddonsnode_controller.go
nixpanic: `default` is just a namespace like any other, I do not think it should be used if none it set. Not sure what is better, fail the `Get()` with an error, or report the error before calling it?

nixpanic: done, added more checks

controllers/csiaddons/csiaddonsnode_controller.go
nixpanic: added a check for this as well

Madhu-1 (approved):

Rakshith-R (approved):

PR #203, opened 2022-08-16
provide option to set MaxConcurrentReconciles

Madhu-1:
Current MaxConcurrentReconciles is 1, which is the default, meaning only one request will be processed at a time. This is not useful when we have 100/1000 CR to reconcile, which can affect the performance. This PR provides a way to set the MaxConcurrentReconciles to the user's desired value so that the user can tune the configuration based on the requirement.

Note:- Currently we have performance issue in DR VolRep because of this one, the failover and failback time are slightly affected because of the default value. with this, things will be better and same change need to be done when we add VolRep Reconciler also.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -29,6 +29,7 @@ import (
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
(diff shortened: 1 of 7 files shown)


Madhu-1:
@nixpanic added doc for the command line arguments PTAL

raghavendra-talur (approved):

cmd/manager/main.go
```diff
@@ -61,12 +62,14 @@ func main() {
...
 	var probeAddr string
 	var reclaimSpaceTimeout time.Duration
+	var maxConcurrentReconciles int
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
 	flag.DurationVar(&reclaimSpaceTimeout, "reclaim-space-timeout", defaultTimeout, "Timeout for reclaimspace operation")
+	flag.IntVar(&maxConcurrentReconciles, "max-concurrent-reconciles", 100, "Maximum number of concurrent reconciles")
```
Rakshith-R, line 72:
```suggestion
	flag.IntVar(&maxConcurrentReconciles, "max-concurrent-reconciles", 100, "Maximum number of concurrent reconciles")
```

Do we really need `100` as default value ? 
most controllers have just a single loop.

Rakshith-R (commented):
The changes seem fine to me, 
Ack, I'll make similar changes to vol-rep controller which will have pr up within next week, 

just left a question regarding the default value.

cmd/manager/main.go
Madhu-1: Yes, having 100 in default doesn't cause any harm, and also, it doesn't make sense to have different configurations per controller. This helps to have concurrent reconciles, which fasten the process when we have more requests.

Rakshith-R (approved):

PR #213, opened 2022-08-23
Ensure only one VR per PVC

Madhu-1:
VolumeReplication should ensure there is only one VR per PVC, as otherwise, it can lead to orchestrating the
PVC is based on multiple VRs to an inconsistent state. With this Patch only one VR will operate on a single PVC.

### DIFF
--- a/controllers/replication.storage/pvc.go
+++ b/controllers/replication.storage/pvc.go
@@ -24,6 +24,8 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
+
+	replicationv1alpha1 "github.com/csi-addons/kubernetes-csi-addons/apis/replication.storage/v1alpha1"
 )
 
 // getPVCDataSource get pvc, pv object from the request.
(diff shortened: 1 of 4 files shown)


controllers/replication.storage/pvc.go
```diff
@@ -56,3 +58,37 @@ func (r VolumeReplicationReconciler) getPVCDataSource(logger logr.Logger, req ty
...
+
+			return fmt.Errorf("failed to update PVC %s annotation for VolumeReplication: %w",
+				pvc.Name, err)
+		}
+
+		return nil
+	}
+
+	if ownerName != req.Name || ownerNamespace != req.Namespace {
+		logger.Info("cannot change the owner of PVC", "PVC name", pvc.Name, "current owner", ownerName, "requested owner", ownerName)
```
Rakshith-R, line 87:
```suggestion
		logger.Info("cannot change the owner of PVC",
		"PVC name", pvc.Name,
		"current owner", fmt.Sprintf("%s/%s",ownerNamesapce,ownerName),
		"requested owner", fmt.Sprintf("%s/%s",req.Namesapce,req.Name))
```

requested owner pointed to ownerName again at the end 

and add namespace here too?

controllers/replication.storage/pvc.go
```diff
@@ -56,3 +58,37 @@ func (r VolumeReplicationReconciler) getPVCDataSource(logger logr.Logger, req ty
...
+
+	if ownerName == "" {
+		pvc.ObjectMeta.Annotations[replicationv1alpha1.VolumeReplicationNameAnnotation] = req.Name
+		pvc.ObjectMeta.Annotations[replicationv1alpha1.VolumeReplicationNamespaceAnnotation] = req.Namespace
+
+		err := r.Update(ctx, pvc)
+		if err != nil {
+			logger.Error(err, "Failed to update PVC annotation", "Name", pvc.Name)
+
+			return fmt.Errorf("failed to update PVC %s annotation for VolumeReplication: %w",
```
Rakshith-R, line 79:
```suggestion
			return fmt.Errorf("failed to update PVC %q annotation for VolumeReplication: %w",
```

Rakshith-R (changes requested):

controllers/replication.storage/pvc.go
Madhu-1: a namespace is not required removed it. A namespace is not required because the VR and PVC should be on the same namespace

Rakshith-R (approved):
LGTM !

controllers/replication.storage/pvc.go
```diff
@@ -56,3 +58,36 @@ func (r VolumeReplicationReconciler) getPVCDataSource(logger logr.Logger, req ty
...
+// annotatePVCWithOwner will add the VolumeReplication details to the PVC annotations.
+func (r *VolumeReplicationReconciler) annotatePVCWithOwner(ctx context.Context, logger logr.Logger, req types.NamespacedName, pvc *corev1.PersistentVolumeClaim) error {
+	if pvc.ObjectMeta.Annotations == nil {
+		pvc.ObjectMeta.Annotations = map[string]string{}
+	}
+
+	ownerName := pvc.ObjectMeta.Annotations[replicationv1alpha1.VolumeReplicationNameAnnotation]
+	if ownerName == "" {
+		pvc.ObjectMeta.Annotations[replicationv1alpha1.VolumeReplicationNameAnnotation] = req.Name
+		err := r.Update(ctx, pvc)
```
ShyamsundarR, line 71: 2 separate VR reconciles could update the annotation at the same time. (2 go routines etc.). This may need some rudimentary locking such that there is only one updater (fetch and update sequence) for a given PVC.

This should be quite rare, but something to keep track of.

Madhu-1: Agree but am thinking if 2 VR acts on the same PVC to add annotation, either one will fail and one will pass and should be fine. what do you think? as its a quite rare condition we can see how to fix it.

PR #214, opened 2022-08-23
set MaxConcurrentReconciles for VolumeReplication

Madhu-1:
Added option to set MaxConcurrentReconciles for VolumeReplication which was pending and for all other controllers this is set in #203

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -145,8 +145,7 @@ func main() {
 		Scheme:   mgr.GetScheme(),
 		Connpool: connPool,
 		Timeout:  defaultTimeout,
-		Log:      ctrl.Log.WithName("controllers").WithName("VolumeReplication"),
-	}).SetupWithManager(mgr); err != nil {
+	}).SetupWithManager(mgr, ctrlOptions); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "VolumeReplication")
 		os.Exit(1)
 	}
--- a/controllers/replication.storage/pvc_test.go
+++ b/controllers/replication.storage/pvc_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package controllers
 
 import (
+	"context"
 	"testing"
 
 	replicationv1alpha1 "github.com/csi-addons/kubernetes-csi-addons/apis/replication.storage/v1alpha1"
@@ -27,7 +28,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
-	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 const (
@@ -102,7 +103,6 @@ func createFakeVolumeReplicationReconciler(t *testing.T, obj ...runtime.Object) 
 	return VolumeReplicationReconciler{
 		Client: client,
 		Scheme: scheme,
-		Log:    logf.Log.WithName("controller_volumereplication_test"),
 	}
 }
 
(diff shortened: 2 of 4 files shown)


Madhu-1:
> Logger and replicationclient is shared in reconciler struct, this change will not be sufficient.
> 
Changed logger code replication client is not a problem it is initialized for each Reconcile.

> We need to implement something similar to networkfenceintance struct

Rakshith-R:
> > Logger and replicationclient is shared in reconciler struct, this change will not be sufficient.
> 
> Changed logger code replication client is not a problem it is initialized for each Reconcile.
> 
> > We need to implement something similar to networkfenceintance struct

But it's set on the struct parameter here 😅
https://github.com/csi-addons/kubernetes-csi-addons/blob/2fc35bbb8a79fc97d2f64e2d9f0ec808c9982769/controllers/replication.storage/volumereplication_controller.go#L360

Madhu-1:
> > > Logger and replicationclient is shared in reconciler struct, this change will not be sufficient.
> > 
> > 
> > Changed logger code replication client is not a problem it is initialized for each Reconcile.
> > > We need to implement something similar to networkfenceintance struct
> 
> But it's set on the struct parameter here sweat_smile
> 
> https://github.com/csi-addons/kubernetes-csi-addons/blob/2fc35bbb8a79fc97d2f64e2d9f0ec808c9982769/controllers/replication.storage/volumereplication_controller.go#L360

it will be set on every reconcile

Rakshith-R:
> > > > Logger and replicationclient is shared in reconciler struct, this change will not be sufficient.
> > > 
> > > 
> > > Changed logger code replication client is not a problem it is initialized for each Reconcile.
> > > > We need to implement something similar to networkfenceintance struct
> > 
> > 
> > But it's set on the struct parameter here sweat_smile
> > https://github.com/csi-addons/kubernetes-csi-addons/blob/2fc35bbb8a79fc97d2f64e2d9f0ec808c9982769/controllers/replication.storage/volumereplication_controller.go#L360
> 
> it will be set on every reconcile

Yes, but now it'll be done parallely which will cause problem.

Madhu-1:
> > > > > Logger and replicationclient is shared in reconciler struct, this change will not be sufficient.
> > > > 
> > > > 
> > > > Changed logger code replication client is not a problem it is initialized for each Reconcile.
> > > > > We need to implement something similar to networkfenceintance struct
> > > 
> > > 
> > > But it's set on the struct parameter here sweat_smile
> > > https://github.com/csi-addons/kubernetes-csi-addons/blob/2fc35bbb8a79fc97d2f64e2d9f0ec808c9982769/controllers/replication.storage/volumereplication_controller.go#L360
> > 
> > 
> > it will be set on every reconcile
> 
> Yes, but now it'll be done parallely which will cause problem.

AFAIK Each Reconcile request will get its own copy of the struct, and Even if we have multiple parallel request this should not an issue.

Rakshith-R:
https://github.com/csi-addons/kubernetes-csi-addons/blob/2fc35bbb8a79fc97d2f64e2d9f0ec808c9982769/cmd/manager/main.go#L143

Negative, the same struct is used.

Madhu-1:
> https://github.com/csi-addons/kubernetes-csi-addons/blob/2fc35bbb8a79fc97d2f64e2d9f0ec808c9982769/cmd/manager/main.go#L143
> 
> Negative, the same struct is used.

Not sure you are understanding what am trying to explain. Even the main struct is initialized here, and the controller framework will send this struct to all to reconcile with its own copy. Have you seen any issues with this change?

Madhu-1:
I tested this one I dont see any problem with ReplicationClient, If you do see please let me know

Rakshith-R (approved):
:+1: , 
Lets take this in, 
I'll send a follow-up pr to add a replicationInstance struct to contain replication client and other required parameters for functions like https://github.com/csi-addons/kubernetes-csi-addons/blob/2fc35bbb8a79fc97d2f64e2d9f0ec808c9982769/controllers/replication.storage/volumereplication_controller.go#L446-L450

yati1998 (approved):

PR #232, opened 2022-09-07
Add LastSyncTime to volume replication status

yati1998:
This PR adds lastsynctime to volumereplication status and updates the reconcile logic to get the LastSynctime


controllers/replication.storage/volumereplication_controller.go
```diff
@@ -125,6 +125,8 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	}
 	// remove the prefix keys in volume replication class parameters
 	parameters := filterPrefixedParameters(replicationParameterPrefix, vrcObj.Spec.Parameters)
+	schedulingTime := parameters["schedulingInterval"]
+	scheduleTime, _ := time.ParseDuration(schedulingTime)
```
nixpanic, line 129: describe the `schedulingInterval` parameter somewhere

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -363,13 +365,30 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+
+	}
 	err = r.updateReplicationStatus(instance, logger, getReplicationState(instance), msg)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
 
 	logger.Info(msg)
 
+	if requeueForInfo {
```
nixpanic, line 384: why add requeue? I guess you want a recurring call to regularly update the state?

yati1998: yes, requeue should do that right? IT will requeue after every interval.

internal/proto/replication.proto
```diff
@@ -174,3 +177,31 @@ message ResyncVolumeResponse{
...
+  // Secrets required by the plugin to complete the request.
+  string secret_name = 3;
+  string secret_namespace = 4;
+}
+// GetVolumeReplicationInfoResponse holds the information to send the
+// volume replication info.
+message GetVolumeReplicationInfoResponse{
+  // Holds the last sync time.
+  // This field is REQUIRED.
+  int64 last_sync_time = 1;
```
Rakshith-R, line 206: https://github.com/csi-addons/spec/pull/44/files#diff-36da0b8129c7cf1c1695bf0ae42ea29185aa75ee81a8a34b7ad1a5d1eedd989f

Use timestamp here too.

controllers/replication.storage/volumereplication_controller.go
Rakshith-R, line 129: This is optional and maybe empty ? 

do we need a default schedule time ? 

cc @Madhu-1

Madhu-1: Yes it can be empty and it's optional

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -363,13 +365,30 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	}
 
 	instance.Status.LastCompletionTime = getCurrentTime()
+	var last_sync_time int64
```
Madhu-1, line 368: Don't use _ in variable names

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -363,13 +365,30 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+	var requeueForInfo bool
+	if instance.Spec.ReplicationState == replicationv1alpha1.Primary {
```
Madhu-1, line 370: Is it required only for primary state?

controllers/replication.storage/volumereplication_controller.go
Madhu-1, line 129: Whhy we are sending parameters?

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -614,6 +633,34 @@ func (r *VolumeReplicationReconciler) enableReplication(vr *volumeReplicationIns
...
+		return 0, err
+	}
+
+	var last_sync_time int64
+
+	last_sync_time = infoResponse.LastSyncTime
+
+	return last_sync_time, nil
+}
+
```
Madhu-1, line 663: Just returned .LastSyncTime no need to have a variable

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -651,3 +698,9 @@ func getCurrentTime() *metav1.Time {
 
 	return &metav1NowTime
 }
+
+func getTime(time time.Time) *metav1.Time {
```
Madhu-1, line 702: No need for a function for this one.

controllers/replication.storage/volumereplication_controller.go
yati1998: We want to update the last sync time after every interval and that is, scheduling interval and hence, it's required.

controllers/replication.storage/volumereplication_controller.go
yati1998: Yes, we need to resync only when it is in primary state.

controllers/replication.storage/volumereplication_controller.go
Madhu-1: from VolumeReplication CR schedulingInterval is optional

yati1998: done

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -125,6 +125,14 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	}
 	// remove the prefix keys in volume replication class parameters
 	parameters := filterPrefixedParameters(replicationParameterPrefix, vrcObj.Spec.Parameters)
+	schedulingTime := parameters["schedulingInterval"]
+	if schedulingTime == "" {
+		schedulingTime = "1m"
```
Rakshith-R, line 130: define it as a constant ? 
and write elaborate how it will be used.

internal/proto/replication.proto
Rakshith-R: This is still not implemented, 
with this change, we can remove some code for conversion in other functions too.

Rakshith-R (changes requested):

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -614,6 +644,30 @@ func (r *VolumeReplicationReconciler) enableReplication(vr *volumeReplicationIns
...
+
+		return 0, resp.Error
+	}
+
+	infoResponse, ok := resp.Response.(*proto.GetVolumeReplicationInfoResponse)
+	if !ok {
+		err := fmt.Errorf("received response of unexpected type")
+		vr.logger.Error(err, "unable to parse response")
+
+		return 0, err
```
Rakshith-R, line 665:
```suggestion
		return nil, resp.Error
	}

	infoResponse, ok := resp.Response.(*proto.GetVolumeReplicationInfoResponse)
	if !ok {
		err := fmt.Errorf("received response of unexpected type")
		vr.logger.Error(err, "unable to parse response")

		return nil, err
```

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -363,13 +372,34 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+	if instance.Spec.ReplicationState == replicationv1alpha1.Primary {
+		lastSyncTime, err := r.getVolumeReplicationInfo(vr)
+		if err != nil {
+			logger.Error(err, "Failed to get volume replication info")
+		}
+		lastSyncTime1 := lastSyncTime.AsTime()
+		metav1Time := metav1.NewTime(lastSyncTime1)
+		instance.Status.LastSyncTime = &metav1Time
+		requeueForInfo = true
+
```
Rakshith-R, line 386: unnecessary newline

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -125,6 +126,14 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 	// remove the prefix keys in volume replication class parameters
 	parameters := filterPrefixedParameters(replicationParameterPrefix, vrcObj.Spec.Parameters)
+	schedulingTime := parameters["schedulingInterval"]
+	if schedulingTime == "" {
+		schedulingTime = "1m"
+	}
+	scheduleTime, err := time.ParseDuration(schedulingTime)
+	if err != nil {
+		logger.Error(err, "failed to parse time: %v", schedulingTime)
+	}
```
Rakshith-R, line 136: Take a look at these lines again.

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -125,6 +126,14 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	}
 	// remove the prefix keys in volume replication class parameters
 	parameters := filterPrefixedParameters(replicationParameterPrefix, vrcObj.Spec.Parameters)
+	schedulingTime := parameters["schedulingInterval"]
+	if schedulingTime == "" {
+		schedulingTime = "1m"
+	}
+	scheduleTime, err := time.ParseDuration(schedulingTime)
```
Rakshith-R, line 133: use rawScheduleTime instead schedulingTime, 
it is confusing.

controllers/replication.storage/volumereplication_controller.go
Rakshith-R: This is still not done.

Rakshith-R (changes requested):

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -363,13 +372,34 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	}
 
 	instance.Status.LastCompletionTime = getCurrentTime()
+
+	var requeueForInfo bool
+	if instance.Spec.ReplicationState == replicationv1alpha1.Primary {
+		lastSyncTime, err := r.getVolumeReplicationInfo(vr)
```
nixpanic, line 378: maybe call this `info` instead of `lastSyncTime`

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -614,6 +644,30 @@ func (r *VolumeReplicationReconciler) enableReplication(vr *volumeReplicationIns
 	return nil
 }
 
+// getVolumeReplicationInfo gets volume replication info.
+func (r *VolumeReplicationReconciler) getVolumeReplicationInfo(vr *volumeReplicationInstance) (*timestamppb.Timestamp, error) {
```
nixpanic, line 648: this does not return an `info` kind of object? return something like what `volumeReplication.GetInfo()` does, or rename the function to return a LastSyncTimestamp?

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -363,13 +372,34 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 
 	instance.Status.LastCompletionTime = getCurrentTime()
+
+	var requeueForInfo bool
+	if instance.Spec.ReplicationState == replicationv1alpha1.Primary {
+		lastSyncTime, err := r.getVolumeReplicationInfo(vr)
+		if err != nil {
+			logger.Error(err, "Failed to get volume replication info")
+		}
+		lastSyncTime1 := lastSyncTime.AsTime()
```
nixpanic, line 382: if this can not fail, just place it in `metav1Time := metav1.NewTime(lastSyncTime.AsTime())`.

`metav1Time` sounds more like a package name, similar to `metav1`. This is just a variable, so why not call it `time`?

controllers/replication.storage/volumereplication_controller.go
ShyamsundarR: If the schedule is not set for this VR, then the pool level schedule (if available) should be used for the reconcile requeue. If that is not available then we can either NOT requeue or requeue for a duration till the last sync time is not empty (as post that there would be no further syncs).

There is a possibility that a pool level schedule is added later, once we make the above not to requeue decision, but that would either require a restart of the plugin (so that all VRs are reconciled), or reconciling based on added pool level schedules, or waiting for the periodic reconcile (24h) at which time this will be picked up.

We possibly want to track the above for the future, for now if there is no schedule we can default to a something (1m/5m), as the current DR stack comes with a schedule and should not break the functionality.

controllers/replication.storage/volumereplication_controller.go
ShyamsundarR: The primary end has the status, that it reads from the secondary, and so we use that to update the lastSyncTime.

In cases where we demoted a Primary and are waiting for both Secondaries to be in sync, the timestamp is still useful. Tells us how far back the sync is, before either image can be gracefully promoted.

The demoted snapshot sync time is not critical at present, it would also just show that the lastSyncTime is in the past as compared to the demoted snapshot, and ideally once the demoted snapshot is synced the timestamps would be the same. So not sure if it is useful as such.

Needs more thought on how to use that information, hence sticking to when image is primary at one end is useful.

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -363,13 +372,34 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	}
 
 	instance.Status.LastCompletionTime = getCurrentTime()
+
+	var requeueForInfo bool
+	if instance.Spec.ReplicationState == replicationv1alpha1.Primary {
```
ShyamsundarR, line 377: @Madhu-1 If an image is in the split-brain state, is there a TS that is usable?

Madhu-1: AFAIK no during split-brain no sync will happen this is not useful.

controllers/replication.storage/volumereplication_controller.go
ShyamsundarR, line 378: Just a heads up!

The last sync time can be "0" or empty, i.e it was never synced even once, we need to determine how that value would be reflected. One option is to set it to the lower bound of UTC epoch (I am a bit confused on what [Timestamp](https://developers.google.com/protocol-buffers/docs/reference/csharp/class/google/protobuf/well-known-types/timestamp) holds, UTC epoch (since 1970), or since `0001-01-01T00:00:00Z`)

controllers/replication.storage/volumereplication_controller.go
yati1998: yeah for now it just returns last sync time, but this name is given just with the thought of future implementation, we may like to return other informations along with the last sync time, hence don't want it to be specific.

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -363,13 +364,42 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+	var requeueForInfo bool
+	rawScheduleTime := parameters["schedulingInterval"]
+	if rawScheduleTime == "" {
+		rawScheduleTime = "1m"
+	}
+	scheduleTime, err := time.ParseDuration(rawScheduleTime)
+	if err != nil {
+		logger.Error(err, "failed to parse time: %v", rawScheduleTime)
+		return ctrl.Result{}, err
+	}
```
Rakshith-R, line 377:
```suggestion
       const defaultScheduleTime = time.Minute

        func getScheduleTime(parameters map[string]string) (time.Duration) {
		rawScheduleTime := parameters["schedulingInterval"]
		if rawScheduleTime == "" {
			return defaultScheduleTime
		}
		scheduleTime, err := time.ParseDuration(rawScheduleTime)
		if err != nil {
			logger.Error(err, "failed to parse time: %v", rawScheduleTime)
			return defaultScheduleTime
		}
		return scheduleTime
	}
```

Create a constant and a function like above and use it . 

@yati1998 ^

Rakshith-R (changes requested):

controllers/replication.storage/volumereplication_controller.go
nixpanic: In split-brain, I do not assume one of the images is marked **Primary**? If that is the case, the lastSyncTime should be the time of the last successful sync, maybe.

controllers/replication.storage/volumereplication_controller.go
nixpanic: The current state is confusing, so you will have to decide what this function does. It can be renamed once it gets extended, no need to prepare function names for things in the future. Leave a comment with the possibility if you think it is useful.

controllers/replication.storage/volumereplication_controller.go
yati1998: since we are updating it in every interval, I guess if we reach split brain case, it would already have that, cc @ShyamsundarR

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -363,13 +364,47 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	}
 
 	instance.Status.LastCompletionTime = getCurrentTime()
+
+	var requeueForInfo bool
+
+	const defaultScheduleTime = time.Minute
```
Rakshith-R, line 370: Let it const for this package, not locally for the function.

Rakshith-R (changes requested):
I still prefer a separate helper function to keep it clean. 
A minor nit otherwise.

ShyamsundarR (commented):
Broadly LGTM! The logical intention seems to be covered and correct.

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -363,16 +365,51 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 	err = r.updateReplicationStatus(instance, logger, getReplicationState(instance), msg)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
 
 	logger.Info(msg)
 
+	scheduleTime := getScheduleTime(parameters, logger)
+	if requeueForInfo {
+		logger.Info("volume is primary, requeuing to get volume replication info")
```
Madhu-1, line 390: logging is not much helpful. Remove it on add the next queue interval time. and also move line 388 inside the if condition

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -614,6 +651,30 @@ func (r *VolumeReplicationReconciler) enableReplication(vr *volumeReplicationIns
 	return nil
 }
 
+// getVolumeReplicationInfo gets volume replication info.
+func (r *VolumeReplicationReconciler) getLastSyncTime(vr *volumeReplicationInstance) (*timestamppb.Timestamp, error) {
```
Madhu-1, line 655: IMO we should not be doing this. Currently, we have only one parameter in the `GetInfo` this looks fine, If we add more parameters we need to write a new function for it. My suggestion is to return a complete response `infoResponse` and use it in caller

internal/client/replication-client.go
```diff
@@ -45,6 +45,8 @@ type VolumeReplication interface {
 	// ResyncVolume RPC call to resync the volume.
 	ResyncVolume(volumeID, replicationID string, force bool, secretName, secretNamespace string, parameters map[string]string) (*proto.
 		ResyncVolumeResponse, error)
+	// GetVolumeReplicationInfo RPC call to get volume replication info.
+	GetVolumeReplicationInfo(volumeID, replicationID string, secretName, secretNamespace string) (*proto.GetVolumeReplicationInfoResponse, error)
```
Madhu-1, line 49:
```suggestion
GetVolumeReplicationInfo(volumeID, replicationID , secretName, secretNamespace string) (*proto.GetVolumeReplicationInfoResponse, error)```

File: internal/client/replication-client.go [hunk 22]
```diff
@@ -143,3 +145,20 @@ func (rc *replicationClient) ResyncVolume(volumeID, replicationID string, force
 
 	return resp, err
 }
+
+// GetVolumeReplicationInfo RPC call to get volume replication info.
+func (rc *replicationClient) GetVolumeReplicationInfo(volumeID, replicationID string,
+	secretName, secretNamespace string) (*proto.GetVolumeReplicationInfoResponse, error) {
```
Madhu-1, line 151:
```suggestion
func (rc *replicationClient) GetVolumeReplicationInfo(volumeID, replicationID,
	secretName, secretNamespace string) (*proto.GetVolumeReplicationInfoResponse, error) {
```

internal/sidecar/service/volumereplication.go
```diff
@@ -191,3 +191,36 @@ func (rs *ReplicationServer) ResyncVolume(
...
+			Secrets:       data,
+			ReplicationId: req.ReplicationId,
+		})
+	if err != nil {
+		klog.Errorf("Failed to get volume replication info: %v", err)
+		return nil, err
+	}
+
+	lastsynctime := resp.GetLastSyncTime()
+	if lastsynctime == nil {
```
Madhu-1, line 219: when this can be nil?

yati1998: in case there is no local snapshot timestamp and the csi driver return nil.

controllers/replication.storage/volumereplication_controller.go
ShyamsundarR: The question really is what does RBD report, and then what should the proto report. In the case of split-brain we should report not-synced time (which IMHO is the "0" timestamp), or if available the last time it was synced...

nixpanic: I agree, on a sync failure (like split-brain), no LastSyncTime should be reported, but ideally an error. Either the automatic rescheduling should stop, or be done with an exponential backoff.

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -363,16 +364,55 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+
+	if instance.Spec.ReplicationState == replicationv1alpha1.Primary {
+		info, err := r.getVolumeReplicationInfo(vr)
+		if err != nil {
+			logger.Error(err, "Failed to get volume replication info")
+			return ctrl.Result{}, err
+		}
+		lastSyncTime := info.GetLastSyncTime()
+
+		lastsynctime := metav1.NewTime(lastSyncTime.AsTime())
```
nixpanic, line 378: very confusing naming, there is `lastSyncTime` and `lastsynctime`. If you really need both, make sure to call them  differently (maybe `ts` or `timestamp` and `lastSyncTime` or just `t`).

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -363,16 +365,51 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+		return ctrl.Result{
+			Requeue:      true,
+			RequeueAfter: scheduleTime,
+		}, nil
+	}
+
 	return ctrl.Result{}, nil
 }
 
+func getScheduleTime(parameters map[string]string, logger logr.Logger) time.Duration {
```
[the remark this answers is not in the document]
nixpanic: this can indeed be easily unit-tested, make sure to include the following cases:
- an empty map (no items)
- a map with a `schedulingInterval` item that is `nil`
- a map with a `schedulingInterval` item that is `""`
- a map with a `schedulingInterval` item that is can not be parsed by `time.ParseDuration()`
- a map with a `schedulingInterval` item that is valid

nixpanic (changes requested):

controllers/replication.storage/volumereplication_controller.go
ShyamsundarR: When the image is in `replaying` state with RBD we get the additional description, if stopped or split-brain the additional data is not available, and we will hence default to the "0" time stamp. We should be fine with that.

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -363,16 +364,55 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	}
 
 	instance.Status.LastCompletionTime = getCurrentTime()
+
+	var requeueForInfo bool
+
+	if instance.Spec.ReplicationState == replicationv1alpha1.Primary {
```
ShyamsundarR, line 370: (nit) When the state is secondary, we could drop the status update and mark time as a nil pointer, so that it does not carry a stale value.

yati1998:
@nixpanic @ShyamsundarR @Rakshith-R @Madhu-1 done with all the changes here as well.

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -363,16 +364,64 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+		instance.Status.LastSyncTime = &lastSyncTime
+		requeueForInfo = true
+	}
 	err = r.updateReplicationStatus(instance, logger, getReplicationState(instance), msg)
 	if err != nil {
 		return ctrl.Result{}, err
 	}
+	if instance.Spec.ReplicationState == replicationv1alpha1.Secondary {
+		instance.Status.LastSyncTime = nil
+	}
```
Rakshith-R, line 388: This should go at line 382, 
otherwise there will be no update to status.

Rakshith-R (changes requested):

controllers/replication.storage/volumereplication_controller.go
nixpanic: this seems to be done

controllers/replication.storage/volumereplication_controller.go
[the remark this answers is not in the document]
nixpanic: this seems to be covered too

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -363,16 +365,51 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+			Requeue:      true,
+			RequeueAfter: scheduleTime,
+		}, nil
+	}
+
 	return ctrl.Result{}, nil
 }
 
+func getScheduleTime(parameters map[string]string, logger logr.Logger) time.Duration {
+	rawScheduleTime := parameters["schedulingInterval"]
```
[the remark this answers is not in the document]
nixpanic: this is still missing, needs to be documented in https://github.com/csi-addons/kubernetes-csi-addons/blob/main/docs/volumereplication.md

controllers/replication.storage/volumereplication_controller.go
nixpanic: the code has been cleaned up and this is addressed now

nixpanic (changes requested):
Just the documentation for the new parameter needs to be added. Otherwise it looks good to me.

Madhu-1:
If the RPC is not implemented, this will be a breaking change. we need to address it later

controllers/replication.storage/volumereplication_controller.go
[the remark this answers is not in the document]
nixpanic: added in https://github.com/csi-addons/kubernetes-csi-addons/pull/232/files#diff-79b4f8b10848d2028958d32f7e3915a8bd43111aa8aca0d3284d1494df0d53ad

nixpanic (approved):

Rakshith-R (approved):
@yati1998 ,can you open a issue to track handling NotImplemented Error ? 
We can set lastSyncTime to nil if getvolrepInfo returns this error.
We'll need to fix this before another release of csi-addons operator.

PR #257, opened 2022-10-20
replication: add webhook for volumereplicationclass

Madhu-1:
Add webhook for VolumeReplicationClass to validate update CR.


.github/workflows/codespell.yml
```diff
@@ -17,5 +17,9 @@ jobs:
         uses: codespell-project/actions-codespell@master
         with:
           skip: .git,go.mod,go.sum,vendor
+          # kubebuilder latest release has a label with name kuberentes
+          # TODO: remove this in next kubebuilder update as this is fixed
+          # in kubebuilder master
+          ignore_words_list: kuberentes
```
nixpanic, line 23: please add a link to the kubebuilder issue in this comment, that makes it easier to verify what next release contains the fix

config/certmanager/certificate.yaml
```diff
@@ -0,0 +1,39 @@
+# The following manifests contain a self-signed issuer CR and a certificate CR.
+# More document can be found at https://docs.cert-manager.io
+# WARNING: Targets CertManager v1.0. Check https://cert-manager.io/docs/installation/upgrading/ for breaking changes.
+apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  labels:
+    app.kuberentes.io/name: issuer
```
nixpanic, line 8: please add a comment here that it is spelled correctly, and point to the codespell config

config/webhook/service.yaml
```diff
@@ -0,0 +1,20 @@
+
+apiVersion: v1
```
nixpanic, line 2: starting with an empty line? Maybe include the yaml-start-of-document `---` instead (and in other yaml files)

PROJECT
```diff
@@ -57,4 +57,7 @@ resources:
   kind: VolumeReplicationClass
```
nixpanic, line 57: did you use a `kubebuilder` command to generate the initial files? Please include the command(s) in the commit message so that others can reproduce it, or follow the steps to add other webhooks.

apis/replication.storage/v1alpha1/volumereplicationclass_webhook.go
```diff
@@ -0,0 +1,76 @@
...
+	"reflect"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+)
+
+// log is for logging in this package.
+var volumereplicationclasslog = logf.Log.WithName("volumereplicationclass-resource")
```
nixpanic, line 30: `vrcLog` would do too. But it isn't used much, so the long name might not be too inconvenient.

apis/replication.storage/v1alpha1/volumereplicationclass_webhook.go
```diff
@@ -0,0 +1,76 @@
...
+
+func (r *VolumeReplicationClass) SetupWebhookWithManager(mgr ctrl.Manager) error {
```
nixpanic, line 32: Why `r` for `VolumeReplicationClass`? I guess it's used elsewhere and can not easily be modified without changing other files. `r` is not a self-explaining name when using it inside functions.

apis/replication.storage/v1alpha1/volumereplicationclass_webhook.go
```diff
@@ -0,0 +1,76 @@
...
+
+	if r.Spec.Provisioner == "" {
+		return errors.New("provisioner name not provided")
+	}
+
+	return nil
+}
+
+// ValidateUpdate implements webhook.Validator. Validates the updated
+// VolumeReplicationClass and does not allow changes to existing spec object.
```
nixpanic, line 55: If it does not allow any changes to the `.Spec` part, why not use `reflect.DeepEqual` for that? No need to compare `.Provisioner` and `.Parameters` separately.

apis/replication.storage/v1alpha1/webhook_suite_test.go
```diff
@@ -0,0 +1,135 @@
...
+		conn, err := tls.DialWithDialer(dialer, "tcp", addrPort, &tls.Config{InsecureSkipVerify: true})
+		if err != nil {
+			return err
+		}
+		conn.Close()
+		return nil
+	}).Should(Succeed())
+
+}, 60)
+
```
nixpanic, line 129: Where are the tests? I expected to find a `By()` or similar. At least one positive and one negative test would be nice.

nixpanic (changes requested):

.github/workflows/codespell.yml
Madhu-1: there was no issue created for it, it was fixed in a PR https://github.com/kubernetes-sigs/kubebuilder/pull/2964 here

config/certmanager/certificate.yaml
Madhu-1: this is auto generated file adding commit might get overriden when its rebuild

config/webhook/service.yaml
Madhu-1: again this is auto generated file

apis/replication.storage/v1alpha1/volumereplicationclass_webhook.go
Madhu-1: this is also auto generated not manually added, i think it need changes in many places

apis/replication.storage/v1alpha1/volumereplicationclass_webhook.go
Madhu-1: not comparing spec because for now we have only two fields in spec but later we might add new fields which are suppose to be editable also that the reason to compare it explicitly

apis/replication.storage/v1alpha1/webhook_suite_test.go
Madhu-1: Again this is an auto-generated file, current we dont have any functional suite tests. We have already 2 suite files in the repo. https://github.com/csi-addons/kubernetes-csi-addons/blob/main/controllers/csiaddons/suite_test.go https://github.com/csi-addons/kubernetes-csi-addons/blob/main/controllers/replication.storage/suite_test.go. i need to see how to have a single suite to run with the mock controller. I will take it up in the next PR to see how to make it possible.

Madhu-1:
I have tested both yaml-based and bundle-based deployments both work as expected.

Rakshith-R (approved):
LGTM, 
Thanks!

apis/replication.storage/v1alpha1/volumereplicationclass_webhook.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+//+kubebuilder:webhook:path=/validate-replication-storage-openshift-io-v1alpha1-volumereplicationclass,mutating=false,failurePolicy=fail,sideEffects=None,groups=replication.storage.openshift.io,resources=volumereplicationclasses,verbs=update,versions=v1alpha1,name=vvolumereplicationclass.kb.io,admissionReviewVersions=v1
+
+var _ webhook.Validator = &VolumeReplicationClass{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (v *VolumeReplicationClass) ValidateCreate() error {
+	return nil
+}
```
Rakshith-R, line 48: The pr description mentions webhooks for create too 
but here it is empty.

Madhu-1: update the PR description, Create is not required as it should/can be handled as part of CRD schema validation.

Rakshith-R: ok thanks !

nixpanic (approved):
I don't know much about WebHooks, but this looks reasonable to me.

PR #260, opened 2022-10-28
replication: update LastSyncTime if its not nill

Madhu-1:
LastSyncTime can be optional and nil also, there is no strict check for it, and if we dont have this check the [default UNIX time](https://go.dev/play/p/81CmSxr-r8S) will get added to the CR, which doesn't make sense. If the time is not present keep the last known LastSyncTime itself.

Depends-On: csi-addons/spec#47

### DIFF
--- a/controllers/replication.storage/volumereplication_controller.go
+++ b/controllers/replication.storage/volumereplication_controller.go
@@ -374,9 +374,10 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 			return ctrl.Result{}, err
 		}
 		ts := info.GetLastSyncTime()
-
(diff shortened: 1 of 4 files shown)


ShyamsundarR (commented):
This works to not update last known lastSyncTime in cases where the call does not return a value, or hold it at epoch till it does.

NOTE: As I have not cross-checked other areas of code not approving as such.

yati1998 (approved):

Madhu-1:
@nixpanic PTAL

ShyamsundarR (commented):
LGTM!

nixpanic (approved):

Madhu-1:
@mergifyio rebase

yati1998 (approved):

PR #263, opened 2022-11-07
replication: reduce the reqeue time for GetReplicationInfo

Madhu-1:
Reduce the schedule time by half to get the latest update and also to avoid the inconsistency between the last sync time in
the VR and the Storage system.
 
The user can see updates for RPO that are not stuck in a bad schedule race, i.e, VolumeReplication checks and finds sync time as t-5m, and just after that storage system updates it to t+x. If we checked every 1/2 of the schedule we will update it to t+x in t+s/2

### DIFF
--- a/controllers/replication.storage/volumereplication_controller.go
+++ b/controllers/replication.storage/volumereplication_controller.go
@@ -392,20 +392,22 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	logger.Info(msg)
 
 	if requeueForInfo {
-		scheduleTime := getScheduleTime(parameters, logger)
+		reconcileInternal := getInfoReconcileInterval(parameters, logger)
 		return ctrl.Result{
 			Requeue:      true,
-			RequeueAfter: scheduleTime,
+			RequeueAfter: reconcileInternal,
 		}, nil
 	}
 
 	return ctrl.Result{}, nil
 }
 
-// getScheduleTime takes parameters and returns the scheduling interval
-// after converting it to time.Duration. If the schedulingInterval is empty
-// or there is error parsing, it is set to the default value.
-func getScheduleTime(parameters map[string]string, logger logr.Logger) time.Duration {
+// getInfoReconcileInterval takes parameters and returns the half of the scheduling
+// interval after converting it to time.Duration. The interval is converted
+// into half to keep the LastSyncTime and Storage LastSyncTime to be in sync.
+// If the schedulingInterval is empty or there is error parsing, it is set to
+// the default value.
+func getInfoReconcileInterval(parameters map[string]string, logger logr.Logger) time.Duration {
 	// the schedulingInterval looks like below, which is the part of volumereplicationclass
 	// and is an optional parameter.
 	// ```parameters:
@@ -421,7 +423,17 @@ func getScheduleTime(parameters map[string]string, logger logr.Logger) time.Dura
 		logger.Error(err, "failed to parse time: %v", rawScheduleTime)
 		return defaultScheduleTime
 	}
-	return scheduleTime
+	// Return schedule internal to avoid frequent reconcile if the
+	// schedulingInterval is less than 2 minutes.
+	if scheduleTime < 2*time.Minute {
+		logger.Info("scheduling interval is less than 2 minutes, not decreasing it to half")
+		return scheduleTime
+	}
+
+	// Reduce the schedule time by half to get the latest update and also to
+	// avoid the inconsistency between the last sync time in the VR and the
+	// Storage system.
+	return scheduleTime / 2
 }
 
 func (r *VolumeReplicationReconciler) getReplicationClient(driverName string) (grpcClient.VolumeReplication, error) {
--- a/docs/volumereplicationclass.md
+++ b/docs/volumereplicationclass.md
@@ -10,8 +10,8 @@
 
 + `replication.storage.openshift.io/replication-secret-name`
 + `replication.storage.openshift.io/replication-secret-namespace`
-  
-``` yaml  
+
+``` yaml
 apiVersion: replication.storage.openshift.io/v1alpha1
 kind: VolumeReplicationClass
 metadata:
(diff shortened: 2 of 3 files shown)


controllers/replication.storage/volumereplication_controller.go
```diff
@@ -421,7 +421,10 @@ func getScheduleTime(parameters map[string]string, logger logr.Logger) time.Dura
 		logger.Error(err, "failed to parse time: %v", rawScheduleTime)
 		return defaultScheduleTime
 	}
-	return scheduleTime
+	// Reduce the schedule time by half to get the latest update and also to
+	// avoid the inconsistency between the last sync time in the VR and the
+	// Storage system.
+	return scheduleTime / 2
```
Rakshith-R, line 427: Shouldn't this be done by the entity creating VRC in the first place?

I don't think it is logical to to add this logic at this level.

Rakshith-R (changes requested):

controllers/replication.storage/volumereplication_controller.go
Madhu-1: Please explain why you dont think this logic should not be added here. Please check the commit message for the proper explanation of why this is done.

Rakshith-R: >Reduce the schedule time by half to get the latest update and also to avoid the inconsistency between the last sync time in
the VR and the Storage system.

SchedulingTime from the VRC implies the time interval between which we will schedule a re-queue to check for updates right ? 
Can the reducing the time by half done at a above level ? 

>The user can see updates for RPO that are not stuck in a bad schedule race, i.e, VolumeReplication checks and finds sync time as t-5m, and just after that storage system updates it to t+x. If we checked every 1/2 of the schedule we will update it to t+x in t+s/2

By checking 1/2 of the schedule, we are not obeying the provided scheduling time at all? 
Is there a reason that the entity handling this cannot create VRC with half of the required shedule?

Madhu-1: >SchedulingTime from the VRC implies the time interval between which we will schedule a re-queue to check for updates right ?
Can the reducing the time by half done at a above level ?

This is not true, This is the time specific to the storage, what is the interval to take the snapshot of the rbd image(for example ceph). we are here using the same time to Requeue to check the LastSyncTime/Update time of the replication.

>By checking 1/2 of the schedule, we are not obeying the provided scheduling time at all?
Is there a reason that the entity handling this cannot create VRC with half of the required schedule?

Please Check the above explanation, the schedule interval is not meant for GetReplicationInfo Requeue time it's for the Storage Replication interval time, we are using the same time to get the updates on the replication time. if we keep both the same, there is a kind of diff between the VR LastSyncTime and the storage LastSyncTime. to keep both the same, we are decreasing the time by half so that we can keep the LastSyncTime and the storageLastSyncTime to almost the same.

Rakshith-R: okay, 

https://github.com/csi-addons/kubernetes-csi-addons/blob/main/docs/volumereplicationclass.md
Can we add description here and how it used in parameters section passed to csi-driver
And for requeue time of reconcile
and mention it in function description too .

Madhu-1: done, ptal

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -402,9 +402,11 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 }
 
-// getScheduleTime takes parameters and returns the scheduling interval
-// after converting it to time.Duration. If the schedulingInterval is empty
-// or there is error parsing, it is set to the default value.
+// getScheduleTime takes parameters and returns the half of the scheduling
+// interval after converting it to time.Duration. The interval is converted
+// into half to keep the LastSyncTime and Storage LastSyncTime to be in sync.
+// If the schedulingInterval is empty or there is error parsing, it is set to
+// the default value.
```
ShyamsundarR, line 409: (partial nit) There is a bit of confusion with the function naming and the current return value. I was a little surprised that `getScheduleTime` returns 1/2 the schedule time, and checked the code to see that it was only called to set the reconcile interval and not to get the schedule time in general.

So function reuse anywhere else could be inappropriate. Can we rename this function to `reconcileInterval`, or `infoReconcileInterval` to disambiguate?

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -421,7 +423,10 @@ func getScheduleTime(parameters map[string]string, logger logr.Logger) time.Dura
 		logger.Error(err, "failed to parse time: %v", rawScheduleTime)
 		return defaultScheduleTime
 	}
-	return scheduleTime
+	// Reduce the schedule time by half to get the latest update and also to
+	// avoid the inconsistency between the last sync time in the VR and the
+	// Storage system.
+	return scheduleTime / 2
```
ShyamsundarR, line 429: (nit) Protect with a lower bound of 1 minute? IOW if schedule is less than 2 minutes, then let's not reconcile that fast? Typically I expect the replication schedule to be around 5 minutes at the lower bound.

controllers/replication.storage/volumereplication_controller.go
Madhu-1: yes correct, i should have changed the function name.

ShyamsundarR (approved):

docs/volumereplicationclass.md
```diff
@@ -21,5 +21,8 @@ spec:
   parameters:
     replication.storage.openshift.io/replication-secret-name: secret-name
     replication.storage.openshift.io/replication-secret-namespace: secret-namespace
+    # schedulingInterval is a vendor specific parameter. It is used to set the
+    # replication scheduling interval for storage volumes that are replication
+    # enabled using related VolumeReplication resource
```
Rakshith-R, line 26: Add a line about re queueing of VR if specified too?

Rakshith-R (approved):

PR #274, opened 2022-11-21
Add Webhooks for CR's

Madhu-1:
This adds the pending webhooks for the below CR's
* ReclaimSpaceJob CR
* ReclaimSpaceCronJob CR
* NetworkFence CR
* csiaddosNode CR

Note:- If any extra fields need to be validated, we can open a separate issue and track it as an enhancement, as this PR adds the basic framework, which will allow us for further enhancements.


yati1998 (approved):

apis/csiaddons/v1alpha1/networkfence_webhook.go
```diff
@@ -0,0 +1,88 @@
...
+		allErrs = append(allErrs, field.Invalid(field.NewPath("spec").Child("parameters"), n.Spec.Parameters, "parameters cannot be changed"))
+	}
+
+	if n.Spec.Secret.Name != oldNetworkFence.Spec.Secret.Name {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "secret", "name"), n.Spec.Secret, "secret name cannot be changed"))
+	}
+
+	if n.Spec.Secret.Namespace != oldNetworkFence.Spec.Secret.Namespace {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "secret", "namespace"), n.Spec.Secret, "secret namespace cannot be changed"))
+	}
```
Rakshith-R, line 74: just wondering, is there any reason to have these fields as immutable ?

rest of the pr looks good to me.

Madhu-1: I don't see a reason to allow to change the secret once created.  Do you see any use case?

Rakshith-R: > I don't see a reason to allow to change the secret once created. Do you see any use case?

no, I can't think of a reason for either side of the argument for secret name/namespace. I was wondering if you had any specific reasons to make them immutable.
But parameters also need to remain unchanged ? 

@raghavendra-talur Can you please take a look networkfence CR webhook once?

Madhu-1: > no, I can't think of a reason for either side of the argument for secret name/namespace. I was wondering if you had any specific reasons to make them immutable.
> But parameters also need to remain unchanged ?

Am considering this should not be changed once created. if they made any mistake with secrets we should not allow it, the user is free to delete and create it again with new secrets. example secrets and parameters cannot be changed once created in PV and storageclass.

Rakshith-R (approved):
Thanks!

PR #285, opened 2023-01-02
reclaimspace: consider VA object only it is attached & not being deleted

Rakshith-R:
This commit makes modification to ignore Volume Attachment objects which are being deleted or not attached.

### DIFF
--- a/controllers/csiaddons/reclaimspacejob_controller.go
+++ b/controllers/csiaddons/reclaimspacejob_controller.go
@@ -324,7 +324,7 @@ func (r *ReclaimSpaceJobReconciler) getTargetDetails(
 		pvName:     pv.Name,
 	}
 	for _, v := range volumeAttachments.Items {
-		if *v.Spec.Source.PersistentVolumeName == pv.Name {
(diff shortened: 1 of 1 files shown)


controllers/csiaddons/reclaimspacejob_controller.go
```diff
@@ -324,7 +324,7 @@ func (r *ReclaimSpaceJobReconciler) getTargetDetails(
 		pvName:     pv.Name,
 	}
 	for _, v := range volumeAttachments.Items {
-		if *v.Spec.Source.PersistentVolumeName == pv.Name {
+		if v.DeletionTimestamp.IsZero() && v.Status.Attached && *v.Spec.Source.PersistentVolumeName == pv.Name {
```
Madhu-1, line 327: This is not true for controller space reclaim, as PV is not required to be attached, right?

Rakshith-R: > This is not true for controller space reclaim, as PV is not required to be attached, right?

We don't need node information for performing controller space reclaim.
It will still happen.

Madhu-1 (approved):

yati1998 (approved):

PR #301, opened 2023-02-08
Add Support to volumeGroup as a source

matancarmeli7:
Add Support to `volumeGroup` as a source, by adding `replicationSource` object which is replacing `volumeID` and can handle `volumeID` or `VolumeGroupID` as a source.


matancarmeli7:
opened new PR #302

PR #302, opened 2023-02-09
Support volumeGroup as source

matancarmeli7:
Add Support to `volumeGroup` as a source, by adding `replicationSource` object which is replacing `volumeID` and can handle `volumeID` or `VolumeGroupID` as a source.

Signed-off-by: matancarmeli7 <matan.carmeli7@gmail.com>


nixpanic:
This PR adds a dependency on functionalities from the IBM Volume Group Controller. I do not think that is what we want in this project. We need to be storage provider agnostic, and only consume the protocol from the [CSI-Addons Spec](https://github.com/csi-addons/spec) and use the standard Kubernetes API. If there is a dependency on a storage provider, other storage providers will have difficulties in implementing the required functionality.

Would it be possible to use the upcoming VolumeGroupSnapshot [KEP-3476](https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/3476-volume-group-snapshot) for this? Or maybe something that works in a similar way, which could be added to the CSI-Addons Spec?

Ideally the features that are available in CSI-Addons will get proposed for the [CSI Spec](https://github.com/container-storage-interface/spec/blob/master/spec.md) in the future.

erantzabari (approved):

nixpanic:
@mergifyio rebase

go.mod
```diff
@@ -17,7 +18,7 @@ require (
 	google.golang.org/grpc/cmd/protoc-gen-go-grpc v1.3.0
 	google.golang.org/protobuf v1.30.0
 	k8s.io/api v0.26.3
-	k8s.io/apimachinery v0.26.3
+	k8s.io/apimachinery v0.27.1
```
nixpanic, line 21: operator-sdk does not like this update. Can you prevent this change?

matancarmeli7: yes I saw, we need to downgrade it first in vg repo, I will do it and will update here

matancarmeli7: done

nixpanic:
VolumeGroup and VolumeGroupContent are not types we like to expose to users. VolumeGroupReplication uses a label-selector instead, which then resolves into a list for VolumeGroupReplicationContent.

PR #304, opened 2023-02-15
add default value to autoResync

matancarmeli7:
Currently, if a user will want to create `VolumeReplication` without the `autoResync` field it will fail because it’s a required field. After this PR it will have a default in the CRD and it won’t fail and will create the object with the default value


Madhu-1 (approved):

ShyamsundarR:
LGTM, from a ramen usage perspective, we always set the value, hence the default will not be picked up (and also the default as false is correct even in case it is not specified)

yati1998 (approved):

erantzabari:
Shyam,
From replication (without Ramen) we do need this default value and the default we need is false.
If from Ramen perspective you always set a value - this PR won't affect the behavior though, will assist us on a setup of replication without Ramen.

PR #312, opened 2023-03-08
build: increase default resource cpu & memory limits

Rakshith-R:
We now have 100 workers by default and watch a lot of resources such as PVCs,  VolumeAttachments and owned CRs. This leads to Out-Of-Memory kills(OOMKILLs) with just limit of 128Mi.
Therefore, raising limit to 512Mi memory and 1000m cpu./

> Note: I am noticing increase of 50Mi usage with 1000 PVC with reclaimspace annotation. This new limit should give us headroom, while the requested resource is still the same.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,54 +64,9 @@ By listing the `CSIAddonsNode` CRs, the CSI-Addons Controller knows how to
 connect to the side-cars. By checking the supported capabilities of the
 side-cars, it can decide where to execute operations that the user requested.
 
-The preferred way to install the controllers is to use yaml files that are
-made available with the
-[latest release](https://github.com/csi-addons/kubernetes-csi-addons/releases/latest).
-The detailed documentation for the installation of the CSI-Addons Controller using
-the yaml file is present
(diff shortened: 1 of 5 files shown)


nixpanic:
512Mi sounds rather a lot? Could you share how many PVCs you used for testing?

I wonder if this should be documented, with some results for reference.

nixpanic (approved):

Rakshith-R:
> 512Mi sounds rather a lot? Could you share how many PVCs you used for testing?

I've used  about 3000 PVCs with reclaimspace annotation(not mounted) and observed increase of about 50Mi per 1000 PVC
approximately. 
Considering the usage depends on PVCs, Pods(volumeattachements), Nodes, and CSI-Addons CRs, we will easily breach the limit of 128Mi on a production cluster. 

> I wonder if this should be documented, with some results for reference.

I've added a section for this.
I don't think we can definitely add results without extensive testing. 
The resource request is untouched, while only increasing the limit on the resource, therefore we should be fine. 
The user can tweak it based on the cluster requirements and parameters.

Thanks!

yati1998 (approved):

PR #363, opened 2023-06-06
build: use Go 1.20 as Kubernetes packages require that too

nixpanic:
In preparation for updating the Kubernetes dependencies to v1.27 or newer, start using Go 1.20 now already.

### DIFF
--- a/.github/workflows/build-push.yaml
+++ b/.github/workflows/build-push.yaml
@@ -15,10 +15,10 @@ jobs:
       - name: Check out the repo
         uses: actions/checkout@v3
 
-      - name: Install Go 1.19
+      - name: Setup Golang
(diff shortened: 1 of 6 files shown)


.github/workflows/build-push.yaml
```diff
@@ -18,7 +18,7 @@ jobs:
       - name: Install Go 1.19
```
Rakshith-R, line 18:
```suggestion
      - name: Set up Golang
```

nit:
Let's change this to a generic name.
Other workflows also need the same change.

Rakshith-R (changes requested):

Rakshith-R (approved):

Madhu-1 (approved):

Issue #347, opened 2023-05-08
Add support for CSI-Addons-config cm for persistent settings


Rakshith-R:
Currently some settings are being read from cmdline args. 
https://github.com/csi-addons/kubernetes-csi-addons/blob/b9a147c19ce3c54860f0bcb6ce6d6705f36da845/cmd/manager/main.go#L69-L76

However, when csi-addons is deployed as operator using olms, even if the user manages
to change these cmdline args from the CSV, the settings are reverted when upgraded.

We need a CSI-Addons-config configmap for users to configure settings which are persisted throughout upgrade.

The CSI-Addons operator should use options from the CSI-Addons-config configmap if it exists.

cc @nixpanic @Madhu-1

PR #364, opened 2023-06-07
controller: add support for global persistent configuration

Rakshith-R:
This commits adds support for configuration options consumed through a optional configmap. This will enable users to create a configmap to supply options that are persisted throughout upgrades.
Current support parameters are `"reclaim-space-timeout"` and `"max-concurrent-reconciles"`.

Closes: #347

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -17,9 +17,17 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"flag"
 	"os"
 	"time"
+
+	csiaddonsv1alpha1 "github.com/csi-addons/kubernetes-csi-addons/apis/csiaddons/v1alpha1"
+	replicationstoragev1alpha1 "github.com/csi-addons/kubernetes-csi-addons/apis/replication.storage/v1alpha1"
+	controllers "github.com/csi-addons/kubernetes-csi-addons/controllers/csiaddons"
+	replicationController "github.com/csi-addons/kubernetes-csi-addons/controllers/replication.storage"
+	"github.com/csi-addons/kubernetes-csi-addons/internal/connection"
+	"github.com/csi-addons/kubernetes-csi-addons/internal/util"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
 	// to ensure that exec-entrypoint and run can make use of them.
@@ -28,17 +36,12 @@ import (
 	"go.uber.org/zap/zapcore"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/client-go/kubernetes"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
-
-	csiaddonsv1alpha1 "github.com/csi-addons/kubernetes-csi-addons/apis/csiaddons/v1alpha1"
-	replicationstoragev1alpha1 "github.com/csi-addons/kubernetes-csi-addons/apis/replication.storage/v1alpha1"
-	controllers "github.com/csi-addons/kubernetes-csi-addons/controllers/csiaddons"
-	replicationController "github.com/csi-addons/kubernetes-csi-addons/controllers/replication.storage"
-	"github.com/csi-addons/kubernetes-csi-addons/internal/connection"
 	//+kubebuilder:scaffold:imports
 )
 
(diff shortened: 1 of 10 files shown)


cmd/manager/main.go
```diff
@@ -49,6 +50,7 @@ var (
 
 const (
 	defaultTimeout = time.Minute * 3
+	configFile     = "/run/config/csi-addons-config.json"
```
nixpanic, line 53: why use a file, and not the Kubernetes API to access the ConfigMap directly? Users can then place keys directly in the ConfigMap, and do not need to worry about weird indirect json syntax.

docs/csi-addons-config.md
```diff
@@ -0,0 +1,26 @@
+# CSI-Addons Operator Configuration
+
+CSI-Addons Operator can consume configuration from a ConfigMap named `csi-addons-config`
+in the same namespace as the operator. This enables configuration of the operator to persist across
+upgrades. The ConfigMap can support the following configuration options:
```
nixpanic, line 5: Should an empty ConfigMap be created by default? That gives users a hint where to place additional options, and they have an idea what to look for.

cmd/manager/main.go
```diff
@@ -60,19 +62,21 @@ func init() {
...
+		cfg                     util.Config
+	)
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
-	flag.DurationVar(&reclaimSpaceTimeout, "reclaim-space-timeout", defaultTimeout, "Timeout for reclaimspace operation")
-	flag.IntVar(&maxConcurrentReconciles, "max-concurrent-reconciles", 100, "Maximum number of concurrent reconciles")
+	flag.DurationVar(&cfg.ReclaimSpaceTimeout, "reclaim-space-timeout", defaultTimeout, "Timeout for reclaimspace operation")
```
nixpanic, line 77: It is cleaner to have default values in `cfg` set by default. Can you introduce `util.NewConfig()` and move the defaults to `internal/utils/config.go`? It'll make the unit test a little easier too, no need to pass the defaults for every test.

nixpanic (commented):
lgtm, a little more cleaning up is possible now. It would be nice to have a single place to configure default values.

docs/csi-addons-config.md
nixpanic: Ok, if its easier to not have the configmap created by default, that is fine too.

nixpanic (dismissed):

cmd/manager/main.go
```diff
@@ -39,6 +39,8 @@ import (
 	controllers "github.com/csi-addons/kubernetes-csi-addons/controllers/csiaddons"
 	replicationController "github.com/csi-addons/kubernetes-csi-addons/controllers/replication.storage"
 	"github.com/csi-addons/kubernetes-csi-addons/internal/connection"
+	"github.com/csi-addons/kubernetes-csi-addons/internal/util"
+	"k8s.io/client-go/kubernetes"
```
Madhu-1, line 43: add a new line

deploy/controller/csi-addons-config.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: csi-addons-config
+  # replace the namespace with the namespace where the operator is deployed.
+  namespace: csi-addons-system
+data:
+  "reclaim-space-timeout": "5m"
+  "max-concurrent-reconciles": "50"
```
Madhu-1, line 9: match the values for what we have in default , if the user just creates this configmap should not have any impact unless he/she modifies the values

docs/csi-addons-config.md
```diff
@@ -0,0 +1,26 @@
...
+
+| Option                        | Default value   | Description                                   |
+| ----------------------------- | --------------- | --------------------------------------------- |
+| `reclaim-space-timeout`       | `"3m"`          | Timeout for reclaimspace operation            |
+| `max-concurrent-reconciles`   | `"100"`         | Maximum number of concurrent reconciles       |
+
+## Example
```
Madhu-1, line 12: let's not duplicate the configmap here as it adds a burden to maintain at 2 places, just add a reference to the configmap here.

internal/util/config.go
```diff
@@ -0,0 +1,105 @@
...
+// The config file should be in JSON format.
+// Example config file content:
+//
+//	{
+//	  "reclaim-space-timeout": "15m",
+//	  "max-concurrent-reconciles": 25
+//	}
+func (cfg *Config) ReadConfigMap(kubeClient *kubernetes.Clientset) error {
+	cm, err := kubeClient.CoreV1().ConfigMaps(cfg.Namespace).
+		Get(context.Background(), csiAddonsConfigMapName, metav1.GetOptions{})
```
Madhu-1, line 67: Instead of creating a new context here, create one context in the main and use it.

config/manager/manager_auth_proxy_patch.yaml
```diff
@@ -31,6 +31,7 @@ spec:
             memory: 64Mi
       - name: manager
         args:
+        - "--namespace=$(POD_NAMESPACE)"
```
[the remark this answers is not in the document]
Rakshith-R: > I think the `env` section is missing here?

its added above in manager.yaml .  https://github.com/csi-addons/kubernetes-csi-addons/pull/364/files/fcaf7571fe7ee850410e6f2da6fb853daa0a023f#diff-ce0d597a9e5f20a8fe61c3ba5e6185a6b90cfef2c00bb5b5ef3e97b15b79f480 

This yaml patches & overrides args, so I added the namespace arg here.

PR #366, opened 2023-06-12
Add lastsyncbytes and lastsyncduration to vol rep status

yati1998:
This PR adds lastsyncbytes and lastsyncduration to volume replication status and generates updated crds.
Also it adds the logic to update the status in the volume replication object.


apis/replication.storage/v1alpha1/volumereplication_types.go
```diff
@@ -87,10 +87,12 @@ type VolumeReplicationStatus struct {
...
 	// +optional
-	ObservedGeneration int64        `json:"observedGeneration,omitempty"`
-	LastStartTime      *metav1.Time `json:"lastStartTime,omitempty"`
-	LastCompletionTime *metav1.Time `json:"lastCompletionTime,omitempty"`
-	LastSyncTime       *metav1.Time `json:"lastSyncTime,omitempty"`
+	ObservedGeneration int64            `json:"observedGeneration,omitempty"`
+	LastStartTime      *metav1.Time     `json:"lastStartTime,omitempty"`
+	LastCompletionTime *metav1.Time     `json:"lastCompletionTime,omitempty"`
+	LastSyncTime       *metav1.Time     `json:"lastSyncTime,omitempty"`
+	LastSyncBytes      int64            `json:"lastSyncBytes,omitempty"`
```
Rakshith-R, line 94:
```suggestion
	LastSyncBytes      *int64            `json:"lastSyncBytes,omitempty"`
```
This should be a pointer since it is a optional parameter,

Rakshith-R (commented):
Can you please add controller and sidecar changes too (in same pr) and (keep it in draft) / open it for review when completed ?

config/crd/bases/replication.storage.openshift.io_volumereplications.yaml
```diff
@@ -179,6 +179,12 @@ spec:
               lastStartTime:
                 format: date-time
                 type: string
+              lastSyncBytes:
+                format: int64
+                type: integer
+              lastSyncDuration:
+                format: date-time
```
nixpanic, line 186: is this really date-time? is there not a `duration` format, there is no date in this, I think?

nixpanic (commented):
the `duration.proto` file should probably be moved to the commit where it is imported.

nixpanic:
@yati1998 make sure to install protoc v3.19.6, or do not commit the changes in the version of the `.proto` files.

There is also a crd generation issue, it seems that the duration can not have a `format: date-time`.

yati1998:
> @yati1998 make sure to install protoc v3.19.6, or do not commit the changes in the version of the `.proto` files.
> 
> There is also a crd generation issue, it seems that the duration can not have a `format: date-time`.

yeah sure, I am testing with changes, once it is successful, I will update all.

internal/proto/networkfence.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.30.0
-// 	protoc        v3.19.6
```
nixpanic, line 4: You are using an old version to re-generate these files, please use protoc v3.19.6

yati1998:
Working on to revert the protobuf version issue, meanwhile please do review the PR.
cc @nixpanic @Rakshith-R @Madhu-1

nixpanic (approved):

internal/sidecar/service/volumereplication.go
```diff
@@ -219,8 +219,12 @@ func (rs *ReplicationServer) GetVolumeReplicationInfo(
...
 		klog.Errorf("Failed to get last sync time: %v", lastsynctime)
 	}
+	lastsyncduration := resp.GetLastSyncDuration()
+	lastsyncbytes := resp.GetLastSyncBytes()
 
 	return &proto.GetVolumeReplicationInfoResponse{
-		LastSyncTime: lastsynctime,
+		LastSyncTime:     lastsynctime,
+		LastSyncDuration: lastsyncduration,
+		LastSyncBytes:    lastsyncbytes,
```
Madhu-1, line 228: instead of new variables you can directly use resp.GetLastSyncDuration() and resp.GetLastSyncBytes()  here

yati1998:
@nixpanic and @Madhu-1 can you please revisit it again.

internal/proto/replication.proto
```diff
@@ -206,4 +207,10 @@ message GetVolumeReplicationInfoResponse{
   // Holds the last sync time.
   // This field is REQUIRED.
   .google.protobuf.Timestamp last_sync_time = 1;
+  // This field is OPTIONAL.
+  .google.protobuf.Duration last_sync_duration = 2;
+  // This field is OPTIONAL.
+  // A value of 0 is equal to an unspecified field value.
+  // The value of this field MUST NOT be negative.
+  int64 last_sync_bytes = 3;
```
Rakshith-R, line 215: Make sure to match the comments with csi-addons spec or close to it.

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -380,6 +380,15 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
 				instance.Status.LastSyncTime = &lastSyncTime
+				td := info.GetLastSyncDuration()
+				if td != nil {
+					lastSyncDuration := metav1.Duration{Duration: td.AsDuration()}
+					instance.Status.LastSyncDuration = &lastSyncDuration
+				}
+				tb := info.GetLastSyncBytes()
+				if tb != 0 {
+					instance.Status.LastSyncBytes = &tb
+				}
```
Rakshith-R, line 391: Fields should be explicitly set to nil to indicate missing value.
otherwise, they may inherit values that were previously set.

Rakshith-R (changes requested):

internal/proto/replication.proto
yati1998: This has to be edited in spec then, cause it's required to clear it here, will send another PR to update spec comments

controllers/replication.storage/volumereplication_controller.go
yati1998: I am not sure about this, considering the corner case where lastsynctime retains its value and on requeue other values will be nil. Will verify this once and update if it's fine.

nixpanic: On line 91 (wow, what a horrible long function this is?!) `instance` is fetched, so it will have a `.Status.LastSyncBytes` from the previous reconcile. If there were 0 bytes in the last sync, the information would not be correct in the object. So yes, setting it explicitly to an initial value is probably a good thing.

internal/proto/replication.proto
Rakshith-R: I meant to update it here.

controllers/replication.storage/volumereplication_controller.go
yati1998: yeah done

yati1998:
Addressed the changes, @Rakshith-R @nixpanic

internal/proto/replication.proto
```diff
@@ -206,4 +207,16 @@ message GetVolumeReplicationInfoResponse{
   // Holds the last sync time.
   // This field is REQUIRED.
   .google.protobuf.Timestamp last_sync_time = 1;
+  // Holds the last sync duration.
+  // last_sync_duration states the time taken to sync
+  // to execute the last sync operation.
```
Rakshith-R, line 212:
```suggestion
  // last_sync_duration represents the time taken
  // to execute the last sync operation.
```

Rakshith-R (commented):
a small nit

Rakshith-R (approved):
Thanks !

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -380,6 +380,17 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 			if ts != nil {
 				lastSyncTime := metav1.NewTime(ts.AsTime())
 				instance.Status.LastSyncTime = &lastSyncTime
+				instance.Status.LastSyncDuration = nil
+				instance.Status.LastSyncBytes = nil
```
Madhu-1, line 384: i dont think you should make this nil. You should preserve the last data because you already have lastSyncTime, Duration, Byte to give to complete sync information.

controllers/replication.storage/volumereplication_controller.go
Madhu-1: @Rakshith-R @nixpanic IMO we should not set this to nil explicitly https://github.com/csi-addons/kubernetes-csi-addons/pull/366#discussion_r1238259918

controllers/replication.storage/volumereplication_controller.go
Rakshith-R: > i dont think you should make this nil. You should preserve the last data because you already have lastSyncTime, Duration, Byte to give to complete sync information.

@Madhu-1 , 
We'll reach this point only when csi driver gives us a valid new lastSyncTime.

It'll be incorrect to keep previous duration and bytes data with new lastSyncTime according to me.

Madhu-1: okay, just read the code, LGTM

Madhu-1 (approved):

PR #372, opened 2023-06-20
build: add `make docker-generate-protobuf` to update genedated `*.pb.go` files

nixpanic:
It seems that not all developers use an updated version of Fedora. For the contributors that need to update `*.proto` files and re-generate the related `*.pb.go` files, the new `make docker-generate-protobuf` command will be very useful.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@
 CONTROLLER_IMG ?= quay.io/csiaddons/k8s-controller
 SIDECAR_IMG ?= quay.io/csiaddons/k8s-sidecar
 BUNDLE_IMG ?= quay.io/csiaddons/k8s-bundle
+TOOLS_IMG ?= quay.io/csiaddons/tools
 
 # set TAG to a release for consumption in the bundle
 TAG ?= latest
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -175,6 +180,11 @@ docker-build-bundle: bundle
 docker-push-bundle:
 	docker push ${BUNDLE_IMG}
 
+.PHONY: docker-build-tools
+docker-generate-protobuf: ./build/Containerfile.tools
+	buildah bud -f $^ -t ${TOOLS_IMG} .
+	podman run --rm -ti --volume=${PWD}:/go/src/github.com/csi-addons/kubernetes-csi-addons:Z ${TOOLS_IMG} make generate-protobuf
```
Madhu-1, line 186: Not all systems will have buildah and podman and we use docker in most places in this Makefile to run commands.

Madhu-1: i that would be a good option as some developers might also not use Fedora.

nixpanic: Done, please have a look again.

yati1998 (approved):
LGTM

Madhu-1 (approved):

PR #383, opened 2023-06-23
build: use Go 1.20 while building in container

nixpanic:
This also moves to use the quay.io mirror of the Golang container-image from docker.io. Docker Hub has pull restrictions, that can prevent pulling the image if it is done too often. Quay.io does not enforce restrictions like that.


Rakshith-R (approved):
Thanks

Madhu-1 (approved):

PR #396, opened 2023-06-27
controller: fix reclaimspace based on ns annotation

Rakshith-R:
Logic used for determining reclaimspace annotation based on ns annotation and driver support had a bug which caused all the PVCs regardless of driversupport being annotated.
This commit makes sure only pvc with driver which
support reclaimspace is annotated/ requeued.

### DIFF
--- a/controllers/csiaddons/persistentvolumeclaim_controller.go
+++ b/controllers/csiaddons/persistentvolumeclaim_controller.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"reflect"
 	"strings"
@@ -55,6 +56,8 @@ var (
 	rsCronJobScheduleTimeAnnotation = "reclaimspace." + csiaddonsv1alpha1.GroupVersion.Group + "/schedule"
 	rsCronJobNameAnnotation         = "reclaimspace." + csiaddonsv1alpha1.GroupVersion.Group + "/cronjob"
 	csiAddonsDriverAnnotation       = "reclaimspace." + csiaddonsv1alpha1.GroupVersion.Group + "/drivers"
+	ErrConnNotFoundRequeueNeeded    = errors.New("connection not found, requeue needed")
+	ErrScheduleNotFound             = errors.New("schedule not found")
 )
 
 const (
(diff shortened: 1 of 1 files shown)


controllers/csiaddons/persistentvolumeclaim_controller.go
```diff
@@ -134,13 +134,22 @@ func (r *PersistentVolumeClaimReconciler) Reconcile(ctx context.Context, req ctr
 			logger.Error(err, "Failed to get Namespace", "Namespace", pvc.Namespace)
 			return ctrl.Result{}, err
 		}
-		schedule, scheduleFound = getScheduleFromAnnotation(&logger, ns.Annotations)
-		// If the schedule is not found, check whether driver supports the
+		nsSchedule, nsScheduleFound := getScheduleFromAnnotation(&logger, ns.Annotations)
```
Madhu-1, line 137: Do we need new variables? Can we assign this to exist schedule,schduleFound?

Rakshith-R: > Do we need new variables? Can we assign this to exist schedule,schduleFound?

This current form simplifies the code and makes it easier to understand, 
Otherwise it was very difficult to follow the flow.
please read the comments inside the statements.

nixpanic: hmm, this change does not improve the function for me. I have read the comments and change a few times now (that was really needed to understand it), so I think it is not clear enough. If you want to improve the behaviour, split things out in helper functions instead.

Rakshith-R: We are evaluating a lot of conditions in these statements,

I don't believe adding helper function which take in a lot of arguments and also return a lot of arguments will make the code flow any cleaner.

Rakshith-R: Moved the statements into a helper function with lots of io args. 
Hopefully it is simpler to follow now :crossed_fingers:

Madhu-1 (dismissed):

Rakshith-R:
@mergifyio rebase

controllers/csiaddons/persistentvolumeclaim_controller.go
nixpanic: Thanks! I am not a big fan of many return values or arguments to a function either, but this is definitely easier to understand than what there was before.

One thing that could make it even cleaner, is returning an error if requeuing shoud not be needed. This, and maybe the schedule-not-found could be pre-defined errors. You can then check with `errors.Is(...)` what early return from the reconcile should be done. Can you consider doing that, and see if it makes it even cleaner?

Rakshith-R: made the suggested changes, ptal.

Rakshith-R: Both the args are now Err types and will be detected using errors.Is().

nixpanic (approved):

Madhu-1 (approved):

PR #385, opened 2023-06-25
Adding version details to sidecar

saranyareddy24:
Fixes https://github.com/csi-addons/kubernetes-csi-addons/issues/377
Fixes https://github.com/csi-addons/kubernetes-csi-addons/issues/378

The output will be printed as:
Version: v0.6.0
Git Commit: 43ebc4baac5f8893a7d061bab9db78f66640b6b8
Go Version: go1.19.9
Compiler: gc
Platform: linux/amd64

PR #397, opened 2023-06-27
build: include `-version` option for csi-addons executable

nixpanic:
The `Makefile` now contains logic to provide a version (git tag) and the
current git commit hash in a `-version` argument to executables. However
the container-images did not use `make`, and therefor the executables
would not have the version information set.

While updating the building of the executables in the container, the
names of the executables has been prefixed with `csi-addons-`, so that a
listing of running processes on a Kubernetes node does not show
`manager` for the CSI-Addons Operator, but `csi-addons-manager`.

Just like for the manager and sidecar executables, add `-version` as an
option to the `csi-addons` tool.

Updates: #385

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,13 +8,13 @@ ENV GOPATH=/workspace/go
 WORKDIR /workspace/go/src/github.com/csi-addons/kubernetes-csi-addons
 
 # Build
-RUN CGO_ENABLED=0 GOOS=linux go build -mod=vendor -a -o /workspace/manager cmd/manager/main.go
+RUN make build
 
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
 FROM gcr.io/distroless/static:nonroot
 WORKDIR /
-COPY --from=builder /workspace/manager .
+COPY --from=builder /workspace/go/src/github.com/csi-addons/kubernetes-csi-addons/bin/csi-addons-manager .
 USER 65532:65532
 
-ENTRYPOINT ["/manager"]
+ENTRYPOINT ["/csi-addons-manager"]
--- a/build/Containerfile.sidecar
+++ b/build/Containerfile.sidecar
@@ -7,16 +7,14 @@ ADD . /workspace/go/src/github.com/csi-addons/kubernetes-csi-addons
 ENV GOPATH=/workspace/go
 WORKDIR /workspace/go/src/github.com/csi-addons/kubernetes-csi-addons
 
-# Build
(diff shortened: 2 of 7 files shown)


build/Containerfile.sidecar
```diff
@@ -7,16 +7,14 @@ ADD . /workspace/go/src/github.com/csi-addons/kubernetes-csi-addons
...
+# Build the sidecar and csi-addons tool for admin usage and testing
+RUN make build
 
 # Use distroless as minimal base image to package the sidecar binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
 FROM gcr.io/distroless/static:latest
 WORKDIR /
-COPY --from=builder /workspace /usr/bin/
+COPY --from=builder /workspace/go/src/github.com/csi-addons/kubernetes-csi-addons/bin/csi-addons-sidecar /usr/sbin/
+COPY --from=builder /workspace/go/src/github.com/csi-addons/kubernetes-csi-addons/bin/csi-addons /usr/bin/
```
Madhu-1, line 18: it should be copied to `/usr/sbin`?

config/manager/manager.yaml
```diff
@@ -22,7 +22,7 @@ spec:
         runAsNonRoot: true
       containers:
         - command:
-            - /manager
+            - /csi-addons-manager
```
Madhu-1, line 25: same change should also be made to CSV? might be missed to updated CSV with git push

nixpanic: Hmm, I don't think the command is included there. I can not find it in `config/manifests/bases/clusterserviceversion.yaml.in`.

build/Containerfile.sidecar
nixpanic: `/usr/sbin` is the path for executables that act like servers or are for admin users. `csi-addons` is a tool that can be used by any user to talk to the csi-driver. I though it would be more suitable in `/usr/bin`. The distinction between the paths is not very important to me, I am happy yo change either, both should be in the `$PATH` when an admin user logs in the pod.

Dockerfile
```diff
@@ -4,17 +4,17 @@ FROM quay.io/projectquay/golang:1.20 as builder
 # Copy the contents of the repository
 ADD . /workspace/go/src/github.com/csi-addons/kubernetes-csi-addons
 
-ENV GOPATH=/workspace/go
+ENV GOPATH=/workspace/go CGO_ENABLED=0
```
Rakshith-R, line 7: `GOOS=linux` seems to be dropped. 
I hope it makes no difference

Rakshith-R (approved):

Madhu-1 (approved):

Issue #405, opened 2023-07-10
Fix disabled linters in Super Linter 


riya-singhal31:
Super linter has few linters (golanci-lint, hadolint, jscpd, kubernetes kubeconform, markdown, protolint) that are disabled currently and can be enabled by fixing the linters.

PR #407, opened 2023-07-11
fix: enable hadolint linter in super linter

riya-singhal31:
Updates: https://github.com/csi-addons/kubernetes-csi-addons/issues/405

### DIFF
--- a/build/Containerfile.sidecar
+++ b/build/Containerfile.sidecar
@@ -2,7 +2,7 @@
 FROM golang:1.20 as builder
 
 # Copy the contents of the repository
-ADD . /workspace/go/src/github.com/csi-addons/kubernetes-csi-addons
+COPY . /workspace/go/src/github.com/csi-addons/kubernetes-csi-addons
 
 ENV GOPATH=/workspace/go CGO_ENABLED=0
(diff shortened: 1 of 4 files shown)


build/Containerfile.tools
```diff
@@ -1,8 +1,10 @@
+# hadolint ignore=DL3007
```
nixpanic, line 1: Same here, I do not know what `DL3007` is, why it is ignored, or why it can not be fixed.

nixpanic (changes requested):

build/Containerfile.tools
riya-singhal31: It is for "Using latest is prone to errors if the image will ever update. Pin the version explicitly to a release tag". 
So, should I put the latest release version here or keep it ignored?

build/Containerfile.sidecar
```diff
@@ -12,6 +12,7 @@ RUN make build
 
 # Use distroless as minimal base image to package the sidecar binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
+# hadolint ignore=DL3007
```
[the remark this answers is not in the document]
riya-singhal31: Yes Niels, Thanks for the review, Updated.

nixpanic (approved):

Issue #204, opened 2022-08-17
Add NetworkFence capability to csi-addons admin/test tool


nixpanic:
NetworkFence is missing in the test tool. It would be nice to have that added.

https://github.com/csi-addons/kubernetes-csi-addons/tree/main/cmd/csi-addons

@Yuggupta27 could you have a look at this?

PR #410, opened 2023-07-14
test: Adding NetworkFence capability to test tool

karthik-us:
Adding the missing NetworkFence capability to the
cmd/csi-addons admin/test tool.

Fixes: #204

### DIFF
--- a/cmd/csi-addons/main.go
+++ b/cmd/csi-addons/main.go
@@ -42,9 +42,9 @@ type command struct {
 	operation        string
 	persistentVolume string
 	drivername       string
-	secretName       string
-	secretNamespace  string
+	secret           string
 	cidrs            string
+	clusterid        string
 	legacy           bool
 }
 
@@ -59,9 +59,9 @@ func init() {
 	flag.StringVar(&cmd.operation, "operation", "", "csi-addons operation")
 	flag.StringVar(&cmd.persistentVolume, "persistentvolume", "", "name of the PersistentVolume")
 	flag.StringVar(&cmd.drivername, "drivername", "", "name of the CSI driver")
-	flag.StringVar(&cmd.secretName, "secretname", "", "name of the kubernetes secret")
-	flag.StringVar(&cmd.secretNamespace, "secretnamespace", "", "namespace of the kubernetes secret")
+	flag.StringVar(&cmd.secret, "secret", "", "kubernetes secret in the format `namespace/name`")
 	flag.StringVar(&cmd.cidrs, "cidrs", "", "comma separated list of cidrs to fence/unfence")
+	flag.StringVar(&cmd.clusterid, "clusterid", "", "clusterID to fence/unfence")
 	flag.BoolVar(&cmd.legacy, "legacy", false, "use legacy format for old Kubernetes versions")
 	flag.BoolVar(&showVersion, "version", false, "print Version details")
 
--- a/cmd/csi-addons/main.go
+++ b/cmd/csi-addons/main.go
@@ -60,8 +60,8 @@ func init() {
 	flag.StringVar(&cmd.persistentVolume, "persistentvolume", "", "name of the PersistentVolume")
 	flag.StringVar(&cmd.drivername, "drivername", "", "name of the CSI driver")
 	flag.StringVar(&cmd.secret, "secret", "", "kubernetes secret in the format `namespace/name`")
-	flag.StringVar(&cmd.cidrs, "cidrs", "", "comma separated list of cidrs to fence/unfence")
-	flag.StringVar(&cmd.clusterid, "clusterid", "", "clusterID to fence/unfence")
+	flag.StringVar(&cmd.cidrs, "cidrs", "", "comma separated list of cidrs")
+	flag.StringVar(&cmd.clusterid, "clusterid", "", "clusterID")
 	flag.BoolVar(&cmd.legacy, "legacy", false, "use legacy format for old Kubernetes versions")
 	flag.BoolVar(&showVersion, "version", false, "print Version details")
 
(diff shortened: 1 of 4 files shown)


karthik-us:
Keeping it as WIP. Yet to test this and do the doc updates. Reviews are welcome though ;-)

riya-singhal31:
@karthik-us the codespell failures were from the 1st commit itself, it didn't already existed in the code. So you may have to squash the commits.

cmd/csi-addons/main.go
```diff
@@ -56,6 +59,9 @@ func init() {
 	flag.StringVar(&cmd.operation, "operation", "", "csi-addons operation")
 	flag.StringVar(&cmd.persistentVolume, "persistentvolume", "", "name of the PersistentVolume")
 	flag.StringVar(&cmd.drivername, "drivername", "", "name of the CSI driver")
+	flag.StringVar(&cmd.secretName, "secretname", "", "name of the kubernetes secret")
```
nixpanic, line 62: I would prefer to have a single option `--secret=namespace/secretname`, you can easily split the argument on the `/`, and have it a little more convenient for users of the tool. Potentially the namespace could be optional, and you can then assume the namespace is the one where the container is running (maybe you can get that from the environment?).

nixpanic (commented):
Looks good mostly.

Were you able to fence and unfence with this? Can you explain (or copy/paste) how you tested it?

cmd/csi-addons/main.go
karthik-us: Taking the secret details in a single option now as `namespace/name` and we also need to pass the clusterID as Parameter for (un)fenceing.

karthik-us:
Test outputs:

`karthik:~$ kr exec -c csi-addons csi-rbdplugin-provisioner-f88868fb7-7bprx -- csi-addons -operation NetworkFence -clusterid rook-ceph -secret rook-ceph/csi-rbd-secret -cidrs <ip-addr>/20 -endpoint unix:///csi/csi-addons.sock`
`Network fence successful`

`kr exec -c csi-addons csi-rbdplugin-provisioner-f88868fb7-7bprx -- csi-addons -operation NetworkUnFence -clusterid rook-ceph -secret rook-ceph/csi-rbd-secret -cidrs <ip-addr>/20 -endpoint unix:///csi/csi-addons.sock`
`Network unfence successful`

riya-singhal31 (approved):
LGTM

cmd/csi-addons/networkfence.go
```diff
@@ -0,0 +1,150 @@
...
+
+	parameters      map[string]string
+	secretName      string
+	secretNamespace string
+	cidrs           []string
+}
+
+var _ = registerOperation("NetworkUnFence", &NetworkUnFenceServer{})
+
+func (ns *NetworkUnFenceServer) Init(c *command) error {
```
Rakshith-R, line 103: The steps are the same for Network[Un]Fence Init functions, 
maybe abstract it into a common helper verifyNetworkFenceParameters() ?

cmd/csi-addons/main.go
```diff
@@ -56,6 +59,9 @@ func init() {
 	flag.StringVar(&cmd.operation, "operation", "", "csi-addons operation")
 	flag.StringVar(&cmd.persistentVolume, "persistentvolume", "", "name of the PersistentVolume")
 	flag.StringVar(&cmd.drivername, "drivername", "", "name of the CSI driver")
+	flag.StringVar(&cmd.secret, "secret", "", "kubernetes secret in the format `namespace/name`")
+	flag.StringVar(&cmd.cidrs, "cidrs", "", "comma separated list of cidrs to fence/unfence")
```
nixpanic, line 63: no need to mention fence/unfence, possibly these parameters will be used for other operations in the future

cmd/csi-addons/networkfence.go
nixpanic: There could be a base struct that gets extended with the `Execute()` function.

nixpanic: Something like this?

```go
type networkFenceBase {
    grpcClient
    parameters ...
}

func (nfb *networkFenceBase) Init() {
    ...
}

type NetworkFenceServer networkFenceBase

var _ = registerOperation("NetworkFence", &NetworkFenceServer{})

func (nfs *NetworkFenceServer) Execute() {
    ...
}
```

karthik-us: @nixpanic the `registerOperation`, `Init` & `Execute` all expects same struct to be passed for that operation, so we can not make use of a base sturct for `Init` & the extended one in `Execute`.
@Rakshith-R Since the arguments are of different types and direct conversion of struct pointers are not allowed, it doesn't seem feasible to have it under a generic function.
Please feel free to let me know if there is any other way that I am missing with which we can acheive this.

cmd/csi-addons/main.go
karthik-us: Done.

cmd/csi-addons/networkfence.go
karthik-us: Yes that did work. Thanks for the help. Updated the patch.

cmd/csi-addons/networkfence.go
```diff
@@ -0,0 +1,119 @@
...
+	// inherit Connect() and Close() from type grpcClient
+	grpcClient
+
+	parameters      map[string]string
+	secretName      string
+	secretNamespace string
+	cidrs           []string
+}
+
+func (ns *networkFenceBase) Init(c *command) error {
```
Rakshith-R, line 39: Can we add unit test for this function to make sure we don't break it in the future?

karthik-us: Done. PTAL.

Rakshith-R (approved):
Thanks, lgtm

riya-singhal31 (approved):

Issue #422, opened 2023-08-07
All the Controller Operations should reach the one Controller (active) not multiple Controllers


Madhu-1:
As of today, the kubernetes csi addons try to connect to the random controller that are registered and try to make the RPC calls to the random controller. This can create a problem if the csi driver has implemented some internal locking mechanism or has some local cache for the lifetime of that instance.

Example as below:-

CephCSI runs deployments for Replication/Reclaimspace etc and we will have two instances running. CephCSI Internally takes a lock and processes a request one at a time based on its internal logic. With the current kubernetes sidecar, it's not a problem because the sidecar runs with a leader election and only one can process a request but with kubernetes-csiaddons it becomes a problem as we don't have any such mechanism to reach the same controller/deployment which is processing the requests.

The request is to provide this kind of functionality so that it will be helpful for the CSI driver who is having this kind of requirement and moreover not to run active/active models as it can lead to many different models.

nixpanic:
The csi-addons side-car should probably use a similar mechanism as the csi-provisioner and others. When the csi-addons sidecar detects that the csi-driver provides the `CONTROLLER_SERVICE` capability, it should try to obtain a [lease](https://kubernetes.io/docs/concepts/architecture/leases/#custom-workload) for the driver.

The kubernetest-csi-addons Operator should only send controller operations to the csi-driver pod that holds the lease for that particular driver.

Rakshith-R:
- Add another field `ConnectedAt` in Connection struct 
https://github.com/csi-addons/kubernetes-csi-addons/blob/0fe08b80cc8f98b70050b8ec128a2949b95f7d98/internal/connection/connection.go#L30-L36

- Fill it while adding connection to pool
https://github.com/csi-addons/kubernetes-csi-addons/blob/0fe08b80cc8f98b70050b8ec128a2949b95f7d98/internal/connection/connection_pool.go#L40

- Sort based on ConnectedAt timestamp and use the oldest client

nixpanic:
A `ConnectedAt` timestamp may not be stable enough. I guess it is possible that the Kubernetes API is not working correctly anymore (some partial network issue?), but the operator and side-car are still connected. This would cause any operation on the side-car to fail when the Kubernetes API is used (almost all operations need to fetch secrets).

Using the Leases API will make sure the Kubernetes API is functioning the way it should. If it does not, the side-car can not extend the lease and it will expire, allowing an other side-car to become the acting controller service.

[...]

PR #496, opened 2023-12-12
connection: track Namespace of driver Pods to locate Leases

nixpanic:
The CSIAddonsNode objects are located in the Namespace where CSI-drivers are running. Use this Namespace in the Connection struct, so that the Lease for a group of CSI-Addons ControllerServers can be found.

Updates: #422

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -119,7 +119,7 @@ help: ## Display this help.
 
 .PHONY: manifests
 manifests: controller-gen kustomize ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
-	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="{./apis/...,./cmd/...,./controllers/...,./sidecar/...}" output:crd:artifacts:config=config/crd/bases
+	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="{./apis/...,./cmd/...,./controllers/...,./internal/...,./sidecar/...}" output:crd:artifacts:config=config/crd/bases
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${CONTROLLER_IMG} rbac-proxy=${RBAC_PROXY_IMG}
 	$(KUSTOMIZE) build config/crd > deploy/controller/crds.yaml
 	$(KUSTOMIZE) build config/rbac > deploy/controller/rbac.yaml
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -22,6 +22,14 @@ rules:
   - list
   - watch
 - apiGroups:
+  - coordination.k8s.io
+  resources:
+  - leases
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
   - ""
   resources:
   - namespaces
--- a/controllers/csiaddons/csiaddonsnode_controller.go
+++ b/controllers/csiaddons/csiaddonsnode_controller.go
@@ -118,7 +118,7 @@ func (r *CSIAddonsNodeReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	}
 
 	logger.Info("Connecting to sidecar")
-	newConn, err := connection.NewConnection(ctx, endPoint, nodeID, driverName, csiAddonsNode.Name)
+	newConn, err := connection.NewConnection(ctx, endPoint, nodeID, driverName, csiAddonsNode.Namespace, csiAddonsNode.Name)
 	if err != nil {
 		logger.Error(err, "Failed to establish connection with sidecar")
 
--- a/internal/connection/connection_pool.go
+++ b/internal/connection/connection_pool.go
@@ -27,6 +27,8 @@ import (
 	"github.com/csi-addons/kubernetes-csi-addons/internal/util"
 )
 
+//+kubebuilder:rbac:groups=coordination.k8s.io,resources=leases,verbs=get;list;watch
+
 // ConnectionPool consists of map of Connection objects and
 // methods Put, Get & Delete which operates with required rw locks
 // to ensure consistency.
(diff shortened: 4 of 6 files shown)


internal/connection/connection_pool.go
```diff
@@ -103,13 +103,38 @@ func (cp *ConnectionPool) GetByNodeID(driverName, nodeID string) map[string]*Con
...
+	}
+
+	// should be impossible to get here, all Connections have a Namespace
+	return ""
+}
+
 // GetLeaderByDriver finds the holder of the lease for the driver, and returns
 // the connection to that particular CSI-Addons sidecar.
+//
+// +kubebuilder:rbac:groups=coordination.k8s.io,resources=leases,verbs=get
```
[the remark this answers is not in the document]
Rakshith-R: let's add it along with other rbac explicitly here https://github.com/csi-addons/kubernetes-csi-addons/blob/b15ae1a35dd181922466020a98dd3fbe4bc1c648/controllers/csiaddons/networkfence_controller.go#L73 

and for separately for each  controller too so we don't run into weird edge cases later on.

Makefile
```diff
@@ -119,7 +119,7 @@ help: ## Display this help.
 
 .PHONY: manifests
 manifests: controller-gen kustomize ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
-	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="{./apis/...,./cmd/...,./controllers/...,./sidecar/...}" output:crd:artifacts:config=config/crd/bases
+	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="{./apis/...,./cmd/...,./controllers/...,./internal/...,./sidecar/...}" output:crd:artifacts:config=config/crd/bases
```
Rakshith-R, line 122: This will not be needed after these changes https://github.com/csi-addons/kubernetes-csi-addons/pull/496/files#r1424922220

internal/connection/connection_pool.go
```diff
@@ -103,13 +103,38 @@ func (cp *ConnectionPool) GetByNodeID(driverName, nodeID string) map[string]*Con
...
+	for _, conn := range cp.pool {
+		if conn.DriverName != driverName {
+			continue
+		}
+
+		return conn.Namespace
+	}
+
+	// should be impossible to get here, all Connections have a Namespace
+	return ""
```
Rakshith-R, line 119: This may happen if we don't have any connections added to the pool yet.
Let's error out. 
We should have connections added when the request is looped again by reconciler hoepfully.

Rakshith-R (changes requested):

internal/connection/connection_pool.go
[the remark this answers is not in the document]
nixpanic: that seems to work, it gets picked up there. I don't underastand why it won't work in `connection_pool.go` :shrug:

config/rbac/role.yaml
```diff
@@ -21,6 +21,12 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - coordination.k8s.io
+  resources:
+  - leases
+  verbs:
+  - get
```
Rakshith-R, line 29:
```suggestion
  - get
  - list
  - watch
```
We all three due to informer cache.
It still gave error with just `get` verb

controllers/csiaddons/csiaddonsnode_controller.go
```diff
@@ -51,6 +51,7 @@ type CSIAddonsNodeReconciler struct {
 	ConnPool *connection.ConnectionPool
 }
 
+//+kubebuilder:rbac:groups=coordination.k8s.io,resources=leases,verbs=get
```
Rakshith-R, line 54:
```suggestion
```
We don't need to find leader in csi-addons-node controller right ?

controllers/csiaddons/persistentvolumeclaim_controller.go
```diff
@@ -64,6 +64,7 @@ const (
 	defaultSchedule = "@weekly"
 )
 
+//+kubebuilder:rbac:groups=coordination.k8s.io,resources=leases,verbs=get
```
Rakshith-R, line 67:
```suggestion
```
We don't need to find leader in pvc controller right ?

controllers/csiaddons/reclaimspacecronjob_controller.go
```diff
@@ -54,6 +54,7 @@ var (
 	scheduledTimeAnnotation = csiaddonsv1alpha1.GroupVersion.Group + "/scheduled-at"
 )
 
+//+kubebuilder:rbac:groups=coordination.k8s.io,resources=leases,verbs=get
```
Rakshith-R, line 57:
```suggestion
```
We don't need to find leader in rscronjob controller too right ?

internal/connection/connection_pool.go
```diff
@@ -103,13 +103,45 @@ func (cp *ConnectionPool) GetByNodeID(driverName, nodeID string) map[string]*Con
...
+	}
+
+	// should be impossible to get here, all Connections have a Namespace
+	return "", fmt.Errorf("failed to find the namespace where driver %q is running", driverName)
+}
+
 // GetLeaderByDriver finds the holder of the lease for the driver, and returns
 // the connection to that particular CSI-Addons sidecar.
+//
+// +kubebuilder:rbac:groups=coordination.k8s.io,resources=leases,verbs=get
```
Rakshith-R, line 127:
```suggestion
```

We don't need it here.

internal/connection/connection_pool.go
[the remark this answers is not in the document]
Rakshith-R: > that seems to work, it gets picked up there. I don't underastand why it won't work in `connection_pool.go` 🤷

probably some complex kubebuilder/controller gen logic

Rakshith-R (changes requested):
left few comments.
It looks almost ready to me.

controllers/csiaddons/csiaddonsnode_controller.go
nixpanic: maybe not

controllers/csiaddons/persistentvolumeclaim_controller.go
nixpanic: we do, because offline reclaim-space needs to run on the leading controller

controllers/csiaddons/reclaimspacecronjob_controller.go
nixpanic: we do, because offline reclaim-space needs to run on the leading controller

internal/connection/connection_pool.go
nixpanic: ah, yes, that is a left-over.

config/rbac/role.yaml
nixpanic: ah, ok, will add them

internal/connection/connection_pool.go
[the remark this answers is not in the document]
nixpanic: Placing it in a different location in `connection_pool.go` seems to work! That seems much cleaner, as it is the location where leases are checked.

Rakshith-R (approved):
Thanks!

riya-singhal31 (approved):

PR #529, opened 2024-03-05
deploy: run all containers with read-only filesystem

nixpanic:
Prevent potential abuse of the container storage a little more, by
running all containers with a read-only filesystem.

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -34,6 +34,7 @@ spec:
                   fieldPath: metadata.namespace
           securityContext:
             allowPrivilegeEscalation: false
+            readOnlyRootFilesystem: true
           livenessProbe:
             httpGet:
               path: /healthz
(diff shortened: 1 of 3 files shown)


config/manager/manager.yaml
```diff
@@ -34,6 +34,7 @@ spec:
                   fieldPath: metadata.namespace
           securityContext:
             allowPrivilegeEscalation: false
+            readOnlyRootFilesystem: true
```
Rakshith-R, line 37: This doesn't cover kube rbac proxy container.
Let's set this in podSpec.securityContext layer ? 

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#podspec-v1-core

nixpanic: This option is not available for the podSpec.securityContext, it can only be set per container.

I indeed missed the kube-rbac-proxy, will add that one too.

Rakshith-R: :+1: ack.

Rakshith-R (approved):
Thanks, LGTM !

iPraveenParihar (approved):

nixpanic:
@mergifyio refresh

PR #530, opened 2024-03-05
ci: fix TestGetNextSchedule failure

karthik-us:
The test case was failing if the test was run at some particular time of the day.

For ex: If the test was triggered @0:15:0, the last scheduled time will be 23:15:0 of the previous day since we are setting it to an hour earlier than the current time. Now when we calculate the getNextSchedule, the next schedule will point to the current day, when the cron job is scheduled to run at midnight every day. So in this case, the lastMissed will be pointing to a valid time. This leads to some test cases to get unexpected value for the lastMissed time which expects it to be time.Time{}. Though we get the expected wantErr in such cases too, since we are comparing the other return values seperately.

To avoid such failures, this patch checks for multiple expected last scheduled times.
Also the last scheduled time is aligned with the cron job schedule to keep it relative.

### DIFF
--- a/controllers/csiaddons/reclaimspacecronjob_controller_test.go
+++ b/controllers/csiaddons/reclaimspacecronjob_controller_test.go
@@ -18,7 +18,6 @@ package controllers
 
 import (
 	"fmt"
-	"reflect"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


controllers/csiaddons/reclaimspacecronjob_controller_test.go
```diff
@@ -203,7 +204,7 @@ func TestGetNextSchedule(t *testing.T) {
 				t.Errorf("getNextSchedule() error = %v, wantErr %v", err, tt.wantErr)
 				return
 			}
-			if !reflect.DeepEqual(gotLastMissed, tt.lastMissed) {
+			if !reflect.DeepEqual(gotLastMissed, tt.lastMissed) && !reflect.DeepEqual(gotLastMissed, time.Time{}) {
```
nixpanic, line 207: Instead of using `reflect`, you could use https://pkg.go.dev/time#Time.Equal maybe?

karthik-us: Done.

nixpanic (approved):

Rakshith-R (approved):

nixpanic:
@mergifyio refresh

PR #540, opened 2024-03-29
replace custom ContainsInSlice() with standard slices.Contains()

nixpanic:
With Go 1.21 the "slices" package can be used for many operations on slices. The custom ContainsInSlice() function can be replaced by the standard slices.Contains().


Rakshith-R (approved):
Thanks, lgtm !

karthik-us (approved):
LGTM.

PR #596, opened 2024-06-19
Replication: set volumeID in ReplicationSource for the volumeReplication

Madhu-1:
update csi-addons spec to the latest commit and set voluemID in the ReplicationSource for the volume replication.


internal/sidecar/service/volumereplication.go
```diff
@@ -63,14 +63,24 @@ func (rs *ReplicationServer) EnableVolumeReplication(
 		klog.Errorf("Failed to get secret %s in namespace %s: %v", req.GetSecretName(), req.GetSecretNamespace(), err)
 		return nil, status.Error(codes.Internal, err.Error())
 	}
+	repReq := &csiReplication.EnableVolumeReplicationRequest{
+		ReplicationId: req.ReplicationId,
```
nixpanic, line 67: attributes of gRPC requests should not be accessed directly, use `req.GetReplicationId()` instead. And everywhere else, specially in new code.

nixpanic (changes requested):

nixpanic (approved):

Rakshith-R (approved):

PR #597, opened 2024-06-20
replication: add missing VRC name for VGR

Madhu-1:
we need to have a VRC name to create the VR, this PR adds a new field to the VGR to pass down the VRC name for the VolumeReplication object.


Rakshith-R (approved):

nixpanic:
@Madhu-1 , the abbreviations make it really difficult to understand the explanation.

I don't understand why a `VolumeGroupReplication` needs the `VolumeReplicationClassName`, why is `VolumeGroupReplicationClassName` not sufficient?

Or, should the `VolumeReplicationClassName` not be a member of `VolumeGroupReplicationClass`?

nixpanic (approved):

PR #585, opened 2024-06-06
reclaimspace: support for storageclass annotation

iPraveenParihar:
This commit adds support for auto ReclaimSpace by StorageClass annotations. 
For newly created PVCs ReclaimSpaceCronjob will be created for PVCs associated
with a StorageClass that has the ReclaimSpace annotation.

PR #604, opened 2024-06-26
reclaimspace: fix schedule handling

iPraveenParihar:
In my earlier PR https://github.com/csi-addons/kubernetes-csi-addons/pull/585, I missed few changes 😞 

This commit addresses a bug in `determineScheduleAndRequeue` where the check for schedule from the StorageClass annotations is never reached.

### DIFF
--- a/controllers/csiaddons/persistentvolumeclaim_controller.go
+++ b/controllers/csiaddons/persistentvolumeclaim_controller.go
@@ -278,29 +278,39 @@ func (r *PersistentVolumeClaimReconciler) determineScheduleAndRequeue(
 		return "", err
 	}
 	schedule, scheduleFound = getScheduleFromAnnotation(logger, ns.Annotations)
-	if !scheduleFound {
-		return "", ErrScheduleNotFound
-	}
+
 	// If the schedule is found, check whether driver supports the
 	// space reclamation using annotation on namespace and registered driver
 	// capability for decision on requeue.
-
-	requeue, supportReclaimspace := r.checkDriverSupportReclaimsSpace(logger, ns.Annotations, driverName)
-	if supportReclaimspace {
-		// if driver supports space reclamation,
-		// return schedule from ns annotation.
-		return schedule, nil
+	if scheduleFound {
+		requeue, supportReclaimspace := r.checkDriverSupportReclaimsSpace(logger, ns.Annotations, driverName)
+		if supportReclaimspace {
+			// if driver supports space reclamation,
+			// return schedule from ns annotation.
+			return schedule, nil
+		}
+		if requeue {
+			// The request needs to be requeued for checking
+			// driver support again.
+			return "", ErrConnNotFoundRequeueNeeded
+		}
 	}
-	if requeue {
-		// The request needs to be requeued for checking
-		// driver support again.
-		return "", ErrConnNotFoundRequeueNeeded
+
+	// For static provisioned PVs, StorageClassName is empty.
+	if len(*pvc.Spec.StorageClassName) == 0 {
+		logger.Info("StorageClassName is empty")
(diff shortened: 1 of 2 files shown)


Madhu-1:
@iPraveenParihar can you please add a unit testing to cover a few test cases for PVC and SC and not found case for namespace for now?

iPraveenParihar:
> @iPraveenParihar can you please add a unit testing to cover a few test cases for PVC and SC and not found case for namespace for now?

@Madhu-1  added!

Rakshith-R (approved):

controllers/csiaddons/persistentvolumeclaim_controller.go
```diff
@@ -278,23 +278,22 @@ func (r *PersistentVolumeClaimReconciler) determineScheduleAndRequeue(
...
+		if supportReclaimspace {
+			// if driver supports space reclamation,
+			// return schedule from ns annotation.
+			return schedule, nil
+		}
+		if requeue {
+			// The request needs to be requeued for checking
+			// driver support again.
+			return "", ErrConnNotFoundRequeueNeeded
+		}
```
Madhu-1, line 296: I was thinking about the same where storageclass is deleted after the PVC creation, we need to consider not found error as the scheduleNotFound so as not to fail the reconcile because it's possible.

iPraveenParihar: well IMO, we should be returning the precise error so that the users are aware of the exact state of the resources and they get it resolved?

Madhu-1: it would have made sense it was a VR controller but this is a pvc controller and user just created the PVC object nothing else. can you please check it again and let me know if am missing something.

iPraveenParihar: > it would have made sense it was a VR controller but this is a pvc controller and user just created the PVC object nothing else. can you please check it again and let me know if am missing something.

@Madhu-1, Apologies. I quite didn't understand this part.
I understand your concern about reconcile failures when StorageClass is not found. However, masking the SC notFound error with scheduleNotFound error will lead to confusion? Its users responsibility to take appropriate actions when SC notFound error is thrown?

Or did I miss your point?

Madhu-1: take the case of what happens below

* PVC is created
* Storageclass is deleted
* The reconciliation triggered for the PVC as we list all pvc and check for annotation, if the SC is not found we keep on reqeueing this PVC which never completes, isn't it?

iPraveenParihar: Hmm, okay. I was talking about the same scenario. In such cases, there are two ways to handle it:

- Return a "StorageClass not found" error, and the reconciliation process continues to requeue the PVC. This allows the user to take appropriate action on the error.
- Return a "ScheduleNotFound" error and exit the reconciliation process. This approach masks the "StorageClass not found" error, potentially causing confusion for the user.

If masking the error isn't a concern, we can agree on returning the ScheduleNotFound error?

Madhu-1: as both are independent objects once created, masking the error makes sense to me and avoids a lot of reconciles, @nixpanic @Rakshith-R WDYT?

Rakshith-R: I would prefer if we just log the sc not found info and move on.

iPraveenParihar:
@Madhu-1 @Rakshith-R PTAL.

Rakshith-R (approved):

Madhu-1 (approved):

PR #605, opened 2024-06-26
Support VolumeGroupReplication with VR

Madhu-1:
This PR adds the support for replicating the group with volumegrouplication


controllers/replication.storage/pvc.go
```diff
@@ -59,6 +60,35 @@ func (r VolumeReplicationReconciler) getPVCDataSource(logger logr.Logger, req ty
 	return pvc, pv, nil
 }
 
+func (r *VolumeReplicationReconciler) getVolumeGroupReplication(logger logr.Logger, req types.NamespacedName) (*replicationv1alpha1.VolumeGroupReplicationContent, error) {
```
iPraveenParihar, line 63:
```suggestion
func (r *VolumeReplicationReconciler) getVolumeGroupReplicationContent(logger logr.Logger, req types.NamespacedName) (*replicationv1alpha1.VolumeGroupReplicationContent, error) {
```
?

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -158,6 +164,20 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		}
 
 		volumeHandle = pv.Spec.CSI.VolumeHandle
+		logger.Info("volume handle", "VolumeHandleName", volumeHandle)
+	case volumeGroupReplicationDataSource:
+		vgrc, vgrErr = r.getVolumeGroupReplicationContent(logger, nameSpacedName)
+		if vgrErr != nil {
+			logger.Error(vgrErr, "failed to get VolumeGroupReplication", "VGRName", instance.Spec.DataSource.Name)
```
iPraveenParihar, line 171:
```suggestion
			logger.Error(vgrErr, "failed to get VolumeGroupReplicationContent", "VGRName", instance.Spec.DataSource.Name)
```

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -158,6 +164,20 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+			setFailureCondition(instance)
+			uErr := r.updateReplicationStatus(instance, logger, getCurrentReplicationState(instance), vgrErr.Error())
+			if uErr != nil {
+				logger.Error(uErr, "failed to update volumeReplication status", "VRName", instance.Name)
+			}
+			return ctrl.Result{}, vgrErr
+		}
+		groupHandle = vgrc.Spec.VolumeGroupReplicationHandle
+		logger.Info("Group handle", "GroupHandleName", replicationHandle)
```
iPraveenParihar, line 180:
```suggestion
		logger.Info("Group handle", "GroupHandleName", groupHandle)
```?

**Review by iPraveenParihar — CHANGES_REQUESTED** [review]
typos

File: controllers/replication.storage/volumereplication_controller.go (hunk 2 above)
Comment (Madhu-1) [reply controllers/replication.storage/volumereplication_controller.go:171]: we cannot exactly tell what is missing VolumeGroupReplicationContent or VolumeGroupReplication, IMO we should keep VolumeGroupReplication similar to PVC error above

**Review by nixpanic — COMMENTED** [review]
few minor things, and a larger interface improvement request to simplify the code and make maintenance easier.

File: internal/proto/replication.proto [hunk 4]
```diff
@@ -220,3 +196,28 @@ message GetVolumeReplicationInfoResponse{
...
+    string volume_group_id = 1;
+  }
+
+  oneof type {
+    // Volume source type
+    VolumeSource volume = 1;
+    // Volume group source type
+    VolumeGroupSource volumegroup = 2;
+  }
+}
```
Rakshith-R, line 223:
```suggestion
message ReplicationSource {
    oneof source {
    // Volume source type
    VolumeSource volume = 1;
    // Volume group source type
    VolumeGroupSource volumegroup = 2;
   }
  }

// VolumeSource contains the details about the volume to be replication
  message VolumeSource {
    // Contains identity information for the existing volume.
    // This field is REQUIRED.
    string volume_id = 1;
  }
  // VolumeGroupSource contains the details about
  // the volume group to be replication
  message VolumeGroupSource {
    // Contains identity information for the existing volume group.
    // This field is REQUIRED.
    string volume_group_id = 1;
}
```



https://protobuf.dev/programming-guides/proto3/#oneof
https://www.tutorialspoint.com/protobuf/protobuf_compound_data_types.htm

oneOf type should be inside the message ?

Rakshith-R: https://github.com/csi-addons/spec/tree/main/replication#replicationsource 
Now I see that its present in csiaddons spec too.

Or am i mistaken :thinking:

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -203,17 +223,19 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 
 			return reconcile.Result{}, err
 		}
+		switch instance.Spec.DataSource.Kind {
```
Nikhil-Ladha, line 226: All the Volume group replication related finalizers will be handled by VGR_contorller, so I don't think we need a switch case here as such, we can simply use an if condition for the pvcDataSource.
Or, do we plan to add any further finalizers here for vgr?

internal/proto/replication.proto
Rakshith-R: please disregard this comment.
I was mistaken :sweat:  I didn't read the indentation properly

more examples in https://github.com/container-storage-interface/spec/blob/master/spec.md

controllers/replication.storage/volumereplication_controller.go
Madhu-1: @Nikhil-Ladha Each PVC needs to be handled by the VGR now, we cannot handle PVC in VR as VR will not get any trigger at PVC level for group replication, based on the dataSource this will set the finalizers and owner on the VGR

Nikhil-Ladha: Ack

controllers/replication.storage/finalizers.go
```diff
@@ -94,3 +95,34 @@ func (r *VolumeReplicationReconciler) removeFinalizerFromPVC(logger logr.Logger,
...
+			return fmt.Errorf("failed to add finalizer (%s) to VolumeGroupReplication resource"+
+				" (%s/%s) %w",
+				vgrReplicationFinalizer, vgr.Namespace, vgr.Name, err)
+		}
+	}
+
+	return nil
+}
+
+// removeFinalizerFromPVC removes the VR finalizer on VolumeGroupReplication.
```
Nikhil-Ladha, line 114:
```suggestion
// removeFinalizerFromVGR removes the VR finalizer on VolumeGroupReplication.
```

internal/sidecar/service/volumereplication.go
```diff
@@ -290,3 +257,28 @@ func (rs *ReplicationServer) GetVolumeReplicationInfo(
...
+	if req.GetVolume() != nil {
+		src.Type = &csiReplication.ReplicationSource_Volume{Volume: &csiReplication.ReplicationSource_VolumeSource{
+			VolumeId: req.GetVolume().GetVolumeId(),
+		}}
+	}
+	if req.GetVolumegroup() != nil {
+		src.Type = &csiReplication.ReplicationSource_Volumegroup{Volumegroup: &csiReplication.ReplicationSource_VolumeGroupSource{
+			VolumeGroupId: req.GetVolumegroup().GetVolumeGroupId(),
+		}}
+	}
```
Rakshith-R, line 281:
```suggestion
        switch {
	case req.GetVolume() != nil :
		src.Type = &csiReplication.ReplicationSource_Volume{
		  Volume: &csiReplication.ReplicationSource_VolumeSource{
			VolumeId: req.GetVolume().GetVolumeId(),
		    } 
		  }
	case req.GetVolumegroup() != nil :
		src.Type = &csiReplication.ReplicationSource_Volumegroup{
		Volumegroup: &csiReplication.ReplicationSource_VolumeGroupSource{
			VolumeGroupId: req.GetVolumegroup().GetVolumeGroupId(),
		  }
		}
	}
```

This ensure we only set one of them ?

cmd/csi-addons/replication.go
```diff
@@ -81,6 +97,35 @@ type EnableVolumeReplication struct {
 	VolumeReplicationBase
 }
 
+func (v VolumeReplicationBase) setReplicationSource(req *proto.ReplicationSource) error {
+	if req == nil {
+		return errors.New("replication source is not set")
+	}
+	if v.volumeID == "" && v.groupID == "" {
+		return errors.New("volumeID or groupID is not set")
```
Rakshith-R, line 105:
```suggestion
		return errors.New("both volumeID and groupID is not set")
```

internal/sidecar/service/volumereplication.go
Madhu-1: AFAIC this can never be possible because its oneOf either one can be set but addressed the changes as requested

cmd/csi-addons/replication.go
```diff
@@ -59,19 +60,34 @@ func (rep *VolumeReplicationBase) Init(c *command) error {
...
+			return errors.New("volume group ID is not set")
+		}
+		rep.groupID = vgrc.Spec.VolumeGroupReplicationHandle
 
-	if pv.Spec.CSI.VolumeHandle == "" {
-		return errors.New("volume ID is not set")
 	}
-	rep.volumeID = pv.Spec.CSI.VolumeHandle
 
 	return nil
```
Rakshith-R, line 92: return in each at the end of each if block
and error out finally since neither pv nor vgrc is set ?

cmd/csi-addons/replication.go
```diff
@@ -81,6 +97,35 @@ type EnableVolumeReplication struct {
 	VolumeReplicationBase
 }
 
+func (v VolumeReplicationBase) setReplicationSource(req *proto.ReplicationSource) error {
+	if req == nil {
```
Rakshith-R, line 101: use 
```
switch{
  case (condition):
  ..
   case (condition): 
}
```

format instead of continuous if blocks?

Rakshith-R: + follow order of

- req ==nil
- both volumeID and groupID set
- volumeID set
- groupID set
outside error out telling neither is set.

Like followed in Init() function ?

cmd/csi-addons/replication.go
```diff
@@ -231,7 +287,13 @@ func (rep *GetVolumeReplicationInfo) Execute() error {
...
 		SecretName:      rep.secretName,
 		SecretNamespace: rep.secretNamespace,
-		VolumeId:        rep.volumeID,
+		ReplicationSource: &proto.ReplicationSource{
+			Type: &proto.ReplicationSource_Volume{
+				Volume: &proto.ReplicationSource_VolumeSource{
+					VolumeId: rep.volumeID,
+				},
+			},
+		},
```
Rakshith-R, line 296: This will fail when its called for group ?

Rakshith-R (commented):
Almost everything looks good to me.

small nit: I noticed there's no `group-replication-client_test.go` similar to `volume-replication-client_test.go`. maybe we need to add that test file or combine both volume and group to be tested as interface functions in `replication-client_test.go`?

Madhu-1:
> Almost everything looks good to me.
> 
> small nit: I noticed there's no `group-replication-client_test.go` similar to `volume-replication-client_test.go`. maybe we need to add that test file or combine both volume and group to be tested as interface functions in `replication-client_test.go`?

@Rakshith-R I did consider that I don't see any value for `volume-replication-client_test.go` as its not testing anything, i do have a plan to remove it in followup and replace that with proper tests

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -203,17 +228,32 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
-		}
+			if err = r.addFinalizerToPVC(logger, pvc); err != nil {
+				logger.Error(err, "Failed to add PersistentVolumeClaim finalizer")
 
-		if err = r.addFinalizerToPVC(logger, pvc); err != nil {
-			logger.Error(err, "Failed to add PersistentVolumeClaim finalizer")
+				return reconcile.Result{}, err
+			}
+		case volumeGroupReplicationDataSource:
+			err = r.annotateVolumeGroupReplicationWithOwner(ctx, logger, req.Name, vgr)
```
Nikhil-Ladha, line 245: Shouldn't VGR be the owner of VR?
I mean the VGR controller is creating the VR, and we will be setting controllerRef on the VR in the VGR controller, so any special benefit of using this?
Also, I am adding this annotation `replication.storage.openshift.io/volumegroupref` on both VR, and VGRContent created by VGR to fetch the VGR later on during reconcile.

controllers/replication.storage/volumereplication_controller.go
```diff
@@ -203,17 +228,32 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+			if err != nil {
+				logger.Error(err, "Failed to annotate VolumeGroupReplication owner")
 
-			return reconcile.Result{}, err
+				return ctrl.Result{}, err
+			}
+
+			if err = r.addFinalizerToVGR(logger, vgr); err != nil {
```
Nikhil-Ladha, line 252: Also, why are we adding VGR finalizer here?
Shouldn't the VGR controller take care of this?

controllers/replication.storage/volumereplication_controller.go
Madhu-1: This is a safeguard to ensure that VR won't get deleted before deleting the VGR (it might be an extra check to make independent decision on VR controller not to depend on the VGR workflow)

controllers/replication.storage/volumereplication_controller.go
Madhu-1: As I mentioned above this is a safe guard check to ensure that the whole flow is followed before deleting the VR as we need to have a VGR object.

controllers/replication.storage/volumereplication_controller.go
Nikhil-Ladha: Ok, makes sense

Rakshith-R (dismissed):
Thanks

Madhu-1:
@iPraveenParihar You have requested changes on this one, can you please review it again.
@nixpanic can you please also review it again, made all the changes

Nikhil-Ladha (commented):
lgtm

cmd/csi-addons/main.go
```diff
@@ -58,6 +59,7 @@ func init() {
 	flag.StringVar(&cmd.stagingPath, "stagingpath", stagingPath, "staging path")
 	flag.StringVar(&cmd.operation, "operation", "", "csi-addons operation")
 	flag.StringVar(&cmd.persistentVolume, "persistentvolume", "", "name of the PersistentVolume")
+	flag.StringVar(&cmd.volumeGroupReplicationContent, "volumegroupreplicaitoncontent", "", "name of the VolumeGroupReplicationContent")
```
nixpanic, line 62: typo: `volumegroupreplicaitoncontent`

cmd/csi-addons/replication.go
```diff
@@ -59,28 +60,67 @@ func (rep *VolumeReplicationBase) Init(c *command) error {
...
+		return errors.New("only one of volumeID or groupID should be set")
+	case v.volumeID != "":
+		req.Type = &proto.ReplicationSource_Volume{
+			Volume: &proto.ReplicationSource_VolumeSource{
+				VolumeId: v.volumeID,
+			},
+		}
+		return nil
+	case v.groupID != "":
+		req = &proto.ReplicationSource{
```
nixpanic, line 112: this does not look correct, you don't modify the passed `req` here.

controllers/replication.storage/replication/replication.go
```diff
@@ -38,16 +40,31 @@ type Response struct {
...
 	VolumeID        string
+	GroupID         string
 	ReplicationID   string
 	Parameters      map[string]string
 	SecretName      string
 	SecretNamespace string
 	Replication     client.VolumeReplication
 }
 
+func (r *Replication) getID() (string, error) {
```
nixpanic, line 51: Is it not needed to check if both are set?

internal/client/group-replication-client.go
```diff
@@ -0,0 +1,181 @@
+/*
+Copyright 2024 The Kubernetes-CSI-Addons Authors.
```
nixpanic, line 2: the filename is a little weird, it is in the `client` package, and also ends in `-client.go`.

internal/client/volume-replication-client.go
```diff
@@ -0,0 +1,181 @@
+/*
+Copyright 2022 The Kubernetes-CSI-Addons Authors.
```
nixpanic, line 2: correct the year?
Also has the unneeded `-client` in the filename.

internal/sidecar/service/volumereplication.go
```diff
@@ -290,3 +257,28 @@ func (rs *ReplicationServer) GetVolumeReplicationInfo(
...
+// setReplicationSource sets the replication source for the given ReplicationSource.
+func setReplicationSource(src *csiReplication.ReplicationSource, req *proto.ReplicationSource) error {
+	if src == nil {
+		src = &csiReplication.ReplicationSource{}
+	}
+
+	switch {
+	case req == nil:
+		return errors.New("replication source is required")
+	case req.GetVolume() == nil && req.GetVolumegroup() == nil:
```
nixpanic, line 270: it looks like we missed something somewhere else, `GetVolumegroup()` should have been `GetVolumeGroup()`

nixpanic (changes requested):
at least one comment needs careful checking, the others are mostly nits

cmd/csi-addons/replication.go
Madhu-1: @nixpanic This is the replication Spec which need to be set nothing else is modified here.

internal/sidecar/service/volumereplication.go
Madhu-1: it was a typo in proto file itself, its updated to use volume_group instead of volumegroup

cmd/csi-addons/replication.go
nixpanic: But is it intended to overwrite the `req` variable this way? The new `req` is not returned, so the caller of the function will never see it. I expect a warning from golangci about this too.

The original (`req.Type`) is not updated as is done on line 105.

Madhu-1: Sorry, its an overlook from my side, you are correct we should not overwrite the req rather set the .Type as its done in other place.

internal/client/group-replication.go
```diff
@@ -0,0 +1,181 @@
...
+
+	"github.com/csi-addons/kubernetes-csi-addons/internal/proto"
+
+	"google.golang.org/grpc"
+)
+
+type volumeGroupReplicationClient struct {
+	client  proto.ReplicationClient
+	timeout time.Duration
+}
```
nixpanic, line 31: consider adding a type assertion get clear errors:

```go
var _ VolumeReplication = &volumeGroupReplicationClient{}
```

nixpanic (commented):
had a quick look, seems reasonable to me. No objections, but I haven't checked the reconciling code in detail

.github/workflows/lint-extras.yaml
```diff
@@ -22,7 +22,7 @@ jobs:
           fetch-depth: 0
 
       - name: Lint Code Base
-        uses: super-linter/super-linter/slim@v6
+        uses: super-linter/super-linter/slim@v6.6.0
```
Madhu-1, line 25: opened tracker for https://github.com/csi-addons/kubernetes-csi-addons/issues/609 to swith to next release.

cmd/csi-addons/replication.go
```diff
@@ -59,28 +60,65 @@ func (rep *VolumeReplicationBase) Init(c *command) error {
...
-		return fmt.Errorf("failed to get pv %q", c.persistentVolume)
-	}
-
-	if pv.Spec.CSI == nil {
-		return fmt.Errorf("pv %q is not a CSI volume", c.persistentVolume)
-	}
-
-	if pv.Spec.CSI.VolumeHandle == "" {
-		return errors.New("volume ID is not set")
+	if c.persistentVolume != "" {
```
iPraveenParihar, line 63: In case where both PV and VGRC are provided, PV will be considered.
error out before this if both are provided.

Rakshith-R (approved):
LGTM

iPraveenParihar (commented):
LGTM, will leave it to @nixpanic for final review and approval.
Thanks!

nixpanic (approved):

iPraveenParihar (approved):

PR #610, opened 2024-07-08
Add volume group replication controller code

Nikhil-Ladha:
This PR adds the following codes/logic:
 
- Added volume group replication controller logic
- Added generated crds, rbacs
- Added docs to create VGR, VGRClass and VGRContent CRs

### DIFF
--- a/api/replication.storage/v1alpha1/volumegroupreplication_types.go
+++ b/api/replication.storage/v1alpha1/volumegroupreplication_types.go
@@ -21,6 +21,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	VolumeGroupReplicationNameAnnotation = "replication.storage.openshift.io/volume-group-replication-name"
+)
+
 // VolumeGroupReplicationSpec defines the desired state of VolumeGroupReplication
 type VolumeGroupReplicationSpec struct {
 	// volumeGroupReplicationClassName is the volumeGroupReplicationClass name for this VolumeGroupReplication resource
--- a/api/replication.storage/v1alpha1/volumegroupreplicationcontent_types.go
+++ b/api/replication.storage/v1alpha1/volumegroupreplicationcontent_types.go
@@ -21,6 +21,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	VolumeGroupReplicationContentNameAnnotation = "replication.storage.openshift.io/volume-group-replication-content-name"
+)
+
 // VolumeGroupReplicationContentSpec defines the desired state of VolumeGroupReplicationContent
 type VolumeGroupReplicationContentSpec struct {
 	// VolumeGroupreplicationRef specifies the VolumeGroupReplication object to which this
--- a/internal/client/volumegroup-client.go
+++ b/internal/client/volumegroup-client.go
@@ -35,7 +35,7 @@ type VolumeGroup interface {
 	// CreateVolumeGroup RPC call to create a volume group.
 	CreateVolumeGroup(volumeGroupName string, volumeIDs []string, secretName, secretNamespace string, parameters map[string]string) (*proto.CreateVolumeGroupResponse, error)
 	// ModifyVolumeGroupMembership RPC call to modify the volume group.
-	ModifyVolumeGroupMembership(volumeGroupID string, volumeIDs []string, secretName, secretNamespace string) (*proto.ModifyVolumeGroupMembershipResponse, error)
+	ModifyVolumeGroupMembership(volumeGroupID string, volumeIDs []string, secretName, secretNamespace string, parameters map[string]string) (*proto.ModifyVolumeGroupMembershipResponse, error)
 	// DeleteVolumeGroup RPC call to delete the volume group.
 	DeleteVolumeGroup(volumeGroupID string, secretName, secretNamespace string) (*proto.DeleteVolumeGroupResponse, error)
 	// ControllerGetVolumeGroup RPC call to fetch the volume group.
@@ -64,12 +64,13 @@ func (vg *volumeGroupClient) CreateVolumeGroup(volumeGroupName string, volumeIDs
 }
 
 func (vg *volumeGroupClient) ModifyVolumeGroupMembership(volumeGroupID string, volumeIDs []string,
-	secretName, secretNamespace string) (*proto.ModifyVolumeGroupMembershipResponse, error) {
+	secretName, secretNamespace string, parameters map[string]string) (*proto.ModifyVolumeGroupMembershipResponse, error) {
 	req := &proto.ModifyVolumeGroupMembershipRequest{
 		VolumeGroupId:   volumeGroupID,
 		VolumeIds:       volumeIDs,
 		SecretName:      secretName,
 		SecretNamespace: secretNamespace,
+		Parameters:      parameters,
 	}
 
 	createCtx, cancel := context.WithTimeout(context.Background(), vg.timeout)
--- a/internal/controller/replication.storage/finalizers.go
+++ b/internal/controller/replication.storage/finalizers.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
@@ -65,43 +66,44 @@ func (r *VolumeReplicationReconciler) removeFinalizerFromVR(logger logr.Logger, 
 	return nil
 }
 
-// addFinalizerToPVC adds the VR finalizer on the PersistentVolumeClaim.
-func (r *VolumeReplicationReconciler) addFinalizerToPVC(logger logr.Logger, pvc *corev1.PersistentVolumeClaim) error {
-	if !slices.Contains(pvc.Finalizers, pvcReplicationFinalizer) {
-		logger.Info("adding finalizer to PersistentVolumeClaim object", "Finalizer", pvcReplicationFinalizer)
-		pvc.Finalizers = append(pvc.Finalizers, pvcReplicationFinalizer)
-		if err := r.Update(context.TODO(), pvc); err != nil {
+// addFinalizerToPVC adds the VR, VGR finalizer on the PersistentVolumeClaim.
+func addFinalizerToPVC(client client.Client, logger logr.Logger, pvc *corev1.PersistentVolumeClaim,
+	replicationFinalizer string) error {
+	if !slices.Contains(pvc.Finalizers, replicationFinalizer) {
+		logger.Info("adding finalizer to PersistentVolumeClaim object", "Finalizer", replicationFinalizer)
+		pvc.Finalizers = append(pvc.Finalizers, replicationFinalizer)
+		if err := client.Update(context.TODO(), pvc); err != nil {
 			return fmt.Errorf("failed to add finalizer (%s) to PersistentVolumeClaim resource"+
 				" (%s/%s) %w",
-				pvcReplicationFinalizer, pvc.Namespace, pvc.Name, err)
+				replicationFinalizer, pvc.Namespace, pvc.Name, err)
 		}
 	}
 
 	return nil
 }
 
-// removeFinalizerFromPVC removes the VR finalizer on PersistentVolumeClaim.
-func (r *VolumeReplicationReconciler) removeFinalizerFromPVC(logger logr.Logger, pvc *corev1.PersistentVolumeClaim,
-) error {
-	if slices.Contains(pvc.Finalizers, pvcReplicationFinalizer) {
-		logger.Info("removing finalizer from PersistentVolumeClaim object", "Finalizer", pvcReplicationFinalizer)
-		pvc.Finalizers = util.RemoveFromSlice(pvc.Finalizers, pvcReplicationFinalizer)
-		if err := r.Update(context.TODO(), pvc); err != nil {
+// removeFinalizerFromPVC removes the VR, VGR finalizer on PersistentVolumeClaim.
+func removeFinalizerFromPVC(client client.Client, logger logr.Logger, pvc *corev1.PersistentVolumeClaim,
+	replicationFinalizer string) error {
+	if slices.Contains(pvc.Finalizers, replicationFinalizer) {
+		logger.Info("removing finalizer from PersistentVolumeClaim object", "Finalizer", replicationFinalizer)
+		pvc.Finalizers = util.RemoveFromSlice(pvc.Finalizers, replicationFinalizer)
+		if err := client.Update(context.TODO(), pvc); err != nil {
 			return fmt.Errorf("failed to remove finalizer (%s) from PersistentVolumeClaim resource"+
 				" (%s/%s), %w",
-				pvcReplicationFinalizer, pvc.Namespace, pvc.Name, err)
+				replicationFinalizer, pvc.Namespace, pvc.Name, err)
 		}
 	}
 
 	return nil
 }
 
-// addFinalizerToVGR adds the VR finalizer on the VolumeGroupReplication.
-func (r *VolumeReplicationReconciler) addFinalizerToVGR(logger logr.Logger, vgr *replicationv1alpha1.VolumeGroupReplication) error {
+// addFinalizerToVGR adds the VGR finalizer on the VolumeGroupReplication resource
+func addFinalizerToVGR(client client.Client, logger logr.Logger, vgr *replicationv1alpha1.VolumeGroupReplication) error {
 	if !slices.Contains(vgr.Finalizers, vgrReplicationFinalizer) {
-		logger.Info("adding finalizer to VolumeGroupReplication object", "Finalizer", vgrReplicationFinalizer)
+		logger.Info("adding finalizer to volumeGroupReplication object", "Finalizer", vgrReplicationFinalizer)
 		vgr.Finalizers = append(vgr.Finalizers, vgrReplicationFinalizer)
-		if err := r.Update(context.TODO(), vgr); err != nil {
+		if err := client.Update(context.TODO(), vgr); err != nil {
 			return fmt.Errorf("failed to add finalizer (%s) to VolumeGroupReplication resource"+
 				" (%s/%s) %w",
 				vgrReplicationFinalizer, vgr.Namespace, vgr.Name, err)
@@ -111,16 +113,53 @@ func (r *VolumeReplicationReconciler) addFinalizerToVGR(logger logr.Logger, vgr 
 	return nil
 }
 
-// removeFinalizerFromVGR removes the VR finalizer on VolumeGroupReplication.
-func (r *VolumeReplicationReconciler) removeFinalizerFromVGR(logger logr.Logger, vgr *replicationv1alpha1.VolumeGroupReplication,
-) error {
+// removeFinalizerFromVGR removes the VGR finalizer from the VolumeGroupReplication instance.
+func removeFinalizerFromVGR(client client.Client, logger logr.Logger, vgr *replicationv1alpha1.VolumeGroupReplication) error {
 	if slices.Contains(vgr.Finalizers, vgrReplicationFinalizer) {
-		logger.Info("removing finalizer from VolumeGroupReplication object", "Finalizer", vgrReplicationFinalizer)
+		logger.Info("removing finalizer from volumeGroupReplication object", "Finalizer", vgrReplicationFinalizer)
+		// Check if owner annotations are removed from the VGR resource
+		if vgr.Annotations[replicationv1alpha1.VolumeGroupReplicationContentNameAnnotation] != "" ||
+			vgr.Annotations[replicationv1alpha1.VolumeReplicationNameAnnotation] != "" {
+			return fmt.Errorf("failed to remove finalizer from volumeGroupReplication object"+
+				",dependent resources are not yet deleted (%s/%s)", vgr.Namespace, vgr.Name)
+		}
 		vgr.Finalizers = util.RemoveFromSlice(vgr.Finalizers, vgrReplicationFinalizer)
-		if err := r.Update(context.TODO(), vgr); err != nil {
+		if err := client.Update(context.TODO(), vgr); err != nil {
 			return fmt.Errorf("failed to remove finalizer (%s) from VolumeGroupReplication resource"+
 				" (%s/%s), %w",
 				vgrReplicationFinalizer, vgr.Namespace, vgr.Name, err)
+		}
+	}
+
+	return nil
+}
+
+// addFinalizerToVGRContent adds the VGR finalizer on the VolumeGroupReplicationContent resource
+func addFinalizerToVGRContent(client client.Client, logger logr.Logger, vgrContent *replicationv1alpha1.VolumeGroupReplicationContent,
+	replicationFinalizer string) error {
+	if !slices.Contains(vgrContent.Finalizers, replicationFinalizer) {
+		logger.Info("adding finalizer to volumeGroupReplicationContent object", "Finalizer", replicationFinalizer)
+		vgrContent.Finalizers = append(vgrContent.Finalizers, replicationFinalizer)
+		if err := client.Update(context.TODO(), vgrContent); err != nil {
+			return fmt.Errorf("failed to add finalizer (%s) to VolumeGroupReplicationContent resource"+
+				" (%s) %w",
+				replicationFinalizer, vgrContent.Name, err)
+		}
+	}
+
+	return nil
+}
+
+// removeFinalizerFromVGRContent removes the VGR finalizer from the VolumeGroupReplicationContent instance.
+func removeFinalizerFromVGRContent(client client.Client, logger logr.Logger, vgrContent *replicationv1alpha1.VolumeGroupReplicationContent,
+	replicationFinalizer string) error {
+	if slices.Contains(vgrContent.Finalizers, replicationFinalizer) {
+		logger.Info("removing finalizer from volumeGroupReplicationContent object", "Finalizer", replicationFinalizer)
+		vgrContent.Finalizers = util.RemoveFromSlice(vgrContent.Finalizers, replicationFinalizer)
+		if err := client.Update(context.TODO(), vgrContent); err != nil {
+			return fmt.Errorf("failed to remove finalizer (%s) from VolumeGroupReplicationContent resource"+
+				" (%s), %w",
+				replicationFinalizer, vgrContent.Name, err)
 		}
 	}
 
--- a/internal/controller/replication.storage/parameters.go
+++ b/internal/controller/replication.storage/parameters.go
@@ -29,8 +29,10 @@ const (
 	// Driver.
 	replicationParameterPrefix = "replication.storage.openshift.io/"
 
-	prefixedReplicationSecretNameKey      = replicationParameterPrefix + "replication-secret-name"      // name key for secret
-	prefixedReplicationSecretNamespaceKey = replicationParameterPrefix + "replication-secret-namespace" // namespace key secret
+	prefixedReplicationSecretNameKey           = replicationParameterPrefix + "replication-secret-name"            // name key for secret
+	prefixedReplicationSecretNamespaceKey      = replicationParameterPrefix + "replication-secret-namespace"       // namespace key secret
+	prefixedGroupReplicationSecretNameKey      = replicationParameterPrefix + "group-replication-secret-name"      // name key for secret
+	prefixedGroupReplicationSecretNamespaceKey = replicationParameterPrefix + "group-replication-secret-namespace" // namespace key secret
 )
 
 // filterPrefixedParameters removes all the reserved keys from the
@@ -53,10 +55,14 @@ func validatePrefixedParameters(param map[string]string) error {
 		if strings.HasPrefix(k, replicationParameterPrefix) {
 			switch k {
 			case prefixedReplicationSecretNameKey:
+				fallthrough
+			case prefixedGroupReplicationSecretNameKey:
 				if v == "" {
 					return errors.New("secret name cannot be empty")
 				}
 			case prefixedReplicationSecretNamespaceKey:
+				fallthrough
+			case prefixedGroupReplicationSecretNamespaceKey:
 				if v == "" {
 					return errors.New("secret namespace cannot be empty")
 				}
--- a/internal/controller/replication.storage/pvc.go
+++ b/internal/controller/replication.storage/pvc.go
@@ -24,14 +24,15 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	replicationv1alpha1 "github.com/csi-addons/kubernetes-csi-addons/api/replication.storage/v1alpha1"
 )
 
 // getPVCDataSource get pvc, pv object from the request.
-func (r VolumeReplicationReconciler) getPVCDataSource(logger logr.Logger, req types.NamespacedName) (*corev1.PersistentVolumeClaim, *corev1.PersistentVolume, error) {
+func (r VolumeReplicationReconciler) getPVCDataSource(ctx context.Context, logger logr.Logger, req types.NamespacedName) (*corev1.PersistentVolumeClaim, *corev1.PersistentVolume, error) {
 	pvc := &corev1.PersistentVolumeClaim{}
-	err := r.Get(context.TODO(), req, pvc)
+	err := r.Get(ctx, req, pvc)
 	if err != nil {
 		if errors.IsNotFound(err) {
 			logger.Error(err, "PVC not found", "PVC Name", req.Name)
@@ -59,22 +60,27 @@ func (r VolumeReplicationReconciler) getPVCDataSource(logger logr.Logger, req ty
 	return pvc, pv, nil
 }
 
-// annotatePVCWithOwner will add the VolumeReplication details to the PVC annotations.
-func (r *VolumeReplicationReconciler) annotatePVCWithOwner(ctx context.Context, logger logr.Logger, reqOwnerName string, pvc *corev1.PersistentVolumeClaim) error {
+// annotatePVCWithOwner will add the VolumeReplication/VolumeGroupReplication details to the PVC annotations.
+func annotatePVCWithOwner(client client.Client, ctx context.Context, logger logr.Logger, reqOwnerName string,
+	pvc *corev1.PersistentVolumeClaim, pvcAnnotation string) error {
 	if pvc.Annotations == nil {
 		pvc.Annotations = map[string]string{}
 	}
 
-	currentOwnerName := pvc.Annotations[replicationv1alpha1.VolumeReplicationNameAnnotation]
-	if currentOwnerName == "" {
-		logger.Info("setting owner on PVC annotation", "Name", pvc.Name, "owner", reqOwnerName)
-		pvc.Annotations[replicationv1alpha1.VolumeReplicationNameAnnotation] = reqOwnerName
-		err := r.Update(ctx, pvc)
-		if err != nil {
-			logger.Error(err, "Failed to update PVC annotation", "Name", pvc.Name)
+	if pvc.Annotations[replicationv1alpha1.VolumeReplicationNameAnnotation] != "" &&
+		pvc.Annotations[replicationv1alpha1.VolumeGroupReplicationNameAnnotation] != "" {
+		logger.Info("PVC can't be part of both VolumeGroupReplication and VolumeReplication")
+		return fmt.Errorf("PVC (%s/%s) can't be owned by both VolumeReplication and VolumeGroupReplication", pvc.Namespace, pvc.Name)
+	}
 
-			return fmt.Errorf("failed to update PVC %q annotation for VolumeReplication: %w",
-				pvc.Name, err)
+	currentOwnerName := pvc.Annotations[pvcAnnotation]
+	if currentOwnerName == "" {
+		logger.Info("setting owner on PVC annotation", "PVC Name", pvc.Name, "PVC Namespace", pvc.Namespace, "owner", reqOwnerName)
+		pvc.Annotations[pvcAnnotation] = reqOwnerName
+		err := client.Update(ctx, pvc)
+		if err != nil {
+			logger.Error(err, "Failed to update PVC annotation", "Name", pvc.Name, "Namespace", pvc.Namespace)
+			return fmt.Errorf("failed to update PVC (%s/%s) annotation for replication: %w", pvc.Namespace, pvc.Name, err)
 		}
 
 		return nil
@@ -83,25 +89,27 @@ func (r *VolumeReplicationReconciler) annotatePVCWithOwner(ctx context.Context, 
 	if currentOwnerName != reqOwnerName {
 		logger.Info("cannot change the owner of PVC",
 			"PVC name", pvc.Name,
+			"PVC Namespace", pvc.Namespace,
 			"current owner", currentOwnerName,
 			"requested owner", reqOwnerName)
 
-		return fmt.Errorf("PVC %q not owned by VolumeReplication %q",
-			pvc.Name, reqOwnerName)
+		return fmt.Errorf("PVC (%s/%s) not owned by correct VolumeReplication/VolumeGroupReplication %q",
+			pvc.Namespace, pvc.Name, reqOwnerName)
 	}
 
 	return nil
 }
 
-// removeOwnerFromPVCAnnotation removes the VolumeReplication owner from the PVC annotations.
-func (r *VolumeReplicationReconciler) removeOwnerFromPVCAnnotation(ctx context.Context, logger logr.Logger, pvc *corev1.PersistentVolumeClaim) error {
-	if _, ok := pvc.Annotations[replicationv1alpha1.VolumeReplicationNameAnnotation]; ok {
-		logger.Info("removing owner annotation from PersistentVolumeClaim object", "Annotation", replicationv1alpha1.VolumeReplicationNameAnnotation)
-		delete(pvc.Annotations, replicationv1alpha1.VolumeReplicationNameAnnotation)
-		if err := r.Update(ctx, pvc); err != nil {
+// removeOwnerFromPVCAnnotation removes the VolumeReplication/VolumeGroupReplication owner from the PVC annotations.
+func removeOwnerFromPVCAnnotation(client client.Client, ctx context.Context, logger logr.Logger, pvc *corev1.PersistentVolumeClaim,
+	pvcAnnotation string) error {
+	if _, ok := pvc.Annotations[pvcAnnotation]; ok {
+		logger.Info("removing owner annotation from PersistentVolumeClaim object", "PVC Name", pvc.Name, "PVC Namespace", pvc.Namespace, "Annotation", pvcAnnotation)
+		delete(pvc.Annotations, pvcAnnotation)
+		if err := client.Update(ctx, pvc); err != nil {
 			return fmt.Errorf("failed to remove annotation %q from PersistentVolumeClaim "+
-				"%q %w",
-				replicationv1alpha1.VolumeReplicationNameAnnotation, pvc.Name, err)
+				"(%s/%s) %w",
+				pvcAnnotation, pvc.Namespace, pvc.Name, err)
 		}
 	}
 
--- a/internal/controller/replication.storage/utils.go
+++ b/internal/controller/replication.storage/utils.go
@@ -1,0 +1,73 @@
+/*
+Copyright 2025 The Kubernetes-CSI-Addons Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"context"
+	"time"
+
+	replicationv1alpha1 "github.com/csi-addons/kubernetes-csi-addons/api/replication.storage/v1alpha1"
+
+	"github.com/go-logr/logr"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func GetReplicationState(instanceState replicationv1alpha1.ReplicationState) replicationv1alpha1.State {
+	switch instanceState {
+	case replicationv1alpha1.Primary:
+		return replicationv1alpha1.PrimaryState
+	case replicationv1alpha1.Secondary:
+		return replicationv1alpha1.SecondaryState
+	case replicationv1alpha1.Resync:
+		return replicationv1alpha1.SecondaryState
+	}
+
+	return replicationv1alpha1.UnknownState
+}
+
+func GetCurrentReplicationState(instanceStatusState replicationv1alpha1.State) replicationv1alpha1.State {
+	if instanceStatusState == "" {
+		return replicationv1alpha1.UnknownState
+	}
+
+	return instanceStatusState
+}
+
+func WaitForVolumeReplicationResource(client client.Client, logger logr.Logger, resourceName string) error {
+	unstructuredResource := &unstructured.UnstructuredList{}
+	unstructuredResource.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   replicationv1alpha1.GroupVersion.Group,
+		Kind:    resourceName,
+		Version: replicationv1alpha1.GroupVersion.Version,
+	})
+	for {
+		err := client.List(context.TODO(), unstructuredResource)
+		if err == nil {
+			return nil
+		}
+		// return errors other than NoMatch
+		if !meta.IsNoMatchError(err) {
+			logger.Error(err, "got an unexpected error while waiting for resource", "Resource", resourceName)
+			return err
+		}
+		logger.Info("resource does not exist", "Resource", resourceName)
+		time.Sleep(5 * time.Second)
+	}
+}
--- a/internal/controller/replication.storage/volumegroupreplication_controller.go
+++ b/internal/controller/replication.storage/volumegroupreplication_controller.go
@@ -18,36 +18,771 @@ package controller
 
 import (
 	"context"
+	"errors"
+	"fmt"
+	"reflect"
+	"slices"
+	"strings"
+	"time"
 
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/selection"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	replicationv1alpha1 "github.com/csi-addons/kubernetes-csi-addons/api/replication.storage/v1alpha1"
+	"github.com/csi-addons/kubernetes-csi-addons/internal/controller/replication.storage/replication"
+)
+
+const (
+	volumeGroupReplicationClass   = "VolumeGroupReplicationClass"
+	volumeGroupReplication        = "VolumeGroupReplication"
+	volumeGroupReplicationContent = "VolumeGroupReplicationContent"
+	volumeGroupReplicationRef     = "replication.storage.openshift.io/volumegroupref"
+	pvcSelector                   = "pvcSelector"
 )
 
 // VolumeGroupReplicationReconciler reconciles a VolumeGroupReplication object
 type VolumeGroupReplicationReconciler struct {
 	client.Client
-	Scheme *runtime.Scheme
+	Scheme   *runtime.Scheme
+	ctx      context.Context
+	log      logr.Logger
+	Recorder record.EventRecorder
 }
 
-//+kubebuilder:rbac:groups=replication.storage.openshift.io,resources=volumegroupreplications,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups="",resources=events,verbs=create;patch
+//+kubebuilder:rbac:groups=replication.storage.openshift.io,resources=volumegroupreplications,verbs=get;list;watch;update;patch
 //+kubebuilder:rbac:groups=replication.storage.openshift.io,resources=volumegroupreplications/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=replication.storage.openshift.io,resources=volumegroupreplications/finalizers,verbs=update
+//+kubebuilder:rbac:groups=replication.storage.openshift.io,resources=volumegroupreplicationclasses,verbs=get;list;watch
+//+kubebuilder:rbac:groups=replication.storage.openshift.io,resources=volumegroupreplicationcontents,verbs=get;list;watch;create;update;delete
+//+kubebuilder:rbac:groups=replication.storage.openshift.io,resources=volumereplications,verbs=get;list;watch;create;update;delete
+//+kubebuilder:rbac:groups=replication.storage.openshift.io,resources=volumereplications/status,verbs=get;list
+//+kubebuilder:rbac:groups=core,resources=persistentvolumeclaims;persistentvolumes,verbs=get;list;watch;update
+//+kubebuilder:rbac:groups=core,resources=persistentvolumeclaims/finalizers,verbs=update
+
+/*
+Steps performed by the reconcile loop:
+- Fetch and validate the VGRClass CR
+- Fetch the matching PVCs based on the selector provided in the VGR CR, and check if they are already bounded to a CSI volume and the driver matches the driver provided in the VGRClass CR.
+- Annotate to the PVCs with owner and add the VGR finalizer to them.
+- Add the label selector to the VGR annotation, such that the PVC triggering a reconcile can fetch the VGR to reconcile
+- Create the VGRContent with the PVs list fetched above, add VGR name/namespace as the annotation to it
+- Wait for the volumes to be grouped, and the VGRContent to be updated with the group handle
+- Then, create the VR CR and add VGR name/namespace as the annotation to it
+- Update the VGR status with the VR status.
+
+In case of deletion:
+- Remove the owner annotations and finalizers from the PVC
+- Check if VR exists, then delete
+- Check if VGRContent exists, then delete
+- Remove VGR finalizer <- This won't happen until the dependent VR and VRContent is deleted. Validated using owner annotations set in both the dependent CRs.
+*/
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
 func (r *VolumeGroupReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	_ = log.FromContext(ctx)
+	r.log = log.FromContext(ctx, "Request.Name", req.Name, "Request.Namespace", req.Namespace)
+	r.ctx = log.IntoContext(ctx, r.log)
 
-	return ctrl.Result{}, nil
+	// Fetch VolumeGroupReplication instance
+	instance := &replicationv1alpha1.VolumeGroupReplication{}
+	err := r.Get(ctx, req.NamespacedName, instance)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			r.log.Info("volumeGroupReplication resource not found")
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, err
+	}
+
+	// Get VolumeGroupReplicationClass instance
+	vgrClassObj, err := r.getVolumeGroupReplicationClass(instance.Spec.VolumeGroupReplicationClassName)
+	if err != nil {
+		r.log.Error(err, "failed to get volumeGroupReplicationClass resource", "VGRClassName", instance.Spec.VolumeGroupReplicationClassName)
+		_ = r.setGroupReplicationFailure(instance, err)
+		return reconcile.Result{}, err
+	}
+
+	// Validate that required parameters are present in the VGRClass resource
+	err = validatePrefixedParameters(vgrClassObj.Spec.Parameters)
+	if err != nil {
+		r.log.Error(err, "failed to validate parameters of volumeGroupReplicationClass", "VGRClassName", instance.Spec.VolumeGroupReplicationClassName)
+		_ = r.setGroupReplicationFailure(instance, err)
+		return reconcile.Result{}, err
+	}
+
+	// Declare all dependent resources
+	vgrContentName := fmt.Sprintf("vgrcontent-%s", instance.UID)
+	if instance.Spec.VolumeGroupReplicationContentName != "" {
+		vgrContentName = instance.Spec.VolumeGroupReplicationContentName
+	}
+	vgrContentObj := &replicationv1alpha1.VolumeGroupReplicationContent{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: vgrContentName,
+		},
+	}
+	vrName := fmt.Sprintf("vr-%s", instance.UID)
+	if instance.Spec.VolumeReplicationName != "" {
+		vrName = instance.Spec.VolumeReplicationName
+	}
+	vrObj := &replicationv1alpha1.VolumeReplication{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      vrName,
+			Namespace: instance.Namespace,
+		},
+	}
+
+	// Create/Update dependent resources only if the instance is not marked for deletion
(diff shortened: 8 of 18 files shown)


Nikhil-Ladha:
The changes are not yet tested, will wait for #605 to get merged and then update a few things and then test it.
In the meantime, please feel free to provide some initial reviews.
Thanks :)

internal/proto/volumegroup.proto
```diff
@@ -0,0 +1,182 @@
...
+
+  // params passed to the plugin to create the volume group.
+  // This field is OPTIONAL.
+  map<string, string> parameters = 2;
+
+  // Secrets required by plugin to complete volume group creation
+  // request.
+  // This field is OPTIONAL. Refer to the `Secrets Requirements`
+  // section on how to use this field.
+  map<string, string> secrets = 3;
```
nixpanic, line 40: secrets should not be sent between controller and sidecar. The sidecar should be able to resolve the secrets based on namespace/secret-name.

internal/proto/volumegroup.proto
```diff
@@ -0,0 +1,182 @@
...
+  // CreateVolumeGroup RPC call to create a volume group.
+  rpc CreateVolumeGroup(CreateVolumeGroupRequest)
+      returns (CreateVolumeGroupResponse) {}
+  // ModifyVolumeGroupMembership RPC call to modify a volume group.
+  rpc ModifyVolumeGroupMembership(ModifyVolumeGroupMembershipRequest)
+      returns (ModifyVolumeGroupMembershipResponse) {}
+  // DeleteVolumeGroup RPC call to delete a volume group.
+  rpc DeleteVolumeGroup(DeleteVolumeGroupRequest)
+      returns (DeleteVolumeGroupResponse) {}
+  // ListVolumeGroups RPC call to list volume groups.
```
nixpanic, line 18: Lots of things commented out, just remove it all.

This .proto file is for the communication between controller and sidecar, it would (usually) be a reduced/simplified version of the CSI-Addons specification. You can use Kubernetes objects/references here, PVs instead of CSI volume_ids and so on.

nixpanic (changes requested):

controllers/replication.storage/utils.go
```diff
@@ -0,0 +1,85 @@
...
+*/
+
+package controllers
+
+import (
+	"context"
+	"time"
+
+	replicationv1alpha1 "github.com/csi-addons/kubernetes-csi-addons/apis/replication.storage/v1alpha1"
+	"github.com/go-logr/logr"
```
Madhu-1, line 24: add a new line for separation

controllers/replication.storage/volumegroupreplication_controller.go
```diff
@@ -33,21 +56,533 @@ type VolumeGroupReplicationReconciler struct {
 	Scheme *runtime.Scheme
 }
 
-//+kubebuilder:rbac:groups=replication.storage.openshift.io,resources=volumegroupreplications,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=replication.storage.openshift.io,resources=volumegroupreplications,verbs=get;list;watch;update;patch
 //+kubebuilder:rbac:groups=replication.storage.openshift.io,resources=volumegroupreplications/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=replication.storage.openshift.io,resources=volumegroupreplications/finalizers,verbs=update
+//+kubebuilder:rbac:groups=replication.storage.openshift.io,resources=volumegroupreplicationclasses,verbs=get;list
+//+kubebuilder:rbac:groups=replication.storage.openshift.io,resources=volumegroupreplicationcontents,verbs=get;list;watch;create;update;delete
+//+kubebuilder:rbac:groups=replication.storage.openshift.io,resources=volumereplications,verbs=get;list;watch;create;update;delete
```
Madhu-1, line 64: We need to add finalizers on VR as well right?

controllers/replication.storage/volumegroupreplication_controller.go
```diff
@@ -33,21 +56,533 @@ type VolumeGroupReplicationReconciler struct {
...
+				err = r.Client.Update(ctx, instance)
+				if err != nil {
+					logger.Error(err, "failed to add pvc selector annotation to VGR")
+					err := r.setGroupReplicationFailure(instance, logger, err)
+					return reconcile.Result{}, err
+				}
+			}
+		} else {
+			logger.Info("No selector provided for PVCs in volumeGroupReplication source")
+			return reconcile.Result{}, nil
```
Madhu-1, line 144: should we allow this one or make the selector mandatory?

controllers/replication.storage/volumegroupreplication_controller.go
```diff
@@ -33,21 +56,533 @@ type VolumeGroupReplicationReconciler struct {
...
+		}
+
+		// Update PersistentVolumeClaimsRefList in VGR Status
+		tmpRefList := []corev1.LocalObjectReference{}
+		for i, pvc := range pvcList.Items {
+			if i < 100 {
```
Madhu-1, line 150: add this check above if we have more than 100 PVC after getMatchingPVCsFromSource

controllers/replication.storage/volumegroupreplication_controller.go
```diff
@@ -33,21 +56,533 @@ type VolumeGroupReplicationReconciler struct {
...
+			labelSelector := ""
+			pvcList, labelSelector, err = r.getMatchingPVCsFromSource(instance, logger)
+			if err != nil {
+				logger.Error(err, "failed to get PVCs using selector")
+				err := r.setGroupReplicationFailure(instance, logger, err)
+				return reconcile.Result{}, err
+			}
+			if len(pvcList.Items) == 0 {
+				logger.Info("No matching PVCs found for the given selectors")
+				return reconcile.Result{}, nil
```
Madhu-1, line 125: update the status with error if no matching pvc's are found

controllers/replication.storage/volumegroupreplication_controller.go
```diff
@@ -33,21 +56,533 @@ type VolumeGroupReplicationReconciler struct {
...
+			}
+
```
Madhu-1, line 127: add a check to ensure that all pvc's are in bound state and all are csi pvcs

controllers/replication.storage/volumegroupreplication_controller.go
```diff
@@ -33,21 +56,533 @@ type VolumeGroupReplicationReconciler struct {
...
+					logger.Error(err, "failed to get volumeGroupReplicationContent resource", "VGRContentName", instance.Spec.VolumeGroupReplicationContentName)
+					err := r.setGroupReplicationFailure(instance, logger, err)
+					return reconcile.Result{}, err
+				}
+				if !reflect.DeepEqual(pvHandlesList, vgrContentObj.Spec.Source.VolumeHandles) {
+					vgrContentObj.Spec.Source.VolumeHandles = pvHandlesList
+					err = r.Client.Update(ctx, vgrContentObj)
+					if err != nil {
+						logger.Error(err, "failed to update volumeGroupReplicationContent resource", "VGRContentName", vgrContentObj.Name)
+						err := r.setGroupReplicationFailure(instance, logger, err)
```
Madhu-1, line 205: err will be nil here if `setGroupReplicationFailure` is successful, return actual error if update fails, revisit this in all places

controllers/replication.storage/volumegroupreplication_controller.go
```diff
@@ -33,21 +56,533 @@ type VolumeGroupReplicationReconciler struct {
...
+					logger.Error(err, "failed to update volumeGroupReplication instance", "VGRName", instance.Name)
+					err := r.setGroupReplicationFailure(instance, logger, err)
+					return reconcile.Result{}, err
+				}
+			}
+		}
+
+		if vgrContentObj.Spec.VolumeGroupReplicationHandle == "" {
+			logger.Info("Either volumegroupreplicationcontent is not yet created or it is still grouping the volumes to be replicated")
+			return reconcile.Result{}, nil
```
Madhu-1, line 253: requeue this one or else if wont get reconciled again?

controllers/replication.storage/volumegroupreplication_controller.go
```diff
@@ -33,21 +56,533 @@ type VolumeGroupReplicationReconciler struct {
...
+		} else {
+			// Create VolumeReplication CR for the current VGR CR, if doesn't exists
+			vrName := fmt.Sprintf("vr-%s", instance.UID)
+			apiGroup := "replication.storage.openshift.io"
+			vrObj = &replicationv1alpha1.VolumeReplication{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      vrName,
+					Namespace: instance.Namespace,
+					Annotations: map[string]string{
+						"replication.storage.openshift.io/volumegroupref": fmt.Sprintf("%s/%s", instance.Name, instance.Namespace),
```
Madhu-1, line 263: create a variable for `"replication.storage.openshift.io/volumegroupref`

controllers/replication.storage/volumegroupreplication_controller.go
```diff
@@ -33,21 +56,533 @@ type VolumeGroupReplicationReconciler struct {
...
+				if !reflect.DeepEqual(foundVRObj, vrObj) {
+					foundVRObj = vrObj
+					err = r.Client.Update(ctx, foundVRObj)
+					if err != nil {
+						logger.Error(err, "failed to update volumeReplication resource", "VRName", vrName)
+						err := r.setGroupReplicationFailure(instance, logger, err)
+						return reconcile.Result{}, err
+					}
+				}
+			}
```
Madhu-1, line 304: just use clientOrUpdate controllerruntime helper function. https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/controller/controllerutil/example_test.go#L41C13-L41C18

controllers/replication.storage/volumegroupreplication_controller.go
```diff
@@ -33,21 +56,533 @@ type VolumeGroupReplicationReconciler struct {
...
+				err = r.Client.Update(ctx, instance)
+				if err != nil {
+					logger.Error(err, "failed to update volumeGroupReplication instance", "VGRName", instance.Name)
+					err := r.setGroupReplicationFailure(instance, logger, err)
+					return reconcile.Result{}, err
+				}
+			}
+		}
+	} else {
+		if instance.Status.PersistentVolumeClaimsRefList != nil {
```
Madhu-1, line 318: Lets remove the VGR and VR once they are gone remove the finalizers and annotation on the PVC?

controllers/replication.storage/volumegroupreplication_controller.go
```diff
@@ -33,21 +56,533 @@ type VolumeGroupReplicationReconciler struct {
...
+		if err = AddFinalizerToVGR(r.Client, logger, instance); err != nil {
+			logger.Error(err, "failed to add VolumeGroupReplication finalizer")
+			return reconcile.Result{}, err
+		}
+
+		// Check if PVCs exist based on provided selectors
+		pvcList := corev1.PersistentVolumeClaimList{}
+		if instance.Spec.Source.Selector != nil {
+			labelSelector := ""
+			pvcList, labelSelector, err = r.getMatchingPVCsFromSource(instance, logger)
```
Madhu-1, line 117: add a check to ensure all pvc belongs to same driver/provisioner in the VolumeGroupReplicationClass.

controllers/replication.storage/volumegroupreplication_controller.go
```diff
@@ -33,21 +56,533 @@ type VolumeGroupReplicationReconciler struct {
...
+			}
+
+			// Check if the resource is present in the cluster
+			vgrObj := &replicationv1alpha1.VolumeGroupReplication{}
+			logger := log.FromContext(context)
+			err := r.Client.Get(context, types.NamespacedName{Name: vgrName, Namespace: vgrNamespace}, vgrObj)
+			if err != nil {
+				logger.Error(err, "Unable to get VolumeGroupReplication resource")
+				return []reconcile.Request{}
+			}
```
Madhu-1, line 422: in which case we need this check? lets issue reconcile if we find it because we already have this check in VGR reconciler

controllers/replication.storage/volumegroupreplication_controller.go
```diff
@@ -33,21 +56,533 @@ type VolumeGroupReplicationReconciler struct {
...
+		},
+	)
+
+	enqueueVGRForPVCRequest := handler.EnqueueRequestsFromMapFunc(
+		func(context context.Context, obj client.Object) []reconcile.Request {
+			// Check if the PVC has any labels defined
+			objLabels := obj.GetLabels()
+			if len(objLabels) == 0 {
+				return []reconcile.Request{}
```
Madhu-1, line 441: there can be a case where someone removed the label from the PVC, we need to remove that from the VGR as well.

controllers/replication.storage/volumegroupreplication_controller.go
```diff
@@ -33,21 +56,533 @@ type VolumeGroupReplicationReconciler struct {
...
+							{
+								NamespacedName: types.NamespacedName{
+									Namespace: vgr.Namespace,
+									Name:      vgr.Name,
+								},
+							},
+						}
+					}
+				}
+			}
```
Madhu-1, line 473: there should be a controller runtime util function for it check something like https://github.com/kubernetes-sigs/controller-runtime/blob/a0c9fd9d3f310f48155ce985366b21914675fbea/pkg/client/example_test.go#L265-L294

controllers/replication.storage/volumegroupreplication_test.go
```diff
@@ -0,0 +1,187 @@
+/*
+Copyright 2022 The Kubernetes-CSI-Addons Authors.
```
Madhu-1, line 2: 2024

Madhu-1 (commented):
still need to do few more rounds of review, just done it for few files

controllers/replication.storage/volumegroupreplication_controller.go
Nikhil-Ladha: That the VR controller will take care off itself, VGR controller won't be adding it.

controllers/replication.storage/volumegroupreplication_controller.go
Nikhil-Ladha: We can make it mandatory, since anyway we are setting the field as immutable.
Let me update that.

controllers/replication.storage/volumegroupreplication_controller.go
Nikhil-Ladha: We are removing the PVC owner annotation below at line: 332, and later on removing the VR as well at line: 344.

controllers/replication.storage/volumegroupreplication_controller.go
Nikhil-Ladha: We are checking the driver details in the `getPVHandles` function at line: 575, though it doesn't error out if the proviosioner is different, but will update that.
I think that should do it?

controllers/replication.storage/volumegroupreplication_controller.go
Nikhil-Ladha: We need it when there is a status update in VR, as well as when the grouping is done and the handle is updated in VGRContent.
Reconcile is being issued at the very next line: 424

controllers/replication.storage/volumegroupreplication_controller.go
Nikhil-Ladha: From what I understand from the [docs](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/handler#EnqueueRequestsFromMapFunc), the enqueue function would be ran for both the old,new objects in case of update function. So, when the request is ran for older object, the details from the pvc annotation will be fetched and the VGR reconcile request will be enqueued, and since the object is already updated in the cluster the selector won't inlcude that PVC.
Though this is all theoritical, will test it to confirm.

controllers/replication.storage/volumegroupreplication_controller.go
Nikhil-Ladha: Well, I will just update the code to fetch the VGR name from the annotation, since now we are adding the owner annotation to PVC.

Nikhil-Ladha: The thing is we can't use a helper function like the one you mentioned is because the selector is not a single field that we can compare or so, and to match labels we need to create a selector requirement as done in the getPVCFromDataSource function.

internal/proto/volumegroup.proto
```diff
@@ -0,0 +1,111 @@
...
+  // modification request.
+  string secret_name = 3;
+  string secret_namespace = 4;
+}
+// ModifyVolumeGroupMembershipResponse holds the information to
+// send when volumeGroup is successfully modified.
+message ModifyVolumeGroupMembershipResponse {
+  // Contains the identifier of the modified volume group.
+  // This field is REQUIRED.
+  string volume_group_id = 1;
```
nixpanic, line 91: this looks pretty useless? Why would the `volume_group_id` need to be returned after the contents of the group was modified?

internal/proto/volumegroup.proto
```diff
@@ -0,0 +1,111 @@
...
+  // Secrets required by the driver to complete the
+  // get request.
+  string secret_name = 2;
+  string secret_namespace = 3;
+}
+// ControllerGetVolumeGroupResponse holds the information to
+// send when volumeGroup information was successfully gathered.
+message ControllerGetVolumeGroupResponse {
+  // This field is REQUIRED
+  string volume_group_id = 1;
```
nixpanic, line 110: What is the use of this? The same `volume_group_id` was passed in the request.

controllers/replication.storage/volumegroupreplication_controller.go
Nikhil-Ladha: @Madhu-1 I was wondering if we should be issuing reconcile here, since anyway we are watching the VGRContent resource, so when the grouping is done the VGRContent would be updated with the groupreplicationhandle, and that would anyway trigger a reconcile fo the VGR resource. So, what's the benefit of requeuing here?

internal/proto/volumegroup.proto
Nikhil-Ladha: Makes sense, we can remove that. Anyway, I think the volume_group_id would be immutable once created.

internal/proto/volumegroup.proto
Nikhil-Ladha: We can remove this as well, since if the volume group is present we can check for an empty (succesfull) response and use the passed group_id.

internal/proto/volumegroup.proto
Nikhil-Ladha: The actual spec is to fetch the complete volumegroup using the group id, but since we are not using that as of now that's the reason it seems redundant for now.
We might probably update the code in future to fetch the volume group and do some assertions/use data from it, so maybe we can let this be like this for now. WDYT?

nixpanic: I think you have two options, it is up to you to decide what you like to do:

1. return the complete VolumeGroup object and inspect/verify the results
2. use an empty message, and rely on an error in case something went wrong

internal/proto/volumegroup.proto
nixpanic: If this message is empty, why call the function at all?

I guess that at one point, you want to include all volumes that are part of the volume group in some VolumeGroupReplicationContent.Status kind of struct?

Nikhil-Ladha: Well, I am already adding all the volumes that are part of the volume group to the VGRContent.Status, by comparing the volume handles passed to the spec of the same.
Here: https://github.com/csi-addons/kubernetes-csi-addons/blob/8e8830de8f338f75e20dc8a1ec665e0440f78fb4/internal/controller/replication.storage/volumegroupreplicationcontent_controller.go#L160

Nikhil-Ladha: So, probably it is better to fetch the list using the `ControllerGetVolumeGroup` method instead of what I am doing right now?

Nikhil-Ladha: Ok, I guess that's not possible since we would be saving the volume names in the status and the info returned is just the group/volume ids.
So, I will just make the function empty. And, one can just use the function to validate that a volume group exists for the provided group_id.

internal/proto/volumegroup.proto
```diff
@@ -0,0 +1,106 @@
...
+  // This field is REQUIRED.
+  string volume_group_id = 1;
+
+  // Secrets required by the driver to complete the
+  // get request.
+  string secret_name = 2;
+  string secret_namespace = 3;
+}
+// ControllerGetVolumeGroupResponse holds the information to
+// send when volumeGroup information was successfully gathered.
```
nixpanic, line 104: this comment does not make much sense anymore now

nixpanic (commented):
Can you also add documentation with an example in the `docs/` directory? That makes it easier to visualize how things are supposed to work.

internal/proto/volumegroup.proto
Nikhil-Ladha: Looking at other proto files, where we return an empty response. We are following a similar convention there as well, so I think we can let it be like this.

nixpanic: the (empty) contents of the message is fine, the comment `ControllerGetVolumeGroupResponse holds the information ..` does not, it does not hold any information at all. Similar for the `*Response` messages above.

internal/controller/replication.storage/volumegroupreplication_controller.go
```diff
@@ -33,21 +57,536 @@ type VolumeGroupReplicationReconciler struct {
...
+//+kubebuilder:rbac:groups=replication.storage.openshift.io,resources=volumereplications,verbs=get;list;watch;create;update;delete
+//+kubebuilder:rbac:groups=replication.storage.openshift.io,resources=volumereplications/status,verbs=get;list
+//+kubebuilder:rbac:groups=core,resources=persistentvolumeclaims;persistentvolumes,verbs=get;list;watch;update
+//+kubebuilder:rbac:groups=core,resources=persistentvolumeclaims/finalizers,verbs=update
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
 func (r *VolumeGroupReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	_ = log.FromContext(ctx)
+	logger := log.FromContext(ctx, "Request.Name", req.Name, "Request.Namespace", req.Namespace)
```
nixpanic, line 73: This function is now so large, it is difficult to see what it does. Can you document the steps that are taken, and split some of those tasks out in helper functions?

internal/controller/replication.storage/volumegroupreplication_controller.go
```diff
@@ -33,21 +57,536 @@ type VolumeGroupReplicationReconciler struct {
...
+		logger.Error(err, "failed to update volumeGroupReplication instance's status", "VGRName", instance.Name)
+		return reconcile.Result{}, err
+	}
+
+	return reconcile.Result{}, nil
 }
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *VolumeGroupReplicationReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	err := r.waitForGroupCrds()
```
nixpanic, line 359: Please include a description of what all needs to be watched for in the comment of the function. If it is commented, there is at least some verification possible that everything was implemented.

internal/controller/replication.storage/volumegroupreplicationcontent_controller.go
```diff
@@ -18,36 +18,252 @@ package controller
...
 //+kubebuilder:rbac:groups=replication.storage.openshift.io,resources=volumegroupreplicationcontents,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=replication.storage.openshift.io,resources=volumegroupreplicationcontents/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=replication.storage.openshift.io,resources=volumegroupreplicationcontents/finalizers,verbs=update
+//+kubebuilder:rbac:groups=replication.storage.openshift.io,resources=volumegroupreplicationclasses,verbs=get;list;watch
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
 func (r *VolumeGroupReplicationContentReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	_ = log.FromContext(ctx)
+	logger := log.FromContext(ctx, "Request.Name", req.Name, "Request.Namespace", req.Namespace)
```
nixpanic, line 59: comments on this function would help a lot too. Possibly split some steps/tasks into helper functions

internal/controller/replication.storage/volumegroupreplication_controller.go
Nikhil-Ladha: Almost all the functions are split into helper functions, it seems big becuase there are few validations and CRs that are created in the Reconcile loop.

Here's the steps that are performed:
- Fetch and validate the VGRClass CR
- Fetch the matching PVCs based on the selector provided in the VGR CR, and check if they are already bounded to a CSI volume and the driver matches the driver provided in the VGRClass CR.
- Annotate to the PVCs with owner and add the VGR finalizer to them.
- Add the label selector to the VGR annotation, such that the PVC triggering a reconcile can fetch the VGR to reconcile
- Create the VGRContent with the PVs list fetched above, add VGR name/namespace as the annotation to it
- Wait for the volumes to be grouped, and the VGRContent to be updated with the group handle
- Then, create the VR CR and add VGR name/namespace as the annotation to it
- Update the VGR status with the VR status.

In case of deletion:
- Remove the owner annotations and finalizers from the PVC
- Check if VR exists, then delete
- Check if VGRContent exists, then delete
- Remove VGR finalizer <- This won't happen until the dependent VR and VRContent is deleted. Validated using owner annotations set in both the dependent CRs.

That's all that is being done.

internal/controller/replication.storage/volumegroupreplicationcontent_controller.go
Nikhil-Ladha: Almost everything is already split into helper functions, will add comments wherever missing.

internal/controller/replication.storage/volumegroupreplication_controller.go
Nikhil-Ladha: I have a doc as well, that have these steps documented: https://docs.google.com/document/d/1fOQPSpxhnQcW0xb0LUaUXHJv707VYqd7QXEcWFN5m5w/edit?usp=sharing

internal/controller/replication.storage/volumegroupreplication_controller.go
Nikhil-Ladha: Added the comments at all the places, wherever applicable.

internal/client/volumegroup-client.go
```diff
@@ -0,0 +1,112 @@
...
+		Name:            volumeGroupName,
+		VolumeIds:       volumeIDs,
+		Parameters:      parameters,
+		SecretName:      secretName,
+		SecretNamespace: secretNamespace,
+	}
+
+	createCtx, cancel := context.WithTimeout(context.Background(), vg.timeout)
+	defer cancel()
+	resp, err := vg.client.CreateVolumeGroup(createCtx, req)
```
nixpanic, line 63: you could also just

```go
return vg.client.CreateVolumeGroup(createCtx, req)
```

Also in other functions below.

internal/sidecar/service/volumegroup.go
```diff
@@ -0,0 +1,158 @@
...
+	data, err := kube.GetSecret(ctx, vg.kubeClient, req.GetSecretName(), req.GetSecretNamespace())
+	if err != nil {
+		klog.Errorf("Failed to get secret %s in namespace %s: %v", req.GetSecretName(), req.GetSecretNamespace(), err)
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	// Get the volumeGroup Id and volumeIds from the request
+	vgReq := &csiVolumeGroup.ModifyVolumeGroupMembershipRequest{
+		VolumeGroupId: req.GetVolumeGroupId(),
+		VolumeIds:     req.GetVolumeIds(),
+		Secrets:       data,
```
nixpanic, line 98: pass parameters too?

api/replication.storage/v1alpha1/volumegroupreplicationcontent_types.go
```diff
@@ -52,9 +56,8 @@ type VolumeGroupReplicationContentSpec struct {
 	// +optional
 	VolumeGroupReplicationClassName string `json:"volumeGroupReplicationClassName"`
 
-	// Source specifies whether the snapshot is (or should be) dynamically provisioned
+	// Source specifies whether the volume is (or should be) dynamically provisioned
```
nixpanic, line 59: I guess that should be "volume group"?

internal/controller/replication.storage/parameters.go
```diff
@@ -53,10 +55,12 @@ func validatePrefixedParameters(param map[string]string) error {
 		if strings.HasPrefix(k, replicationParameterPrefix) {
 			switch k {
 			case prefixedReplicationSecretNameKey:
+			case prefixedGroupReplicationSecretNameKey:
```
nixpanic, line 58: I think you need to do

```go
			switch k {
			case prefixedReplicationSecretNameKey:
				fallthrough
			case prefixedGroupReplicationSecretNameKey:
				if v == "" {
...
```

internal/controller/replication.storage/volumegroupreplication_controller.go
nixpanic: It would be great to have the list of steps included in the description of the function. When someone looks at this function in 3 months time, nobody will know that it is explained in this PR or some google doc.

api/replication.storage/v1alpha1/volumegroupreplicationcontent_types.go
Nikhil-Ladha: No, becuase we are storing the volume handles in the source of the VGRContent. So, I think it should be "volumes" only.

internal/controller/replication.storage/parameters.go
Nikhil-Ladha: Interestingly, when I tested my changes it was working fine without the `fallthrough` keyword 🤔

internal/controller/replication.storage/volumegroupreplication_controller.go
Nikhil-Ladha: Added the steps as a comment above the the `Reconcile` function definition.

internal/proto/volumegroup.proto
```diff
@@ -0,0 +1,106 @@
...
+  string secret_name = 2;
+  string secret_namespace = 3;
+}
+// DeleteVolumeGroupResponse is the response to send when
+// volumeGroup is successfully deleted.
+message DeleteVolumeGroupResponse {
+}
+// ModifyVolumeGroupMembershipRequest holds the required
+// information to modify a volume group
+message ModifyVolumeGroupMembershipRequest {
```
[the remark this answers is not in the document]
Nikhil-Ladha: Yes, updated it now.

internal/client/volumegroup-client.go
Nikhil-Ladha: Done

internal/sidecar/service/volumegroup.go
Nikhil-Ladha: Done

internal/controller/replication.storage/parameters.go
Nikhil-Ladha: Okay, I just check on a code locally, and seems like it is needed. Not sure, why didn't it compain during the testing.
Added it now.

api/replication.storage/v1alpha1/volumegroupreplicationcontent_types.go
nixpanic: Ah, righ. Can you include an example of a VGRContent in the docs?

Nikhil-Ladha: Sure, somehow I missed it _(though I had it in the back of my mind that I need to add that)_

Nikhil-Ladha:
> Can you include some testing results in a PR-comment?

Sure, I can add the creation of the CRs and the group as the comment.
The grouping still fails becuase the backend is still not ready, waiting for https://github.com/ceph/ceph-csi/pull/4739 to be merged and the omap generation process to be complete.

internal/controller/replication.storage/volumegroupreplication_controller.go
```diff
@@ -18,13 +18,37 @@ package controller
 
 import (
 	"context"
+	"fmt"
+	"reflect"
+	"strings"
 
+	"github.com/csi-addons/kubernetes-csi-addons/internal/controller/replication.storage/replication"
```
Rakshith-R, line 25: reorganize imports, internal pkgs go last

internal/controller/replication.storage/volumegroupreplication_controller.go
```diff
@@ -33,21 +57,568 @@ type VolumeGroupReplicationReconciler struct {
...
+- Fetch and validate the VGRClass CR
+- Fetch the matching PVCs based on the selector provided in the VGR CR, and check if they are already bounded to a CSI volume and the driver matches the driver provided in the VGRClass CR.
+- Annotate to the PVCs with owner and add the VGR finalizer to them.
+- Add the label selector to the VGR annotation, such that the PVC triggering a reconcile can fetch the VGR to reconcile
+- Create the VGRContent with the PVs list fetched above, add VGR name/namespace as the annotation to it
+- Wait for the volumes to be grouped, and the VGRContent to be updated with the group handle
+- Then, create the VR CR and add VGR name/namespace as the annotation to it
+- Update the VGR status with the VR status.
+
+In case of deletion:
```
Rakshith-R, line 81: Handle deletion case of object first.
maybe define an instance object like https://github.com/csi-addons/kubernetes-csi-addons/blob/main/internal/controller/replication.storage/volumereplication_controller.go#L616
and then write  helper functions with it ? 

The reconcile function  is >250 lines.

Nikhil-Ladha: > Handle deletion case of object first.

I feel, if we handle deletion first then it becomes a bit arbritary while reading the code to understand what exactly is happening and why. Having it in the end tells the reader that these all operations happened when the CR was created and that's why all these cleanups are needed.

> maybe define an instance object like https://github.com/csi-addons/kubernetes-csi-addons/blob/main/internal/controller/replication.storage/volumereplication_controller.go#L616
and then write helper functions with it ?

I don't see any use case of defining such an instance object here, the only helper function here is the `getPVCFromDataSource` and it doesn't makes sense to add this extra type for a single function.

> The reconcile function is >250 lines.

It's a bit long I agree, but is there any rules defined for the lenght of the reconcile function?
Secondly, the operations that are happening in the reconcile function are very basic and can't be broken into individual helper functions. Those, that can be broken are already done. If you have any suggestions for it, I would be happy to know.

Nikhil-Ladha: Updated the code to move the creation of CRs into their individual functions.
Now it's < 250 :)

Nikhil-Ladha:
> Looks very promising to me, thanks!
> 
> Can you include some testing results in a PR-comment?

Here's the result of creating the VGR:
```
apiVersion: v1
items:
- apiVersion: replication.storage.openshift.io/v1alpha1
  kind: VolumeGroupReplication
  metadata:
    annotations:
      pvcSelector: group=replication
      replication.storage.openshift.io/volume-replication-name: vr-c4e68900-e0c5-465b-802a-c6fd011ff40c
      replication.storage.openshift.io/volumegroupreplication-content-name: vgrcontent-c4e68900-e0c5-465b-802a-c6fd011ff40c
    creationTimestamp: "2024-08-08T05:53:25Z"
    finalizers:
    - replication.storage.openshift.io/vgr-protection
    generation: 3
    labels:
      app.kubernetes.io/created-by: kubernetes-csi-addons
      app.kubernetes.io/instance: volumegroupreplication-sample
      app.kubernetes.io/managed-by: kustomize
      app.kubernetes.io/name: volumegroupreplication
      app.kubernetes.io/part-of: kubernetes-csi-addons
    name: volumegroupreplication-sample
    namespace: rook-ceph
    resourceVersion: "193975"
    uid: c4e68900-e0c5-465b-802a-c6fd011ff40c
  spec:
    autoResync: false
    replicationState: primary
    source:
      selector:
        matchLabels:
          group: replication
    volumeGroupReplicationClassName: volumegroupreplicationclass-sample
    volumeGroupReplicationContentName: vgrcontent-c4e68900-e0c5-465b-802a-c6fd011ff40c
    volumeReplicationClassName: rbd-volumereplicationclass
    volumeReplicationName: vr-c4e68900-e0c5-465b-802a-c6fd011ff40c
  status:
    conditions:
    - lastTransitionTime: "2024-08-08T05:53:26Z"
      message: ""
      observedGeneration: 1
      reason: FailedToPromote
      status: "False"
      type: Completed
    - lastTransitionTime: "2024-08-08T05:53:26Z"
      message: ""
      observedGeneration: 1
      reason: Error
      status: "True"
      type: Degraded
    - lastTransitionTime: "2024-08-08T05:53:26Z"
      message: ""
      observedGeneration: 1
      reason: NotResyncing
      status: "False"
      type: Resyncing
    message: 'volume 0001-0009-rook-ceph-0000000000000002-104cfb18-4ea6-4916-b301-94401d466e34
      not found: Failed as image not found (internal RBD image not found)'
    observedGeneration: 1
    persistentVolumeClaimsRefList:
    - name: test-pvc
    state: Unknown
```

The CR is created and the dependend CRs i.e, VR and VGRContent created as well

```
apiVersion: v1
items:
- apiVersion: replication.storage.openshift.io/v1alpha1
  kind: VolumeGroupReplicationContent
  metadata:
    annotations:
      replication.storage.openshift.io/volumegroupref: volumegroupreplication-sample/rook-ceph
    creationTimestamp: "2024-08-08T05:53:25Z"
    finalizers:
    - replication.storage.openshift.io/vgr-protection
    generation: 2
    name: vgrcontent-c4e68900-e0c5-465b-802a-c6fd011ff40c
    resourceVersion: "193969"
    uid: 10c4a992-107b-49ad-aa77-6bc989f42184
  spec:
    provisioner: rook-ceph.rbd.csi.ceph.com
    source:
      volumeHandles:
      - 0001-0009-rook-ceph-0000000000000002-6526bb74-044a-489f-a821-7d8c042719b4
    volumeGroupReplicationClassName: volumegroupreplicationclass-sample
    volumeGroupReplicationHandle: 0001-0009-rook-ceph-0000000000000002-104cfb18-4ea6-4916-b301-94401d466e34
    volumeGroupReplicationRef:
      apiVersion: replication.storage.openshift.io/v1alpha1
      kind: VolumeGroupReplication
      name: volumegroupreplication-sample
      namespace: rook-ceph
      uid: c4e68900-e0c5-465b-802a-c6fd011ff40c
  status:
    persistentVolumeRefList:
    - name: pvc-7228cbb8-7a2f-4daf-bea4-4dabbd2ecabe
```
```
apiVersion: v1
items:
- apiVersion: replication.storage.openshift.io/v1alpha1
  kind: VolumeReplication
  metadata:
    annotations:
      replication.storage.openshift.io/volumegroupref: volumegroupreplication-sample/rook-ceph
    creationTimestamp: "2024-08-08T05:53:26Z"
    finalizers:
    - replication.storage.openshift.io
    generation: 1
    name: vr-c4e68900-e0c5-465b-802a-c6fd011ff40c
    namespace: rook-ceph
    ownerReferences:
    - apiVersion: replication.storage.openshift.io/v1alpha1
      kind: VolumeGroupReplication
      name: volumegroupreplication-sample
      uid: c4e68900-e0c5-465b-802a-c6fd011ff40c
    resourceVersion: "193974"
    uid: 2f1f45bf-417b-41ad-9973-71741dda555d
  spec:
    autoResync: false
    dataSource:
      apiGroup: replication.storage.openshift.io
      kind: VolumeGroupReplication
      name: volumegroupreplication-sample
    replicationHandle: ""
    replicationState: primary
    volumeReplicationClass: rbd-volumereplicationclass
  status:
    conditions:
    - lastTransitionTime: "2024-08-08T05:53:26Z"
      message: ""
      observedGeneration: 1
      reason: FailedToPromote
      status: "False"
      type: Completed
    - lastTransitionTime: "2024-08-08T05:53:26Z"
      message: ""
      observedGeneration: 1
      reason: Error
      status: "True"
      type: Degraded
    - lastTransitionTime: "2024-08-08T05:53:26Z"
      message: ""
      observedGeneration: 1
      reason: NotResyncing
      status: "False"
      type: Resyncing
    message: 'volume 0001-0009-rook-ceph-0000000000000002-104cfb18-4ea6-4916-b301-94401d466e34
      not found: Failed as image not found (internal RBD image not found)'
    observedGeneration: 1
    state: Unknown
```

nixpanic:
There are a few minor conflicts that need to be resolved. 

Is there a way to set the `status.conditions[*].message` instead, or in addition to the single (last?) `status.message`?

Nikhil-Ladha:
> There are a few minor conflicts that need to be resolved.

Resolved the conflicts.

> Is there a way to set the `status.conditions[*].message` instead, or in addition to the single (last?) `status.message`?

Well, we are just copying the status from VolumeReplication CR, so I don't think there is any way we can change that other than updating the VolumeReplication CR :/

Nikhil-Ladha:
Can we please have second round of review for the PR?
cc @nixpanic @Rakshith-R @Madhu-1

internal/controller/replication.storage/volumegroupreplication_controller.go
```diff
@@ -33,21 +59,682 @@ type VolumeGroupReplicationReconciler struct {
...
+		}
+	}
+
+	return pvcRefList, nil
+}
+
+func (r *VolumeGroupReplicationReconciler) createOrUpdateVolumeGroupReplicationContentCR(vgr *replicationv1alpha1.VolumeGroupReplication,
+	vgrContentObj *replicationv1alpha1.VolumeGroupReplicationContent, vgrClass string, pvHandlesList []string) error {
+	vgrRef := fmt.Sprintf("%s/%s", vgr.Name, vgr.Namespace)
+	_, err := controllerutil.CreateOrUpdate(context.TODO(), r.Client, vgrContentObj, func() error {
```
Madhu-1, line 665: save the context in VolumeGroupReplicationReconciler and use it

internal/controller/replication.storage/volumegroupreplication_controller.go
```diff
@@ -33,21 +59,682 @@ type VolumeGroupReplicationReconciler struct {
...
+		if vgr.Spec.VolumeGroupReplicationContentName != "" && vgrContentObj.CreationTimestamp.IsZero() {
+			return fmt.Errorf("dependent VGRContent resource is not yet created, waiting for it to be created")
```
Madhu-1, line 667: use errors.New()

internal/controller/replication.storage/volumegroupreplicationcontent_controller.go
```diff
@@ -18,36 +18,298 @@ package controller
...
+			if instance.Spec.VolumeGroupReplicationHandle != "" {
+				// Fetch the volume group using the group id
+				tmpVG, err := volumeGroupClient.ControllerGetVolumeGroup(instance.Spec.VolumeGroupReplicationHandle, secretName, secretNamespace)
+				if err != nil {
+					if status.Code(err) == codes.NotFound {
+						logger.Info("failed to get volume group, the volume group has already been deleted or doesn't exists")
+					} else {
+						logger.Error(err, "failed to get the volume group to be deleted", "Volume Group", instance.Spec.VolumeGroupReplicationHandle)
+						return reconcile.Result{}, err
+					}
```
Madhu-1, line 147: why do we need to do get?

We should issue modify and delete directly

Madhu-1 (commented):
need to do one more round of review

internal/controller/replication.storage/pvc.go
```diff
@@ -86,22 +92,23 @@ func (r *VolumeReplicationReconciler) annotatePVCWithOwner(ctx context.Context,
...
 }
 
-// removeOwnerFromPVCAnnotation removes the VolumeReplication owner from the PVC annotations.
-func (r *VolumeReplicationReconciler) removeOwnerFromPVCAnnotation(ctx context.Context, logger logr.Logger, pvc *corev1.PersistentVolumeClaim) error {
-	if _, ok := pvc.ObjectMeta.Annotations[replicationv1alpha1.VolumeReplicationNameAnnotation]; ok {
-		logger.Info("removing owner annotation from PersistentVolumeClaim object", "Annotation", replicationv1alpha1.VolumeReplicationNameAnnotation)
-		delete(pvc.ObjectMeta.Annotations, replicationv1alpha1.VolumeReplicationNameAnnotation)
-		if err := r.Client.Update(ctx, pvc); err != nil {
+// RemoveOwnerFromPVCAnnotation removes the VolumeReplication/VolumeGroupReplication owner from the PVC annotations.
+func RemoveOwnerFromPVCAnnotation(client client.Client, logger logr.Logger, pvc *corev1.PersistentVolumeClaim,
```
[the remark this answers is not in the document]
Nikhil-Ladha: We are exporting these functions because mutliple controllers VR, VGRContent and VGR use them to add finalizers/annotations to the respective owners. If we don't export them, then we would have to re-define the same function under each controller.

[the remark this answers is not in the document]
Madhu-1: As everything is in the same package, we dont need to export them, can you please check this?

internal/controller/replication.storage/volumegroupreplication_controller.go
```diff
@@ -33,21 +59,682 @@ type VolumeGroupReplicationReconciler struct {
...
+		}
+
+		// Check if PVCs exist based on provided selectors
+		pvcList, pvHandlesList, labelSelector, err := r.getMatchingPVCsFromSource(instance, logger, vgrClassObj)
+		if err != nil {
+			logger.Error(err, "failed to get PVCs using selector")
+			_ = r.setGroupReplicationFailure(instance, logger, err)
+			return reconcile.Result{}, err
+		}
+		if len(pvcList) > 100 {
```
Rakshith-R, line 159: Let's make this configurable with default as 100  ? 
similar to https://github.com/csi-addons/kubernetes-csi-addons/blob/main/docs/csi-addons-config.md

internal/controller/replication.storage/volumegroupreplication_controller.go
```diff
@@ -33,21 +59,682 @@ type VolumeGroupReplicationReconciler struct {
...
+
+		// If it's not in oldPVCs, it's an addition
+		if _, exists := oldPVCs[pvc.Name]; !exists {
+			toAddPVCs = append(toAddPVCs, pvc)
+		}
+	}
+
+	// Identify PVCs that should be removed
+	for name := range oldPVCs {
+		if _, exists := currentPVCs[name]; !exists {
```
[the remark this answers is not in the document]
Nikhil-Ladha: If we remove the PVC, it will be there in `oldPVCs` because that contains the list from the VGR's status. We need the `currentPVCs` list to identify the diff, and remove the finalizers, annotation for the removed PVCs.

internal/controller/replication.storage/volumegroupreplicationcontent_controller.go
Nikhil-Ladha: Earlier, we had the bug for DeleteVolumeGroup RPC for not being idempetent so added this and later on forgot to remove it.
Fixed it now.

internal/controller/replication.storage/volumegroupreplication_controller.go
Nikhil-Ladha: Done

api/replication.storage/v1alpha1/volumegroupreplicationcontent_types.go
```diff
@@ -21,6 +21,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	VolumeGroupReplicationContentNameAnnotation = "replication.storage.openshift.io/volumegroupreplication-content-name"
```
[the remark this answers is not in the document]
Nikhil-Ladha: Ack

Nikhil-Ladha:
@Madhu-1 addressed all your suggestions, please take a look now.

internal/controller/replication.storage/volumegroupreplication_controller.go
Nikhil-Ladha: Sounds good, but probably we can take it up as a follow up PR?

Madhu-1 (approved):
high level LGTM, we can do cosmetic changes as followup

internal/controller/replication.storage/finalizers.go
```diff
@@ -126,3 +133,35 @@ func (r *VolumeReplicationReconciler) removeFinalizerFromVGR(logger logr.Logger,
 
 	return nil
 }
+
+// addFinalizerToVGRContent adds the VR, VGR finalizer on the VolumeGroupReplicationContent resource
```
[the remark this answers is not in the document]
Nikhil-Ladha: Ah, correct. Will update it.

internal/controller/replication.storage/utils.go
```diff
@@ -0,0 +1,73 @@
+/*
+Copyright 2024 The Kubernetes-CSI-Addons Authors.
```
Rakshith-R, line 2:
```suggestion
Copyright 2025 The Kubernetes-CSI-Addons Authors.
```

internal/controller/replication.storage/volumegroupreplicationclass.go
```diff
@@ -0,0 +1,43 @@
+/*
+Copyright 2024 The Kubernetes-CSI-Addons Authors.
```
Rakshith-R, line 2:
```suggestion
Copyright 2025 The Kubernetes-CSI-Addons Authors.
```

internal/controller/replication.storage/volumegroupreplication_test.go
```diff
@@ -0,0 +1,203 @@
+/*
+Copyright 2024 The Kubernetes-CSI-Addons Authors.
```
Rakshith-R, line 2:
```suggestion
Copyright 2025 The Kubernetes-CSI-Addons Authors.
```

internal/controller/replication.storage/volumegroupreplication_controller.go
```diff
@@ -18,36 +18,752 @@ package controller
...
+
+		// Add the string representation of the labelSelector to the VGR annotation
+		if instance.ObjectMeta.Annotations == nil {
+			instance.ObjectMeta.Annotations = make(map[string]string)
+		}
+
+		// We need to save the label selector in the annotation, so that an event in PVC
+		// triggers the reconcile particularly for the VGR that the PVC is part of by comparing
+		// the labels on the pvc with the pvcSelector annotation in VGR
+		if instance.ObjectMeta.Annotations["pvcSelector"] != labelSelector {
```
Rakshith-R, line 181: define "pvcSelector" as a const

Rakshith-R (changes requested):

internal/controller/replication.storage/volumegroupreplication_controller.go
```diff
@@ -18,36 +18,752 @@ package controller
...
+			if err != nil {
+				return reconcile.Result{}, err
+			}
+		}
+
+		// Remove volumereplication owner annotation from VGR
+		err = r.removeOwnerFromVGRAnnotation(instance)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
```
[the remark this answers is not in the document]
Nikhil-Ladha: No, VGRC is created by VGR so, first VGRC should be deleted before VGR.
Coming to PVC, images should be first removed from the group and that implies freeing up the PVCs before deleting the VGR.
Hopefully, that clears the doubt.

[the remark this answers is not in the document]
Nikhil-Ladha: The VR might get deleted earlier if the VGRC deletion is stuck due to some reason, therefore to avoid such a scenario the VR deletion is take care of by the owner i.e, VGR.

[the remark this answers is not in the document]
Rakshith-R: > The VR might get deleted earlier if the VGRC deletion is stuck due to some reason, therefore to avoid such a scenario the VR deletion is take care of by the owner i.e, VGR.

I couldn't find where VR is deleted
is the VR not deleted explicitly ? 

Let's change the PVC finalizer removal to the last.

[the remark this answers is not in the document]
Nikhil-Ladha: > is the VR not deleted explicitly ?

No, it is garbage collected after deletion of VGR.

> Let's change the PVC finalizer removal to the last.

Yeah, actually rethinking about it a bit, we should probably move it to VGRContent reconciler? So that in case the `ModifyVolumeGroup` RPC fails we won't be having un-protected PVCs i.e, free to delete. The finalizers should only be removed once the images have been removed from the group successfully.

WDYT @Madhu-1 @Rakshith-R ?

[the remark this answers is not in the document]
Nikhil-Ladha: > I think VR should be explicitly deleted before VolGrpRepContent deletion.

But, that would cause stale VRs in the cluster. I don't think we would want that.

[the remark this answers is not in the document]
Nikhil-Ladha: Updated the order of PVCs though.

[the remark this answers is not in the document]
Rakshith-R: > > I think VR should be explicitly deleted before VolGrpRepContent deletion.
> 
> But, that might cause stale VGRCs in the cluster (i.e, without a VR). I don't think we would want that.

I don't see how that can happen.
Delete VR
Then delete VGRC 

If we need to add additional checks above this block,  we'll need to do so.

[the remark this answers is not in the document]
Nikhil-Ladha: > Delete VR

Deleted
> Then delete VGRC

Stuck deletion because the `Modify` volume group failed.
Then, we can come across this scenario.

Even if we add checks like finalizers to avoid this, that implies VR will be stuck still VGRC deletion. So, what's the benefit of having the extra check?

The initial code was based on this approach itself.
That is: PVC deleted, VR deleted, VGRC deleted and then VGR deleted. But, I have seen the above happening and that's the reason this was done.

internal/controller/replication.storage/volumegroupreplication_controller.go
nixpanic: It is fine to make this change in a follow-up PR. Please open an issue and link it here.

internal/controller/replication.storage/volumegroupreplication_controller.go
```diff
@@ -18,36 +18,752 @@ package controller
...
+	if err := r.Client.Status().Update(r.ctx, instance); err != nil {
+		r.log.Error(err, "failed to update volumeGroupReplication status", "VGRName", instance.Name)
+		return err
+	}
+
+	return nil
+}
+
+// getMatchingPVCsFromSource fecthes the PVCs based on the selectors defined in the VolumeGroupReplication resource
+func (r *VolumeGroupReplicationReconciler) getMatchingPVCsFromSource(instance *replicationv1alpha1.VolumeGroupReplication,
```
nixpanic, line 484: Why return the pvc-handles too? It feels that this function does more than it needs to do. It is a little cleaner to have a function to fetch the handle from a PVC and call that when the pvc-handle is needed.

internal/controller/replication.storage/volumegroupreplication_controller.go
```diff
@@ -18,36 +18,752 @@ package controller
...
+		})
+
+		// If it's not in oldPVCs, it's an addition
+		if _, exists := oldPVCs[pvc.Name]; !exists {
+			toAddPVCs = append(toAddPVCs, pvc)
+		}
+	}
+
+	// Identify PVCs that should be removed
+	for name := range oldPVCs {
```
nixpanic, line 587: In other for loops you use `_` for the index, here you don't. Please follow the same coding style everywhere.

internal/controller/replication.storage/volumereplication_controller.go
```diff
@@ -113,7 +111,7 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	vrcObj, err := r.getVolumeReplicationClass(logger, instance.Spec.VolumeReplicationClass)
 	if err != nil {
 		setFailureCondition(instance, "failed to get volumeReplication class", err.Error(), instance.Spec.DataSource.Kind)
-		uErr := r.updateReplicationStatus(instance, logger, getCurrentReplicationState(instance), err.Error())
+		uErr := r.updateReplicationStatus(instance, logger, GetCurrentReplicationState(instance.Status.State), err.Error())
```
nixpanic, line 114: changes like this could have been a separate commit

internal/controller/replication.storage/finalizers.go
```diff
@@ -65,43 +66,44 @@ func (r *VolumeReplicationReconciler) removeFinalizerFromVR(logger logr.Logger,
...
 }
 
-// addFinalizerToPVC adds the VR finalizer on the PersistentVolumeClaim.
-func (r *VolumeReplicationReconciler) addFinalizerToPVC(logger logr.Logger, pvc *corev1.PersistentVolumeClaim) error {
-	if !slices.Contains(pvc.ObjectMeta.Finalizers, pvcReplicationFinalizer) {
-		logger.Info("adding finalizer to PersistentVolumeClaim object", "Finalizer", pvcReplicationFinalizer)
-		pvc.ObjectMeta.Finalizers = append(pvc.ObjectMeta.Finalizers, pvcReplicationFinalizer)
-		if err := r.Client.Update(context.TODO(), pvc); err != nil {
+// addFinalizerToPVC adds the VR, VGR finalizer on the PersistentVolumeClaim.
+func addFinalizerToPVC(client client.Client, logger logr.Logger, pvc *corev1.PersistentVolumeClaim,
```
nixpanic, line 70: What is the reason to remove this function (and others) from the struct, and make it a stand-alone helper? Callers should not need to care about the `client.Client` that is passed.

api/replication.storage/v1alpha1/volumegroupreplication_types.go
```diff
@@ -21,6 +21,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	VolumeGroupReplicationNameAnnotation = "replication.storage.openshift.io/volume-group-replication-name"
```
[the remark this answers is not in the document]
Nikhil-Ladha: Yes, VR is already using the openshift naming.
That's the reason I used the same to keep it consistent.

internal/controller/replication.storage/finalizers.go
Nikhil-Ladha: We are now referencing this function in multiple controllers. So, as to mkae the `client` value independent of the caller I moved it out of the struct. Otherwise, we would be just duplicating the function in both the controllers.

internal/controller/replication.storage/volumereplication_controller.go
Nikhil-Ladha: Ack, moved it to a separate commit.

internal/controller/replication.storage/pvc.go
```diff
@@ -59,22 +60,27 @@ func (r VolumeReplicationReconciler) getPVCDataSource(logger logr.Logger, req ty
...
+	pvc *corev1.PersistentVolumeClaim, pvcAnnotation string) error {
 	if pvc.ObjectMeta.Annotations == nil {
 		pvc.ObjectMeta.Annotations = map[string]string{}
 	}
 
-	currentOwnerName := pvc.ObjectMeta.Annotations[replicationv1alpha1.VolumeReplicationNameAnnotation]
+	if pvc.ObjectMeta.Annotations[replicationv1alpha1.VolumeReplicationNameAnnotation] != "" &&
+		pvc.ObjectMeta.Annotations[replicationv1alpha1.VolumeGroupReplicationNameAnnotation] != "" {
+		logger.Info("PVC can't be part of both VolumeGroupReplication and VolumeReplication")
+		return fmt.Errorf("PVC (%q/%q) can't be owned by both VolumeReplication and VolumeGroupReplication", pvc.Namespace, pvc.Name)
```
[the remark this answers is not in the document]
Nikhil-Ladha: Done

internal/controller/replication.storage/volumegroupreplication_controller.go
Nikhil-Ladha: Done, created a separate function for pvHandles

internal/controller/replication.storage/volumegroupreplication_controller.go
Nikhil-Ladha: Actually, this is a map that we are iterating over.
Though, I have updated it use `_` style for values, as those are unused.

internal/controller/replication.storage/volumegroupreplication_controller.go
Nikhil-Ladha: Created issue #787 for it

internal/controller/replication.storage/volumegroupreplication_controller.go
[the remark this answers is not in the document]
Rakshith-R: Changing the deletion flow to fit a particular scenario sounds like a hack.

How is VR deletion affected by Modify volume group call ?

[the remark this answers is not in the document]
Rakshith-R: > It's not a particular scenario but a generic use case that can happen for any storage vendor using this addon.
> 
> > How is VR deletion affected by Modify volume group call ?
> 
> Because we would be deleting the VR first, and then proceeding to delete the VGRC, so if something wrong goes during VGRC deletion, then the VGRC deletion would be stuck but VR would be deleted.

VRGC stuck in deletion while VR is deleted is still fine right ? 

Same could happen during creation phase
VRGC creation gets stuck due to some issue
then VR will never be created at all.

@nixpanic @Madhu-1 Can you please add your opinions on this ?

nixpanic (approved):

internal/controller/replication.storage/volumegroupreplication_controller.go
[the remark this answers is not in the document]
Nikhil-Ladha: > VRGC stuck in deletion while VR is deleted is still fine right ?

How is that fine? That leaves the clusters in an inconsistent state with mirroring disabled.

> Same could happen during creation phase
VRGC creation gets stuck due to some issue
then VR will never be created at all.

That's actually fine, because if the VR is not created that means the group is still not mirror enabled/promoted.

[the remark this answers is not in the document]
Rakshith-R: Why is that a problem?
Deletion of VGR is meant to disable mirroring and delete grouping. 

Same logic as if you have answered for creation applies for deletion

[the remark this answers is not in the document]
nixpanic: > VRGC stuck in deletion while VR is deleted is still fine right ?

Yes, similar to PV and PVC deletion.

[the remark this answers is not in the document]
Nikhil-Ladha: So, you suggest that we should delete the VR explicitly too and delete VGRC post that, is that right?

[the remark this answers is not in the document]
Nikhil-Ladha: Updated the code to delete VR explicitly as well. Please take a look now.
Thanks.

internal/controller/replication.storage/volumegroupreplication_controller.go
```diff
@@ -18,36 +18,772 @@ package controller
...
+		r.log.Info("volumeReplication resource not found")
+	} else {
+		err = r.Client.Delete(r.ctx, vrObj)
+		if err != nil {
+			r.log.Error(err, "failed to delete dependent volumeReplication resource")
+			return err
+		}
+	}
+
+	return nil
```
Rakshith-R, line 788: There's no need for else block here.
```suggestion
	} 
		err = r.Client.Delete(r.ctx, vrObj)
		if err != nil {
			r.log.Error(err, "failed to delete dependent volumeReplication resource")
			return err
		}

	return nil
```

Rakshith-R (commented):
small nit
other parts look good to me

internal/controller/replication.storage/volumegroupreplication_controller.go
Nikhil-Ladha: @Rakshith-R ^^
We can avoid the `else` statement by adding a `return nil` at the end of `if` block, if that seems better?

Rakshith-R: > @Rakshith-R ^^ We can avoid the `else` block by adding a `return nil` at the end of `if` block, if that seems better?

yes, let's do it this way.

The current flow makes it a bit confusing.

Nikhil-Ladha: Done

Rakshith-R (approved):
Thanks, lgtm

internal/controller/replication.storage/finalizers.go
```diff
@@ -65,43 +66,44 @@ func (r *VolumeReplicationReconciler) removeFinalizerFromVR(logger logr.Logger,
...
+	replicationFinalizer string) error {
+	if !slices.Contains(pvc.ObjectMeta.Finalizers, replicationFinalizer) {
```
nixpanic, line 72: The new golangci-lint version does not like the `.ObjectMeta` parent-struct notation. You will need to drop that. Other parent-structs needs adjusting for that as well, calls like `r.Client.Update()` seem to be relatively common.

internal/controller/replication.storage/pvc.go
```diff
@@ -24,14 +24,15 @@ import (
...
 
 	replicationv1alpha1 "github.com/csi-addons/kubernetes-csi-addons/api/replication.storage/v1alpha1"
 )
 
 // getPVCDataSource get pvc, pv object from the request.
-func (r VolumeReplicationReconciler) getPVCDataSource(logger logr.Logger, req types.NamespacedName) (*corev1.PersistentVolumeClaim, *corev1.PersistentVolume, error) {
+func (r VolumeReplicationReconciler) getPVCDataSource(ctx context.Context, logger logr.Logger, req types.NamespacedName) (*corev1.PersistentVolumeClaim, *corev1.PersistentVolume, error) {
 	pvc := &corev1.PersistentVolumeClaim{}
-	err := r.Client.Get(context.TODO(), req, pvc)
+	err := r.Client.Get(ctx, req, pvc)
```
nixpanic, line 35: use `r.Get()` instead of `r.Client.Get()` as indicated by the latest golangci-linter

internal/controller/replication.storage/volumegroupreplication_controller.go
```diff
@@ -18,36 +18,772 @@ package controller
...
+	vrObj := &replicationv1alpha1.VolumeReplication{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      vrName,
+			Namespace: instance.Namespace,
+		},
+	}
+
+	// Create/Update dependent resources only if the instance is not marked for deletion
+	if instance.GetDeletionTimestamp().IsZero() {
+
```
nixpanic, line 153: nit: no need for this empty line

nixpanic (commented):
Looks good, but need to resolve conflicts. Mainly the updated golangci-lint version does not like referencing parent structs like `r.Client.Get()` when `r.Get()` is the same, or `pvc.ObjectMeta.Name` when you can use the shorter `pvc.Name`.

nixpanic (approved):

Rakshith-R (approved):

PR #608, opened 2024-07-01
Implement EncryptionKeyRotation controllers

black-dragon74:
This patch implements the sidecar service for
EncryptionKeyRotation and registers it

A new option is added to csi-addons test tool
to test the EncryptionKeyRotation operation

PR #616, opened 2024-07-15
Modify pvc controller to reconcile encryptionkeyrotation

black-dragon74:
This patch modified the PVC controller so that it
reconciles the pvcs for encryptionkeyrotation
operations.

**Note:** Please review the most recent commit only as the rest of them are unrelated to this feature. The extra commits are due to this PR being based off of impl-pvkr branch.

Depends-on: #608 

Regards

### DIFF
--- a/internal/controller/csiaddons/persistentvolumeclaim_controller.go
+++ b/internal/controller/csiaddons/persistentvolumeclaim_controller.go
@@ -56,9 +56,14 @@ type PersistentVolumeClaimReconciler struct {
 var (
 	rsCronJobScheduleTimeAnnotation = "reclaimspace." + csiaddonsv1alpha1.GroupVersion.Group + "/schedule"
 	rsCronJobNameAnnotation         = "reclaimspace." + csiaddonsv1alpha1.GroupVersion.Group + "/cronjob"
-	csiAddonsDriverAnnotation       = "reclaimspace." + csiaddonsv1alpha1.GroupVersion.Group + "/drivers"
-	ErrConnNotFoundRequeueNeeded    = errors.New("connection not found, requeue needed")
-	ErrScheduleNotFound             = errors.New("schedule not found")
+	rsCSIAddonsDriverAnnotation     = "reclaimspace." + csiaddonsv1alpha1.GroupVersion.Group + "/drivers"
+
+	krcJobScheduleTimeAnnotation = "keyrotation." + csiaddonsv1alpha1.GroupVersion.Group + "/schedule"
+	krcJobNameAnnotation         = "keyrotation." + csiaddonsv1alpha1.GroupVersion.Group + "/cronjob"
+	krCSIAddonsDriverAnnotation  = "keyrotation." + csiaddonsv1alpha1.GroupVersion.Group + "/drivers"
+
+	ErrConnNotFoundRequeueNeeded = errors.New("connection not found, requeue needed")
+	ErrScheduleNotFound          = errors.New("schedule not found")
 )
 
 const (
@@ -68,6 +73,7 @@ const (
 //+kubebuilder:rbac:groups=core,resources=persistentvolumeclaims,verbs=get;list;watch;patch
 //+kubebuilder:rbac:groups=core,resources=persistentvolumeclaims/finalizers,verbs=update
 //+kubebuilder:rbac:groups=csiaddons.openshift.io,resources=reclaimspacecronjobs,verbs=get;list;watch;create;delete;update
+//+kubebuilder:rbac:groups=csiaddons.openshift.io,resources=encryptionkeyrotationcronjobs,verbs=get;list;watch;create;delete;update
 //+kubebuilder:rbac:groups=core,resources=namespaces,verbs=get;list;watch
 //+kubebuilder:rbac:groups=storage.k8s.io,resources=storageclasses,verbs=get;list;watch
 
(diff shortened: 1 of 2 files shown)


internal/controller/csiaddons/persistentvolumeclaim_controller.go
```diff
@@ -114,6 +118,11 @@ func (r *PersistentVolumeClaimReconciler) Reconcile(ctx context.Context, req ctr
 		return ctrl.Result{}, nil
 	}
 
+	err = r.processKeyRotation(ctx, &logger, &req, pvc, pv)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
 	rsCronJob, err := r.findChildCronJob(ctx, &logger, &req)
```
Madhu-1, line 126: move the reclaimspacecron job to its own helper function

internal/controller/csiaddons/persistentvolumeclaim_controller.go
```diff
@@ -258,7 +267,7 @@ func (r *PersistentVolumeClaimReconciler) determineScheduleAndRequeue(
 	driverName string,
 ) (string, error) {
 	annotations := pvc.GetAnnotations()
-	schedule, scheduleFound := getScheduleFromAnnotation(logger, annotations)
+	schedule, scheduleFound := getScheduleFromAnnotation(rsCronJobScheduleTimeAnnotation, logger, annotations)
```
Madhu-1, line 270: pass the annotation as part of `determineScheduleAndRequeue function argument so the caller can pass the required key to check

internal/controller/csiaddons/persistentvolumeclaim_controller.go
```diff
@@ -491,3 +505,182 @@ func (r PersistentVolumeClaimReconciler) supportsReclaimSpace(driverName string)
...
+				logger.Error(err, "failed to delete extraneous child encryptionkeyrotationcronjob", "EncryptionKeyrotationCronJobName", &job.Name)
+				return nil, fmt.Errorf("failed to delete extraneous child encryptionkeyrotationcronjob: %w", err)
+			}
+		}
+	}
+
+	return activeJob, nil
+}
+
+func (r *PersistentVolumeClaimReconciler) determineEncryptionKeyRotationSchedule(
```
Madhu-1, line 541: lets not have duplicate helper use existing on and pass the required key as input

Madhu-1 (commented):
some nits, looks like this also need larger refactor to avoid duplication, lets address that later

internal/controller/csiaddons/persistentvolumeclaim_controller.go
```diff
@@ -114,6 +118,11 @@ func (r *PersistentVolumeClaimReconciler) Reconcile(ctx context.Context, req ctr
 		return ctrl.Result{}, nil
 	}
 
+	err = r.processKeyRotation(ctx, &logger, &req, pvc, pv)
```
iPraveenParihar, line 121: If keyRotation errors out for some reason, then ReclaimSpace will not be processed?

internal/controller/csiaddons/persistentvolumeclaim_controller.go
black-dragon74: You mean the entirety of reclaimspace operations? Like the way it is done for key rotation?

Madhu-1: yes correct

internal/controller/csiaddons/persistentvolumeclaim_controller.go
black-dragon74: The logic differs between the existing function and this one. We do not check namespaces and pvc annotations but only the storage class.

internal/controller/csiaddons/persistentvolumeclaim_controller.go
black-dragon74: Correct, one of the side effects of having a single pv controller for both the operations :( We could change the ordering but then the situation would be reversed, if relciamspace fails, encryption key rotation would not be processed. 

Using sync would not benefit as if the current reconcile returns (from any sub-routine), the reconcile will end.

One possible solution would be to implement a new storage class controller for key rotation but then a SC can have a lot of PVCs and we will have to process each one of them, a very significant overhead. This was discussing during the design phase.

Madhu-1: can we keep this a two different operations like run in parallel and reqeue if one fails?

internal/controller/csiaddons/persistentvolumeclaim_controller.go
Madhu-1: why, if we just refactor we can support all the functionalities what we have for reclaimspace, any concern with that?

black-dragon74: Did not know we could do that (as it differed from the design). In that case, we can make key rotation work with namespace annotations as well. Win win!

internal/controller/csiaddons/persistentvolumeclaim_controller.go
black-dragon74: Better yet, let's just log the encryption key rotation if it fails and let reclaim space continue. In the end we check the return values for both and requeue if any of them errored out.

black-dragon74: This is implemented now, PTAL @Madhu-1 @iPraveenParihar

internal/controller/csiaddons/persistentvolumeclaim_controller.go
```diff
@@ -245,44 +172,48 @@ func (r *PersistentVolumeClaimReconciler) checkDriverSupportReclaimsSpace(logger
...
-//   - If driver does not support reclaimSpace, return ErrScheduleNotFound.
-//     Depending on requeue value, it will throw ErrorConnNotFoundRequeueNeeded.
+//   - Check if the schedule is present in the storageclass annotations. If yes,
+//     use that.
 func (r *PersistentVolumeClaimReconciler) determineScheduleAndRequeue(
 	ctx context.Context,
 	logger *logr.Logger,
 	pvc *corev1.PersistentVolumeClaim,
 	driverName string,
+	annotation string,
```
Madhu-1, line 182: annotationKey?

internal/controller/csiaddons/persistentvolumeclaim_controller.go
```diff
@@ -245,44 +172,48 @@ func (r *PersistentVolumeClaimReconciler) checkDriverSupportReclaimsSpace(logger
...
 
 	// If the schedule is found, check whether driver supports the
 	// space reclamation using annotation on namespace and registered driver
 	// capability for decision on requeue.
 	if scheduleFound {
+		// If the check is for key rotation
+		// return the schedule directly
+		if annotation == krcJobScheduleTimeAnnotation {
+			return schedule, nil
+		}
```
Madhu-1, line 207: we need to have a driver support check right? we cannot just return from here, the flow should remain same for both key rotation and reclaimspace.

black-dragon74: The capability is already checked for in `getClientByNode` in `EncryptionKeyRotationJobController`. If no nodeclient is found, the reconcile ends there. May I know the benefit it brings to add a second check here as well?

internal/controller/csiaddons/persistentvolumeclaim_controller.go
```diff
@@ -241,48 +169,87 @@ func (r *PersistentVolumeClaimReconciler) checkDriverSupportReclaimsSpace(logger
...
+
+		// We cannot have a generic solution for all CSI drivers to get the driver
+		// name from PV and check if driver supports space reclamation or not and
+		// requeue the request if the driver is not registered in the connection
+		// pool. This can put the controller in a requeue loop. Hence we are
+		// reading the driver name from the namespace annotation and checking if
+		// the driver is registered in the connection pool and if not we are not
+		// requeuing the request.
+		// Depending on requeue value, it will return ErrorConnNotFoundRequeueNeeded.
```
Madhu-1, line 252: move this comment to top of above if check and move reclaimspace check into `else if annotationKey == reclaimspaceScheduleTimeAnnotation`

Madhu-1 (commented):
LGTM. reviewed only last commit.

black-dragon74:
> LGTM. reviewed only last commit.

There was a logic bug on line 712. Along with that I had forgotten to add the field indexer for EKRCronJobs owned by PVCs.

Both issues are fixed and a thorough testing has been done on my end.

```go
err = r.Client.Update(ctx, krcJob) // was err = r.Client.Update(ctx, newKrcJob)
```

Code for new field indexer (lines 301-322)
```go
err = mgr.GetFieldIndexer().IndexField(
		context.Background(),
		&csiaddonsv1alpha1.EncryptionKeyRotationCronJob{},
		jobOwnerKey,
		func(rawObj client.Object) []string {
			job, ok := rawObj.(*csiaddonsv1alpha1.EncryptionKeyRotationCronJob)
			if !ok {
				return nil
			}
			owner := metav1.GetControllerOf(job)
			if owner == nil {
				return nil
			}
			if owner.APIVersion != "v1" || owner.Kind != "PersistentVolumeClaim" {
				return nil
			}

			return []string{owner.Name}
		})
	if err != nil {
		return err
	}
```

Regards

nixpanic:
Please combine commits in the PR that are related together. Ideally you have a few commits left that cleanly implement a functionality and also include the improvements for review comments.

black-dragon74:
> Please combine commits in the PR that are related together. Ideally you have a few commits left that cleanly implement a functionality and also include the improvements for review comments.

Only the last commit in this PR is related to the PVC controller. Rest are due to this PR being rebased off of the PR implementing the controllers for key rotation. 

I will merge the commits in that PR and will update this one too :)

black-dragon74:
> Please combine commits in the PR that are related together.

PTAL @nixpanic

api/csiaddons/v1alpha1/encryptionkeyrotationcronjob_types.go
```diff
@@ -0,0 +1,128 @@
...
+	// Information when was the last time the job was successfully scheduled.
+	LastScheduleTime *metav1.Time `json:"lastScheduleTime,omitempty"`
+
+	// Information when was the last time the job successfully completed.
+	LastSuccessfulTime *metav1.Time `json:"lastSuccessfulTime,omitempty"`
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+//+kubebuilder:printcolumn:JSONPath=".spec.schedule",name=Schedule,type=string
```
nixpanic, line 99: either add spaces before `+kubebuilder`, or do not. Don't mix them in the same block please.

cmd/manager/main.go
```diff
@@ -25,6 +25,7 @@ import (
 	csiaddonsv1alpha1 "github.com/csi-addons/kubernetes-csi-addons/api/csiaddons/v1alpha1"
 	replicationstoragev1alpha1 "github.com/csi-addons/kubernetes-csi-addons/api/replication.storage/v1alpha1"
 	"github.com/csi-addons/kubernetes-csi-addons/internal/connection"
+
```
nixpanic, line 28: this is an unrelated change

.github/workflows/test-golang.yaml
```diff
@@ -24,7 +24,7 @@ jobs:
         # action @v1.3.0 can install protoc 3.19, @v2 installs protoc v21.0+
         uses: arduino/setup-protoc@v1.3.0
         with:
-          version: "3.19.6"
+          version: "3.20.2"
```
nixpanic, line 27: it would be better to move this, and the changes that it causes, to it's own commit

internal/controller/csiaddons/persistentvolumeclaim_controller.go
```diff
@@ -491,3 +484,274 @@ func (r PersistentVolumeClaimReconciler) supportsReclaimSpace(driverName string)
...
+
+func (r *PersistentVolumeClaimReconciler) processKeyRotation(
+	ctx context.Context,
+	logger *logr.Logger,
+	req *reconcile.Request,
+	pvc *corev1.PersistentVolumeClaim,
+	pv *corev1.PersistentVolume) error {
+	krcJob, err := r.findChildEncryptionKeyRotationCronJob(ctx, logger, req)
+	if err != nil {
+		return err // ctrl.Result{}
```
nixpanic, line 675: what do you mean by these `ctrl.Result{}` on each return? Maybe a description of the function in a comment would be good.

nixpanic (commented):
Added a few small comments. Ideally every function that is added contains a description of what it does. This is more important for larger functions (say > 15-20 lines of code).

api/csiaddons/v1alpha1/encryptionkeyrotationcronjob_types.go
black-dragon74: Resolved, would like to add, kubebuilder now adds spaces by default on v4 layout.

.github/workflows/test-golang.yaml
black-dragon74: Done.

internal/controller/csiaddons/encryptionkeyrotationjob_controller.go
```diff
@@ -0,0 +1,332 @@
...
+	// Set defaults for optional values
+	if krJob.Spec.BackoffLimit == 0 {
+		krJob.Spec.BackoffLimit = defaultBackoffLimit
+	}
+	if krJob.Spec.RetryDeadlineSeconds == 0 {
+		krJob.Spec.RetryDeadlineSeconds = defaultRetryDeadlineSeconds
+	}
+
+	// Reconcile the resource
+	err = r.reconcileEncryptionKeyRotationJob(
```
iPraveenParihar, line 118:
```suggestion
err = r.reconcile(
```
It is good practice to maintain consistent naming conventions throughout the codebase 😀

nixpanic (approved):

nixpanic:
only reviewed the last commit, because everything else is part of #608

black-dragon74:
Rebased it to main.

/cc @Rakshith-R @nixpanic @Madhu-1

Madhu-1 (approved):

Rakshith-R (approved):

Madhu-1:
@mergifyio rebase

PR #625, opened 2024-07-25
Remove special handling of targetpath for deprecated Kubernetes 1.24

### LINKED ISSUE #623: ReclaimSpaceJob (and potentially others) failing creation on Kubernetes 1.28+
When running on OKD, creating a ReclaimSpaceJob fails with:

> Failed to make node request: failed to convert Kubernetes minor version "28+" to int: strconv.Atoi: parsing "28+": invalid syntax

The versions given are:

```
Client Version: 4.14.6
Kustomize Version: v5.0.1
Server Version: 4.15.0-0.okd-2024-02-10-035534
Kubernetes Version: v1.28.2-3572+f1618d54a81f9f-dirty
```

The [type conversion](https://github.com/csi-addons/kubernetes-csi-addons/blob/main/internal/sidecar/service/reclaimspace.go#L220) may be too primitive to catch all kinds of Kubernetes version, perhaps a [SemVer library](https://pkg.go.dev/github.com/Masterminds/semver/v3) should be used instead.

psy-q:
This handling is no longer necessary as Kuberenetes 1.24 is not supported anymore.

Additionally, the version comparison that was used will break on some Kubernetes versions (any SemVer with a non-integer component). Removing the check fixes this problem as well.

Closes #623

### DIFF
--- a/internal/sidecar/service/reclaimspace.go
+++ b/internal/sidecar/service/reclaimspace.go
@@ -20,7 +20,6 @@ import (
 	"crypto/sha256"
 	"fmt"
 	"path/filepath"
-	"strconv"
 
 	kube "github.com/csi-addons/kubernetes-csi-addons/internal/kubernetes"
 	"github.com/csi-addons/kubernetes-csi-addons/internal/proto"
@@ -144,11 +143,7 @@ func (rs *ReclaimSpaceServer) NodeReclaimSpace(
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
-	stPath, err := rs.getStagingTargetPath(pv)
-	if err != nil {
-		klog.Errorf("Failed to get staging target path: %v", err)
-		return nil, status.Error(codes.Internal, err.Error())
-	}
+	stPath := rs.getStagingTargetPath(pv)
 
 	accessType := csi.VolumeCapability_Mount{
 		Mount: &csi.VolumeCapability_MountVolume{},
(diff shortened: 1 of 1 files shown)


iPraveenParihar:
Thanks @psy-q, Can you please squash your commits into one and add commit sign-off?

psy-q:
Thanks, I've squashed the commits now. Sorry for the noisy PR, it's my first contribution in Go. If there's something I've missed, please let me know.

internal/sidecar/service/reclaimspace.go
```diff
@@ -212,19 +212,14 @@ func (rs *ReclaimSpaceServer) getStagingTargetPath(pv *corev1.PersistentVolume)
...
-	minor, err := strconv.Atoi(version.Minor)
-	if err != nil {
-		return "", fmt.Errorf("failed to convert Kubernetes minor version %q to int: %w", version.Minor, err)
-	}
-
-	// 'encode' major/minor in a single integer
-	legacyVersion := 1024 // Kubernetes 1.24 => 1 * 1000 + 24
-	if ((major * 1000) + minor) < (legacyVersion) {
+	legacyVersion, _ := semver.NewVersion("v1.24")
+	if semanticVersion.LessThan(legacyVersion) {
```
Madhu-1, line 222: can we use `https://pkg.go.dev/k8s.io/apimachinery/pkg/util/version#Version.LessThan` package for this one?  we don't need to extract the version and do compare by ourself.

psy-q: Good idea, I've tried switching to that version and got rid of the vendored items.

Madhu-1: @psy-q Sorry again i missed adding one point, we can remove this entire check as we don't need to support 1.24 anyone as its deprecated or unsupported Kubernetes version. IMO we can remove all these checks

psy-q:
I've spent some time yesterday trying to write tests for this but:

* Running the code in question requires a `ReclaimSpaceServer`
  * That needs a fake Kubernetes server, but I'm not sure `fake` can fake different version responses
  * The server needs a fake gRPC client, I've tried just mocking the `Connect()` and `Close()` functions but didn't get far
* It needs a mocked PVC to pass to `getStagingTargetPath()` -- depending on how convincing the mock needs to be, this is probably the easiest part with some code from pvc_test.go

 I'm out of my depth here and not getting any results but I'd be willing to work on a test a bit more if there's some easier way to get a `ReclaimSpaceServer`.

nixpanic (approved):
This looks good to me, thanks!

If you want to test this out, the easiest is to introduce a function like

```go
func needsLegacyGlobalmount(serverVersion string) bool {
    semanticVersion := version.MustParseSemantic(serverVersion)
    legacyVersion := version.MustParseSemantic("v1.24.0")

    return semanticVersion.LessThan(legacyVersion)
}
```

This then makes it possible to test with different versions in `reclaimspace_test.go`.

internal/sidecar/service/reclaimspace.go
Madhu-1: @psy-q @nixpanic instead of getting it fixed why not remove it as its unsupported version?

Rakshith-R (approved):

Rakshith-R (changes requested):
missed madhu's comment regarding 1.24's deprecation

Rakshith-R:
https://github.com/csi-addons/kubernetes-csi-addons/pull/625#discussion_r1696466678

> Sorry again i missed adding one point, we can remove this entire check as we don't need to support 1.24 anymore as its a deprecated or unsupported Kubernetes version. IMO we can remove all these checks

The discussion was marked resolved and therefore not visible in the thread.
I think we can remove the check entirely too.

internal/sidecar/service/reclaimspace.go
psy-q: Ah, that's good news, I will remove the check and the legacy global mount section.

psy-q: I've removed the check. I noticed that the way the function is called, it checks for errors but `err` is always `nil` for `getStagingTargetPath()`. Should I remove the error check as well?

psy-q:
I've noticed [this error checking](https://github.com/csi-addons/kubernetes-csi-addons/blob/db90dbd2016e82a7b6c3e2c43b4968bb09f93a03/internal/sidecar/service/reclaimspace.go#L146) is now pointless since it's always nil.

Should I remove the check as well with this PR or leave that part as it is?

Rakshith-R:
> I've noticed [this error checking](https://github.com/csi-addons/kubernetes-csi-addons/blob/db90dbd2016e82a7b6c3e2c43b4968bb09f93a03/internal/sidecar/service/reclaimspace.go#L146) is now pointless since it's always nil.
> 
> Should I remove the check as well with this PR or leave that part as it is?

Sure, you can remove it too in this PR.

Madhu-1 (approved):

psy-q (commented):
The code in question has been removed.

nixpanic (approved):

Rakshith-R (approved):
Thanks

Rakshith-R:
@mergifyio rebase

Issue #636, opened 2024-08-06
Skip reconcile of PVCs with key rotation annotation


black-dragon74:
#614 added support to consider existing PVCs whenever a SC object is modified. However it requeues the PVCs that have key rotation annotation present.

This issue is to track the patch to skip reconciliation of such PVCs

Issue #626, opened 2024-07-26
it looks like we will skip these pvcs, won't this cause a problem when an annotation is deleted on the storageclass and the controller-added annotation exists on the PVC, IMO we need an owner to identify who added the annotation is it by the controller or not.


Madhu-1:
it looks like we will skip these pvcs, won't this cause a problem when an annotation is deleted on the storageclass and the controller-added annotation exists on the PVC, IMO we need an owner to identify who added the annotation is it by the controller or not.

_Originally posted by @Madhu-1 in https://github.com/csi-addons/kubernetes-csi-addons/pull/614#discussion_r1680814593_

black-dragon74:
Closing this as the new approach for schedule parsing using `state=managed/unmanaged` annotation addresses this.

PR #637, opened 2024-08-06
Skip PVCs with key rotation annotation when SC is changed

black-dragon74:
Existing PVCs with keyoration annotation are enqueued when a SC object is modified.

This patch skips reconcile of such PVCs

Closes: #636

### DIFF
--- a/internal/controller/csiaddons/persistentvolumeclaim_controller.go
+++ b/internal/controller/csiaddons/persistentvolumeclaim_controller.go
@@ -293,7 +293,8 @@ func (r *PersistentVolumeClaimReconciler) determineScheduleAndRequeue(
 // storageClassEventHandler returns an EventHandler that responds to changes
 // in StorageClass objects and generates reconciliation requests for all
 // PVCs associated with the changed StorageClass.
-// PVCs with rsCronJobScheduleTimeAnnotation are not enqueued.
+// PVCs with either rsCronJobScheduleTimeAnnotation or
+// krcJobScheduleTimeAnnotation are not enqueued.
 func (r *PersistentVolumeClaimReconciler) storageClassEventHandler() handler.EventHandler {
 	return handler.EnqueueRequestsFromMapFunc(
 		func(ctx context.Context, obj client.Object) []reconcile.Request {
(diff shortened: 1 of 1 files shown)


internal/controller/csiaddons/persistentvolumeclaim_controller.go
```diff
@@ -317,6 +318,9 @@ func (r *PersistentVolumeClaimReconciler) storageClassEventHandler() handler.Eve
 				if _, ok := pvc.GetAnnotations()[rsCronJobScheduleTimeAnnotation]; ok {
 					continue
 				}
+				if _, ok := pvc.GetAnnotations()[krcJobScheduleTimeAnnotation]; ok {
```
Rakshith-R, line 321: What if 
- A storageclass already has rsCronJob annotation
- PVCs already have rsCronJob annotation
- krcJob annotation is added to that same storageclass

I think we need a more elaborate check here which also considers annotations on chnaged storageclass.

nixpanic:
I think the following does not result in the expected annotation:

1. a StorageClass has the annotation added
2. a PVC gets the annotation because of (1)
3. the annotation in the StorageClass is updated
4. the PVC will **not** get an updated annotation

It would be expected that the PVC in step (4) will get the updated annotation too.

You could add an annotation like `keyrotation.csiaddons.openshift.io/managed-by: storageclass/name-of-the-class` and use that while checking if the annotation needs updating. It would need to take priority of the source of the annotation into account too, as the annotation set on the Namespace should override it.

black-dragon74:
Looking at the edge cases, what if we do not skip the reconcile and let the predicates handle it? We have predicates in place for both SC and PVCs that only reconcile when the value of the relevant annotations has been updated.

@iPraveenParihar WDYT?

iPraveenParihar:
> I think the following does not result in the expected annotation:
> 
> 1. a StorageClass has the annotation added
> 2. a PVC gets the annotation because of (1)
> 3. the annotation in the StorageClass is updated
> 4. the PVC will **not** get an updated annotation
> 
> It would be expected that the PVC in step (4) will get the updated annotation too.
> 
> You could add an annotation like `keyrotation.csiaddons.openshift.io/managed-by: storageclass/name-of-the-class` and use that while checking if the annotation needs updating. It would need to take priority of the source of the annotation into account too, as the annotation set on the Namespace should override it.

I've been working on a similar approach for ReclaimSpace https://github.com/csi-addons/kubernetes-csi-addons/issues/626. I've been occupied with other priority tasks, but I'll get back to it as soon as I have some time.



> Looking at the edge cases, what if we do not skip the reconcile here and delegate the decision to the predicate functions? We have predicates in place for both SC and PVCs that only reconcile when the value of the relevant annotations has been updated.
> 
> @iPraveenParihar WDYT?

@black-dragon74, from SC predicate you can't enqueue PV reconcile requests.

black-dragon74:
> @black-dragon74, from SC predicate you can't enqueue PV reconcile requests.

Yes. I get that. But predicate is run before the reconcile happens, no?

Madhu-1:
@black-dragon74 Is this still required?

black-dragon74:
Closing this one, I will follow this up in another PR with a different approach.

PR #641, opened 2024-08-13
fix pvc reclaimspace & keyrotation annotation filter

iPraveenParihar:
This commit adds a check in StorageClass event handler to filter out PVC not needed for Reconcile.

- If the StorageClass has ReclaimSpace annotation,
   PVCs without ReclaimSpace annotations will be enqueued.
 - If the StorageClass has KeyRotation annotation,
    PVCs without the KeyRotation annotation will be enqueued.

### DIFF
--- a/internal/controller/csiaddons/persistentvolumeclaim_controller.go
+++ b/internal/controller/csiaddons/persistentvolumeclaim_controller.go
@@ -293,7 +293,12 @@ func (r *PersistentVolumeClaimReconciler) determineScheduleAndRequeue(
 // storageClassEventHandler returns an EventHandler that responds to changes
 // in StorageClass objects and generates reconciliation requests for all
 // PVCs associated with the changed StorageClass.
-// PVCs with rsCronJobScheduleTimeAnnotation are not enqueued.
+//
+// PVCs are enqueued for reconciliation if one of the following is true -
+//   - If the StorageClass has ReclaimSpace annotation,
+//     PVCs without ReclaimSpace annotations will be enqueued.
+//   - If the StorageClass has KeyRotation annotation,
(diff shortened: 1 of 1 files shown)


Madhu-1:
@iPraveenParihar please add more details why not to requeue in this specific case which we are addressing here?

iPraveenParihar:
> @iPraveenParihar please add more details why not to requeue in this specific case which we are addressing here?

Updated the PR description.

Madhu-1:
> This commit adds a check in StorageClass event handler to filter out PVC not needed for Reconcile.
> 
> * If the StorageClass has ReclaimSpace annotation,
>   PVCs without ReclaimSpace annotations will be enqueued.
> * If the StorageClass has KeyRotation annotation,
>   PVCs without the KeyRotation annotation will be enqueued.

what will happen if i remove the annotation on the storageclass? i would expect that the annotations on the PVC's will be removed and reclaimspace/keyRotation CR's will get deleted, isn't that the case? have we documented what manual step user need to do in that case?

Madhu-1:
> > This commit adds a check in StorageClass event handler to filter out PVC not needed for Reconcile.
> > 
> > * If the StorageClass has ReclaimSpace annotation,
> >   PVCs without ReclaimSpace annotations will be enqueued.
> > * If the StorageClass has KeyRotation annotation,
> >   PVCs without the KeyRotation annotation will be enqueued.
> 
> what will happen if i remove the annotation on the storageclass? i would expect that the annotations on the PVC's will be removed and reclaimspace/keyRotation CR's will get deleted, isn't that the case? have we documented what manual step user need to do in that case?

i see that its documented for now.

Rakshith-R (approved):

Madhu-1 (approved):

Issue #649, opened 2024-08-21
Some const are unreachable due to moving controllers to internal package


ELENAGER:
In DR code we are using the following:
const (
	ConditionCompleted = "Completed"
	ConditionDegraded  = "Degraded"
	ConditionResyncing = "Resyncing"
)

const (
	Success         = "Success"
	Promoted        = "Promoted"
	Demoted         = "Demoted"
	FailedToPromote = "FailedToPromote"
	FailedToDemote  = "FailedToDemote"
	Error           = "Error"
	VolumeDegraded  = "VolumeDegraded"
	Healthy         = "Healthy"
	ResyncTriggered = "ResyncTriggered"
	FailedToResync  = "FailedToResync"
	NotResyncing    = "NotResyncing"
)
Now the error occurs because the package github.com/csi-addons/kubernetes-csi-addons/internal/controller/replication.storage is inside an internal directory, which means it is designed to be used only by the code within the same module and cannot be imported from outside the module.

Is it possible to add a public interface? Or to move it to utils?

PR #650, opened 2024-08-21
replication: move conditions to api

Madhu-1:
The consumers of the Replication API expect the conditions to be something imported and comparable to the other repos. For the same reason for moving the conditions to replication
types.

fixes: #649


ELENAGER (approved):
LGTM

yati1998 (approved):

Rakshith-R (approved):

PR #655, opened 2024-08-30
deploy: remove extra prefix: csi-addons-

Rakshith-R:
Unnecessary prefix: csi-addons- was added by mistake in default kustomization.yaml in below commit

https://github.com/csi-addons/kubernetes-csi-addons/commit/90c23d244b198b6df9d2350bd1ab80d8e0c1d784#diff-c006b1ef995eeeb57408cce42bd84db8dfc0cbe18df03222857665cebb1ac6fd

### DIFF
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -1,13 +1,6 @@
 # Adds namespace to all resources.
 namespace: csi-addons-system
 
-# Value of this field is prepended to the
-# names of all resources, e.g. a deployment named
-# "wordpress" becomes "alices-wordpress".
-# Note that it should also match with the prefix (text before '-') of the namespace
-# field above.
-namePrefix: csi-addons-
-
 # Labels to add to all resources and selectors.
 #labels:
 #- includeSelectors: true


iPraveenParihar:
@Rakshith-R, remove it from here as well?
https://github.com/csi-addons/kubernetes-csi-addons/blob/3da396a242f247f8b983339a889dcaeee9c69c91/config/manager/kustomization.yaml#L9

iamniting (changes requested):
We should keep this one and remove from 
https://github.com/csi-addons/kubernetes-csi-addons/blob/48f0e109537c65f4feeacaebca2606378ff51187/config/rbac/kustomization.yaml#L10

https://github.com/csi-addons/kubernetes-csi-addons/blob/48f0e109537c65f4feeacaebca2606378ff51187/config/manager/kustomization.yaml#L9

config/default/kustomization.yaml
```diff
@@ -1,13 +1,6 @@
 # Adds namespace to all resources.
 namespace: csi-addons-system
 
-# Value of this field is prepended to the
-# names of all resources, e.g. a deployment named
-# "wordpress" becomes "alices-wordpress".
-# Note that it should also match with the prefix (text before '-') of the namespace
-# field above.
-namePrefix: csi-addons-
```
Madhu-1, line 9: This looks to be in the right place, needs to be removed from the manager and rbac folder but can be revisited later as it will also do the same thing, we need to change few things on how we are doing the Kustomization in makefile

Madhu-1 (approved):

iPraveenParihar (approved):

PR #663, opened 2024-09-11
ci: use docker-generate-protobuf by default in Makefile

Rakshith-R:
This commit modifies Makefile to use docker-generate-protobuf instead of generate-protobuf by default.


Makefile
```diff
@@ -151,7 +151,7 @@ vet: ## Run go vet against code.
 	go vet ./...
 
 .PHONY: test
-test: manifests generate generate-protobuf fmt vet envtest ## Run tests.
+test: manifests generate docker-generate-protobuf fmt vet envtest ## Run tests.
```
Madhu-1, line 154: @Rakshith-R what is the use of this change?

Rakshith-R: > @Rakshith-R what is the use of this change?

It avoids running `generate-protobuf` on the user's system.
`docker-generate-protobuf` takes care of all the dependencies.

`docker-build` calls `test` which in turn calls `generate-protobuf` which is very inconvenient.

Madhu-1: we might get in to the problem when there is a protobuf version update, Currently we are seeing the difference in the CI and pushing the changes as extra commit to keep the consistent pb files for the internal protobuf, if we are planning to change it we need to avoid having this problem please make required changes to `github actions` as well.

Madhu-1: i looked at the Makefile , the above is not required. please take care of fixing the CI problem

black-dragon74:
We need to update protobuf versions inside the tools container as well.

Makefile
Madhu-1: i think problem is docker build running test, why do we need to run test why building the container it should only do the build not test? can you adjust generating the deps when running docker-build

Rakshith-R:
> We need to update protobuf versions inside the tools container as well.

where exactly ?

internal/proto/encryptionkeyrotation.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.34.2
-// 	protoc        v3.20.2
+// 	protoc        v3.19.6
```
Madhu-1, line 4: We should keep this to the same version as we have in GitHub action https://github.com/csi-addons/kubernetes-csi-addons/blob/bd88536de3da60a31dc293c8654ca9eb11589bca/build/Containerfile.tools#L14 is the place we are installing this in fedora or remove installing protobuf in the GitHub action, we need to be consistency

black-dragon74:
> > We need to update protobuf versions inside the tools container as well.
> 
> where exactly ?

Right now the tools container uses the latest available version from Fedora. We should pin the version to that in the spec repo. We could have something like [this](https://github.com/csi-addons/spec/blob/main/release-tools/build.make) for protoc tools.

internal/proto/encryptionkeyrotation.pb.go
nixpanic: I'd prefer to remove the protoc installation in the github action. The CI job calls `make test`, so it should run in the container now too.

https://github.com/csi-addons/kubernetes-csi-addons/blob/main/.github/workflows/test-golang.yaml#L23-L27

nixpanic: Once this is addressed, it should be good to go.

nixpanic (changes requested):

nixpanic:
@Rakshith-R please update this, or let me know if you want me to take it over. Thanks!

nixpanic (approved):

Madhu-1 (approved):

PR #664, opened 2024-09-11
replication: add new Validated condition

Rakshith-R:
This commit adds new Validated condition.
This is initially used to indicate the csi driver
responded with FailedPrecondition grpc code for
EnableReplication request using
`PrerequisiteNotMet` reason.

### DIFF
--- a/internal/controller/replication.storage/status.go
+++ b/internal/controller/replication.storage/status.go
@@ -65,6 +65,40 @@ func setFailedPromotionCondition(conditions *[]metav1.Condition, observedGenerat
 		ObservedGeneration: observedGeneration,
 		Status:             metav1.ConditionFalse,
 	})
+	setStatusCondition(conditions, &metav1.Condition{
+		Type:               v1alpha1.ConditionValidated,
+		Reason:             v1alpha1.PrerequisiteMet,
+		ObservedGeneration: observedGeneration,
+		Status:             metav1.ConditionTrue,
+	})
+}
+
+// sets conditions when volume promotion was failed due to failed validation.
+func setFailedValidationCondition(conditions *[]metav1.Condition, observedGeneration int64) {
+	setStatusCondition(conditions, &metav1.Condition{
+		Type:               v1alpha1.ConditionCompleted,
+		Reason:             v1alpha1.FailedToPromote,
+		ObservedGeneration: observedGeneration,
+		Status:             metav1.ConditionFalse,
+	})
+	setStatusCondition(conditions, &metav1.Condition{
+		Type:               v1alpha1.ConditionDegraded,
+		Reason:             v1alpha1.Error,
+		ObservedGeneration: observedGeneration,
+		Status:             metav1.ConditionTrue,
+	})
+	setStatusCondition(conditions, &metav1.Condition{
+		Type:               v1alpha1.ConditionResyncing,
+		Reason:             v1alpha1.NotResyncing,
+		ObservedGeneration: observedGeneration,
+		Status:             metav1.ConditionFalse,
+	})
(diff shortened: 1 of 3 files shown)


internal/controller/replication.storage/volumereplication_controller.go
```diff
@@ -741,13 +741,19 @@ func (r *VolumeReplicationReconciler) enableReplication(vr *volumeReplicationIns
...
+		return nil
+	}
 
-		return resp.Error
+	vr.logger.Error(resp.Error, "failed to enable volume replication")
+
+	if resp.HasKnownGRPCError(enableReplicationKnownErrors) {
+		setFailedValidationCondition(&vr.instance.Status.Conditions, vr.instance.Generation)
+	} else {
+		setFailedPromotionCondition(&vr.instance.Status.Conditions, vr.instance.Generation)
```
Madhu-1, line 753: can you please use `setFailureCondition(instance)` as it internally takes care of setting the same.

Madhu-1: we already have a generic helper function to set the conditions i don't like to see others calling it even though we have a helper function to do it. and for this again it's a bug we should not be setting `setFailedPromotionCondition` condition if the enable replication fails, the condition should reflect what went wrong.

internal/controller/replication.storage/status.go
```diff
@@ -65,6 +65,40 @@ func setFailedPromotionCondition(conditions *[]metav1.Condition, observedGenerat
 		ObservedGeneration: observedGeneration,
 		Status:             metav1.ConditionFalse,
 	})
+	setStatusCondition(conditions, &metav1.Condition{
```
Madhu-1, line 68: This doesnt look correct, even during the promotion failure there could be a state where the Precondition is also not met.

internal/controller/replication.storage/volumereplication_controller.go
Rakshith-R: > we already have a generic helper function to set the conditions i don't like to see others calling it even though we have a helper function to do it. and for this again it's a bug we should not be setting `setFailedPromotionCondition` condition if the enable replication fails, the condition should reflect what went wrong.

Enabling replication is part of promotion. I don't see a issue with that part.

Why do you want to force a helper function that is used at a lot of other places to fit into a new scenario ?
If we do that, we'll need more arguments and the function will not be so generic anymore.

Madhu-1: even thought its part of the condition we should reflect on what exactly went wrong https://github.com/csi-addons/kubernetes-csi-addons/issues/666 is the tracker for it.

>Why do you want to force a helper function that is used at a lot of other places to fit into a new scenario ?
If we do that, we'll need more arguments and the function will not be so generic anymore.

i don't see any new change in the behaviour to call `setFailureCondition` instead of `setFailedPromotionCondition`

internal/controller/replication.storage/status.go
Rakshith-R: > This doesnt look correct, even during the promotion failure there could be a state where the Precondition is also not met.

Let's fix it when we are verifying all the conditions in the future.
As you already mentioned in your new tracker, not all conditions currently expressed maybe very granular and correct right now.

internal/controller/replication.storage/volumereplication_controller.go
Rakshith-R: I really don't get why you want to call that function when you already know which function it will in turn call.

There are a lot of places already calling functions from status.go directly.

Madhu-1 (approved):

iPraveenParihar (approved):

PR #676, opened 2024-09-18
Use protobuf encoding for core K8s apis

Nikhil-Ladha:
For core K8s API objects like Pods, Nodes, etc., we can use protobuf encoding which reduces CPU consumption related to (de)serialization, reduces overall latency of the API call, reduces memory footprint, reduces the amount of work performed by the GC and results in quicker propagation of objects to event handlers of shared informers.

### DIFF
--- a/cmd/csi-addons/main.go
+++ b/cmd/csi-addons/main.go
@@ -25,6 +25,7 @@ import (
 
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 )
@@ -121,6 +122,8 @@ func getKubernetesClient() *kubernetes.Clientset {
 	if err != nil {
 		panic(err.Error())
 	}
+
+	config.ContentType = runtime.ContentTypeProtobuf
 
 	clientset, err := kubernetes.NewForConfig(config)
 	if err != nil {


Nikhil-Ladha:
> From: https://kubernetes.io/docs/reference/using-api/api-concepts/#alternate-representations-of-resources:
> 
> > Not all API resource types support Protobuf; specifically, Protobuf isn't available for resources that are defined as [CustomResourceDefinitions](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/) or are served via the [aggregation layer](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/apiserver-aggregation/).
> 
> Does it work for CRDs that we have?

Right, it doesn't work for CRDs.
We can't use it for the csi-addons controller code, will have to update and check the pod deployment once.

Nikhil-Ladha:
Updated it to add the protobuf encoding only for CLI tool interactions after testing, in other places we have use the client to interact with CRDs.

cmd/csi-addons/main.go
```diff
@@ -122,6 +123,8 @@ func getKubernetesClient() *kubernetes.Clientset {
 		panic(err.Error())
 	}
 
+	config.ContentType = runtime.ContentTypeProtobuf
```
Madhu-1, line 126: can you please check again, i see we are creating client in 3 places https://github.com/search?q=repo%3Acsi-addons%2Fkubernetes-csi-addons%20kubernetes.NewForConfig&type=code

Nikhil-Ladha: Yes, that's what I said above that in other 2 places the client is used to interact with CRDs. That's the reason we can't use the protobuf encoding.

Madhu-1 (approved):

black-dragon74 (approved):

PR #683, opened 2024-09-24
rebase: use toolchain for 1.22.6

Madhu-1:
we dont have downstream golang with version 1.22.6 and csi-addons needs 1.22.6 to build for that reason we can use the toolchain to enforce the soft dependency.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,8 @@
 module github.com/csi-addons/kubernetes-csi-addons
 
-go 1.22.6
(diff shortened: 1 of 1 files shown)


Madhu-1:
I was able to build the dependency projects (Rook) with a lower version instead of the hard 1.22.6 requirement.

iPraveenParihar (approved):

nixpanic (approved):

PR #685, opened 2024-09-24
controller/pvc: add checks for nil/empty pvc.Spec.StorageClassName

Rakshith-R:
`pvc.Spec.StorageClassName` is a string pointer and can be nil or empty. This commit adds check for both nil and empty.

### DIFF
--- a/internal/controller/csiaddons/persistentvolumeclaim_controller_test.go
+++ b/internal/controller/csiaddons/persistentvolumeclaim_controller_test.go
@@ -323,6 +323,14 @@ func TestDetermineScheduleAndRequeue(t *testing.T) {
 		assert.Equal(t, "", schedule)
 	})
 
+	// test for StorageClassName is nil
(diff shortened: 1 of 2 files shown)


iPraveenParihar (dismissed):
Thanks!

internal/controller/csiaddons/persistentvolumeclaim_controller_test.go
```diff
@@ -323,6 +323,15 @@ func TestDetermineScheduleAndRequeue(t *testing.T) {
 		assert.Equal(t, "", schedule)
 	})
 
+	// test for StorageClassName is nil
+	t.Run("StorageClassName is nil", func(t *testing.T) {
+		sc.Name = "non-existent-sc"
```
Madhu-1, line 328: Can you pleas remove this line, This is not required for this testcase

Madhu-1 (approved):

iPraveenParihar (approved):

PR #691, opened 2024-10-15
update the vr condition

yati1998:
this commit updates the vr conditions to
include descriptive message for every
operations


internal/controller/replication.storage/status.go
```diff
@@ -76,24 +83,28 @@ func setFailedPromotionCondition(conditions *[]metav1.Condition, observedGenerat
...
 	})
 	setStatusCondition(conditions, &metav1.Condition{
+		Message:            "error detected while promotion",
 		Type:               v1alpha1.ConditionDegraded,
 		Reason:             v1alpha1.Error,
 		ObservedGeneration: observedGeneration,
 		Status:             metav1.ConditionTrue,
 	})
 	setStatusCondition(conditions, &metav1.Condition{
+		Message:            "Volume is resynced",
```
nixpanic, line 100: _resynced_ in the message, type is _resyncing_?

internal/controller/replication.storage/status.go
```diff
@@ -48,24 +51,28 @@ func setPromotedCondition(conditions *[]metav1.Condition, observedGeneration int
...
```
nixpanic, line 68: _resynced_ in the message, type is _resyncing_?

internal/controller/replication.storage/status.go
```diff
@@ -48,24 +51,28 @@ func setPromotedCondition(conditions *[]metav1.Condition, observedGeneration int
 // sets conditions when volume promotion was failed.
 func setFailedPromotionCondition(conditions *[]metav1.Condition, observedGeneration int64) {
 	setStatusCondition(conditions, &metav1.Condition{
+		Message:            "Volume failed to promote",
```
nixpanic, line 54: Make error/failed cases a little more clear by starting the message with _failed_ or _error_.

internal/controller/replication.storage/status.go
```diff
@@ -104,12 +115,14 @@ func setFailedValidationCondition(conditions *[]metav1.Condition, observedGenera
...
 func setNotDegradedCondition(conditions *[]metav1.Condition, observedGeneration int64) {
 	setStatusCondition(conditions, &metav1.Condition{
+		Message:            "Volume is demoted",
 		Type:               v1alpha1.ConditionDegraded,
 		Reason:             v1alpha1.Healthy,
 		ObservedGeneration: observedGeneration,
 		Status:             metav1.ConditionFalse,
 	})
 	setStatusCondition(conditions, &metav1.Condition{
+		Message:            "volume is resynced",
```
nixpanic, line 125: mostly you have **V**olume with a capital, not here?

internal/controller/replication.storage/status.go
```diff
@@ -120,18 +133,21 @@ func setNotDegradedCondition(conditions *[]metav1.Condition, observedGeneration
...
 func setDemotedCondition(conditions *[]metav1.Condition, observedGeneration int64) {
 	setStatusCondition(conditions, &metav1.Condition{
+		Message:            "Volume is demoted",
 		Type:               v1alpha1.ConditionCompleted,
 		Reason:             v1alpha1.Demoted,
 		ObservedGeneration: observedGeneration,
 		Status:             metav1.ConditionTrue,
 	})
 	setStatusCondition(conditions, &metav1.Condition{
+		Message:            "Volume is Degraded",
```
nixpanic, line 143: Mostly the state is written lowercase. Make sure you use the same formatting everywhere (whichever you prefer).

internal/controller/replication.storage/status.go
```diff
@@ -142,18 +158,21 @@ func setDemotedCondition(conditions *[]metav1.Condition, observedGeneration int6
 // sets conditions when volume demotion was failed.
 func setFailedDemotionCondition(conditions *[]metav1.Condition, observedGeneration int64) {
 	setStatusCondition(conditions, &metav1.Condition{
+		Message:            "Failed to demote",
```
nixpanic, line 161: similar here, suddenly **F**ailed starts with a CAPITAL

nixpanic:
It would be good to have an example in the PR description. Ideally do not abbreviate things like "vr" in commit subjects.

internal/controller/replication.storage/status.go
yati1998: the type says resyncing, but the time we see the status, the volume would be resynced right?
Will be good to have more clearity.
cc @Madhu-1 @ShyamsundarR can you please add your point here?

nixpanic (dismissed):
Thanks!

internal/controller/replication.storage/status.go
```diff
@@ -26,18 +26,21 @@ import (
...
 	})
 	setStatusCondition(conditions, &metav1.Condition{
+		Message:            "Volume is Healthy",
 		Type:               v1alpha1.ConditionDegraded,
 		Reason:             v1alpha1.Healthy,
 		ObservedGeneration: observedGeneration,
 		Status:             metav1.ConditionFalse,
 	})
 	setStatusCondition(conditions, &metav1.Condition{
+		Message:            "Volume is not resyncing",
```
iPraveenParihar, line 43: the action/state in the message field are inconsistent in terms of capitalization. It would be better to follow the same everywhere? either capitalize or leave them in lowercase. WDYT?
```suggestion
		Message:            "Volume is not Resyncing",
```

yati1998: all of them seems to be in lower case already?
Am I missing it somewhere?

yati1998: ohh I see, couple of states were capital, modified them.

internal/controller/replication.storage/status.go
```diff
@@ -186,18 +208,21 @@ func setResyncCondition(conditions *[]metav1.Condition, observedGeneration int64
...
 	})
 	setStatusCondition(conditions, &metav1.Condition{
+		Message:            "error detected while resyncing",
 		Type:               v1alpha1.ConditionDegraded,
 		Reason:             v1alpha1.Error,
 		ObservedGeneration: observedGeneration,
 		Status:             metav1.ConditionTrue,
 	})
 	setStatusCondition(conditions, &metav1.Condition{
+		Message:            "failed to resync",
```
Madhu-1, line 225: can we define a const/type for all these messages?

nixpanic: @ShyamsundarR please have a look and share your opinion here.

yati1998:
@ShyamsundarR I have updated the pr to include specific errors for failed conditions, do take a look

internal/controller/replication.storage/status.go
```diff
@@ -26,18 +26,21 @@ import (
...
 func setPromotedCondition(conditions *[]metav1.Condition, observedGeneration int64) {
 	setStatusCondition(conditions, &metav1.Condition{
+		Message:            "Volume is promoted",
 		Type:               v1alpha1.ConditionCompleted,
 		Reason:             v1alpha1.Promoted,
 		ObservedGeneration: observedGeneration,
 		Status:             metav1.ConditionTrue,
 	})
 	setStatusCondition(conditions, &metav1.Condition{
+		Message:            "Volume is healthy",
```
nirs, line 36: This message does not add enough info toe understand the meaning of the condition.

internal/controller/replication.storage/status.go
```diff
@@ -26,18 +26,21 @@ import (
...
 		Type:               v1alpha1.ConditionDegraded,
 		Reason:             v1alpha1.Healthy,
 		ObservedGeneration: observedGeneration,
 		Status:             metav1.ConditionFalse,
 	})
 	setStatusCondition(conditions, &metav1.Condition{
+		Message:            "Volume is not resyncing",
```
nirs, line 43: Same here, it is jut the text version of NotResynching or Resynching=false. The text should explain the status.

internal/controller/replication.storage/status.go
```diff
@@ -46,26 +49,30 @@ func setPromotedCondition(conditions *[]metav1.Condition, observedGeneration int
 }
 
 // sets conditions when volume promotion was failed.
-func setFailedPromotionCondition(conditions *[]metav1.Condition, observedGeneration int64) {
+func setFailedPromotionCondition(conditions *[]metav1.Condition, observedGeneration int64, err string) {
```
nirs, line 52: err is not a good name for an error message, this is the common name for objects implementing the error interface. Maybe use `errorMesssage`?

internal/controller/replication.storage/status.go
```diff
@@ -46,26 +49,30 @@ func setPromotedCondition(conditions *[]metav1.Condition, observedGeneration int
...
 	})
 	setStatusCondition(conditions, &metav1.Condition{
+		Message:            "error detected while promotion",
 		Type:               v1alpha1.ConditionDegraded,
 		Reason:             v1alpha1.Error,
 		ObservedGeneration: observedGeneration,
 		Status:             metav1.ConditionTrue,
 	})
 	setStatusCondition(conditions, &metav1.Condition{
+		Message:            "Volume is not resyncing",
```
nirs, line 68: Why are we repeating the same value multiple times? If we need to repeat the same message, they should be constants.

internal/controller/replication.storage/status.go
```diff
@@ -46,26 +49,30 @@ func setPromotedCondition(conditions *[]metav1.Condition, observedGeneration int
...
+func setFailedPromotionCondition(conditions *[]metav1.Condition, observedGeneration int64, err string) {
 	setStatusCondition(conditions, &metav1.Condition{
+		Message:            err,
 		Type:               v1alpha1.ConditionCompleted,
 		Reason:             v1alpha1.FailedToPromote,
 		ObservedGeneration: observedGeneration,
 		Status:             metav1.ConditionFalse,
 	})
 	setStatusCondition(conditions, &metav1.Condition{
+		Message:            "error detected while promotion",
```
nirs, line 61: It is strange that we set the standard message for this condition, but custom message for the previous condition. Based on the function name this should affect only the PromotedCondition.

internal/controller/replication.storage/status.go
```diff
@@ -76,24 +83,28 @@ func setFailedPromotionCondition(conditions *[]metav1.Condition, observedGenerat
 // sets conditions when volume promotion was failed due to failed validation.
 func setFailedValidationCondition(conditions *[]metav1.Condition, observedGeneration int64) {
 	setStatusCondition(conditions, &metav1.Condition{
+		Message:            "failed to promote",
```
nirs, line 86: This message is not helpful - why did it fail? what is the the way to recover from the error?

In ramen we want to propagate the errors to the ramen resources conditions. If we include good error message in the ramen resource debugging issues will be come much easier, since the error messages would propagate to the DRPC on the hub.

But in this case the error is actually wrong - since the issue is failing initial validation. We did not try to promote and did not fail to promote since promotion is not possible.

Maybe use: "Promotion not possible due to failed precondition"

internal/controller/replication.storage/status.go
```diff
@@ -76,24 +83,28 @@ func setFailedPromotionCondition(conditions *[]metav1.Condition, observedGenerat
...
 		Type:               v1alpha1.ConditionCompleted,
 		Reason:             v1alpha1.FailedToPromote,
 		ObservedGeneration: observedGeneration,
 		Status:             metav1.ConditionFalse,
 	})
 	setStatusCondition(conditions, &metav1.Condition{
+		Message:            "error detected while promotion",
```
nirs, line 93: This condition is not related to promotion. Why do we need to talk about promotion here?

internal/controller/replication.storage/status.go
```diff
@@ -76,24 +83,28 @@ func setFailedPromotionCondition(conditions *[]metav1.Condition, observedGenerat
...
 		Type:               v1alpha1.ConditionDegraded,
 		Reason:             v1alpha1.Error,
 		ObservedGeneration: observedGeneration,
 		Status:             metav1.ConditionTrue,
 	})
 	setStatusCondition(conditions, &metav1.Condition{
+		Message:            "Volume is not resyncing",
```
nirs, line 100: This message is capitalised, while other are not. I'm not sure what is the wanted format but all messages must be consistent.

internal/controller/replication.storage/status.go
```diff
@@ -76,24 +83,28 @@ func setFailedPromotionCondition(conditions *[]metav1.Condition, observedGenerat
...
 		Type:               v1alpha1.ConditionResyncing,
 		Reason:             v1alpha1.NotResyncing,
 		ObservedGeneration: observedGeneration,
 		Status:             metav1.ConditionFalse,
 	})
 	setStatusCondition(conditions, &metav1.Condition{
+		Message:            "failed to meet prerequisite",
```
nirs, line 107: Good summary of the error but we really want the reason that will help the admin debug this problem. What is the actual issue?

internal/controller/replication.storage/volumereplication_controller.go
```diff
@@ -124,7 +124,7 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	err = validatePrefixedParameters(vrcObj.Spec.Parameters)
 	if err != nil {
 		logger.Error(err, "failed to validate parameters of volumeReplicationClass", "VRCName", instance.Spec.VolumeReplicationClass)
-		setFailureCondition(instance)
+		setFailureCondition(instance, "failed to validate parameters of volumeReplicationClass")
```
nirs, line 127: Start to look helpful, but which parameter was invalid?

internal/controller/replication.storage/volumereplication_controller.go
```diff
@@ -160,7 +160,7 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		pvc, pv, pvErr = r.getPVCDataSource(logger, nameSpacedName)
 		if pvErr != nil {
 			logger.Error(pvErr, "failed to get PVC", "PVCName", instance.Spec.DataSource.Name)
-			setFailureCondition(instance)
+			setFailureCondition(instance, "failed to get PVC")
```
nirs, line 163: failed to find the PVC? "Get" is a technical term, the function used to read an object from the api server, the human term is finding or looking up.

But it is not clear if the issue is missing pvc (ErrNotExist) or an error accessing the api server (maybe temporary error?). When we fix the code to handle both case we can provide a more useful error message.

nirs (commented):
All the messages are very short and not descriptive. The VR .Status.Message looks like good error text (maybe too verbose sometimes).

Do we assume that users will use .Status.Message for reporting errors? How do we know which condition is related to .Status.Message?

I think we need to include .Status.Message in the relevant .Conditions[N].Message instead of the short and non descriptive messages.

Ramen looks only at the VR conditions, and it will be very easy to use if we can use the condition message in the ramen condition. I don't see how we can use the global .Status.Message.

internal/controller/replication.storage/volumereplication_controller.go
```diff
@@ -112,7 +112,7 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	// Get VolumeReplicationClass
 	vrcObj, err := r.getVolumeReplicationClass(logger, instance.Spec.VolumeReplicationClass)
 	if err != nil {
-		setFailureCondition(instance)
+		setFailureCondition(instance, "filed to get volumeReplication class")
```
am-agrawa, line 115:
```suggestion
		setFailureCondition(instance, "failed to get volumeReplication class")
```

internal/controller/replication.storage/status.go
yati1998: since this is a sucessful condition, we dont have any information to be printed, VR is in healthy state. Can you please share some examples of what other information you want to get printed here?

internal/controller/replication.storage/status.go
yati1998: same explanation as above, for this case, resync is not trigerred, hence volume is not resyncing, not sure what would be the detailed status for this?

internal/controller/replication.storage/status.go
yati1998: yes, this condition gets excuted when the precondition fails due to initial validation and as mentioned earlier, the reason for error will be reflected in status.message. yes we can use the above msg, i will update it thanks

internal/controller/replication.storage/status.go
yati1998: this condition is reached while we were trying to promote and hence the error was detected right?

internal/controller/replication.storage/status.go
yati1998: ohh missed this, will update

internal/controller/replication.storage/status.go
yati1998: that will be in status.message

yati1998:
The following comments have been addressed in the above PR:
1. Error message from ceph csi is included in conditions of VR so that it can be picked up by VRG
2.  Error messages have been make const

The VR after this change looks like below:
```
Status:
  Conditions:
    Last Transition Time:  2024-10-15T04:21:29Z
    Message:              failed to demote 
    Observed Generation:   3
    Reason:                FailedToDemote
    Status:                False
    Type:                  Completed
    Last Transition Time:  2024-10-15T04:21:29Z
    Message:          Volume is degraded     
    Observed Generation:   3
    Reason:                volume 0001-0009-rook-ceph-0000000000000002-07d77e2c-5689-4493-b583-cf06c92e1db1 not found: Failed as image not found (internal RBD image not found)
    Status:                True
    Type:                  Degraded
    Last Transition Time:  2024-10-15T04:19:17Z
    Message:           volume is not resyncing
    Observed Generation:   3
    Reason:                NotResyncing
    Status:                False
    Type:                  Resyncing
  Last Completion Time:    2024-10-15T04:19:18Z
  Message:                 volume 0001-0009-rook-ceph-0000000000000002-07d77e2c-5689-4493-b583-cf06c92e1db1 not found: Failed as image not found (internal RBD image not found)
  Observed Generation:     3
  State:                   Primary
Events:                    <none>
```

cc @nirs @Madhu-1 @nixpanic @ShyamsundarR  please add your reviews to the changes

nirs:
> this commit updates the Volume Replication(VR) conditions to include descriptive message for every operations
> 
> Example of the VR status:
> 
> ```
> Status:
>   Conditions:
>     Last Transition Time:  2024-10-15T04:21:29Z
>     Message:              failed to demote 
>     Observed Generation:   3
>     Reason:                FailedToDemote
>     Status:                False
>     Type:                  Completed
>     Last Transition Time:  2024-10-15T04:21:29Z
>     Message:          Volume is degraded     
>     Observed Generation:   3
>     Reason:                Error
>     Status:                True
>     Type:                  Degraded
>     Last Transition Time:  2024-10-15T04:19:17Z
>     Message:            failed to resync   
>     Observed Generation:   3
>     Reason:                NotResyncing
>     Status:                False
>     Type:                  Resyncing
>   Last Completion Time:    2024-10-15T04:19:18Z
>   Message:                 volume 0001-0009-rook-ceph-0000000000000002-07d77e2c-5689-4493-b583-cf06c92e1db1 not found: Failed as image not found (internal RBD image not found)
>   Observed Generation:     3
>   State:                   Primary
> Events:                    <none>
> ```

This example does not reflect the actual change.

api/replication.storage/v1alpha1/volumereplication_types.go
```diff
@@ -36,6 +36,25 @@ const (
...
 )
 
+// These are valid messages for various
+// conditions and states of Volume
+// replication.
+const (
+	MessagePromoted               = "volume is promoted"
+	MessageHealthy                = "volume is healthy"
+	MessageNotResyncing           = "volume is not resyncing"
+	MessagePromotionErrorDetected = "error detected while promotion"
```
nirs, line 46: Not clear what does this mean - promotion failed? promotion succeeded with warnings that are safe to ignore?

Same for other "errors detected".

api/replication.storage/v1alpha1/volumereplication_types.go
```diff
@@ -36,6 +36,25 @@ const (
 	ConditionValidated = "Validated"
 )
 
+// These are valid messages for various
+// conditions and states of Volume
+// replication.
+const (
+	MessagePromoted               = "volume is promoted"
+	MessageHealthy                = "volume is healthy"
```
nirs, line 44: This is not verbose enough, not clear what is "healthy".

api/replication.storage/v1alpha1/volumereplication_types.go
```diff
@@ -36,6 +36,25 @@ const (
 	ConditionValidated = "Validated"
 )
 
+// These are valid messages for various
+// conditions and states of Volume
+// replication.
+const (
+	MessagePromoted               = "volume is promoted"
```
nirs, line 43: This does not explain more than Promoted=true. Maybe something like "volume promoted to primary and replicating to secondary volume"

api/replication.storage/v1alpha1/volumereplication_types.go
```diff
@@ -36,6 +36,25 @@ const (
...
+	MessageNotResyncing           = "volume is not resyncing"
+	MessagePromotionErrorDetected = "error detected while promotion"
+	MessageDemotionErrorDetected  = "error detected while demotion"
+	MessageResyncErrorDetected    = "error detected while resyncing"
+	MessageValidated              = "volume is validated"
+	MessageFailedPromoted         = "failed to promote volume"
+	MessageFailedPreCondition     = "failed to meet prerequisite"
+	MessageDemoted                = "volume is demoted"
+	MessageDegraded               = "volume is degraded"
+	MessageResyncTriggered        = "resync is triggered"
```
nirs, line 54: rsync was triggered? a more verbose message would be more clear, like "volume is resyncing changes from ... to ..."

internal/controller/replication.storage/status.go
```diff
@@ -46,26 +49,30 @@ func setPromotedCondition(conditions *[]metav1.Condition, observedGeneration int
...
 		Type:               v1alpha1.ConditionCompleted,
 		Reason:             v1alpha1.FailedToPromote,
 		ObservedGeneration: observedGeneration,
 		Status:             metav1.ConditionFalse,
 	})
 	setStatusCondition(conditions, &metav1.Condition{
+		Message:            v1alpha1.MessagePromotionErrorDetected,
 		Type:               v1alpha1.ConditionDegraded,
-		Reason:             v1alpha1.Error,
+		Reason:             errFromCephCSI,
```
nirs, line 63: types.Condition is documenting the Reason and Message:

```go
  	// reason contains a programmatic identifier indicating the reason for the condition's last transition.
	// Producers of specific condition types may define expected values and meanings for this field,
	// and whether the values are considered a guaranteed API.
	// The value should be a CamelCase string.
	// This field may not be empty.
	// +required
	// +kubebuilder:validation:MaxLength=1024
	// +kubebuilder:validation:MinLength=1
	// +kubebuilder:validation:Pattern=`^[A-Za-z]([A-Za-z0-9_,:]*[A-Za-z0-9_])?$`
	Reason string `json:"reason" protobuf:"bytes,5,opt,name=reason"`
	// message is a human readable message indicating details about the transition.
	// This may be an empty string.
	// +required
	// +kubebuilder:validation:MaxLength=32768
	Message string `json:"message" protobuf:"bytes,6,opt,name=message"`
```

Using Reason for the details of the message contradicts these docs. This means that our API will not behave like most other APIs, and will be harder to use.

This is also harder to use - ramen will have to read both .Mesage and .Reason, and construct a new mesage from both, handling all cases (message empty, reason empty, both empty, ...). It is much easier to use when we can copy the message from the condition as is.

I think we should do instead:

```go
    Message: fmt.Sprintf("Failed to promote: %s", errFromCephCSI)
```

The "Failed to promote: %s" part can be a constant, helper developers use consistent messages and making changes easy. 

See also https://github.com/evankanderson/k8s-community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties

nirs (commented):
Looks much better now with the constants!

But I think using .Reason for the details is wrong, and we already discussed this and agreed that the details will be in the .Message field.

api/replication.storage/v1alpha1/volumereplication_types.go
yati1998: it's just a small message indicating that there is some error, and the error will be printed in the reason

yati1998: the promotion failed is mentioned the the condition type promote, I think if we add same message to volume degradation condition it will not make much sense.

api/replication.storage/v1alpha1/volumereplication_types.go
yati1998: this just states that volume is in healthy state, not sure what other information you want to be printed here. @Madhu-1 any suggestions from your end?

internal/controller/replication.storage/status.go
```diff
@@ -46,26 +50,30 @@ func setPromotedCondition(conditions *[]metav1.Condition, observedGeneration int
 }
 
 // sets conditions when volume promotion was failed.
-func setFailedPromotionCondition(conditions *[]metav1.Condition, observedGeneration int64) {
+func setFailedPromotionCondition(conditions *[]metav1.Condition, observedGeneration int64, errMessage string, errFromCephCSI string) {
```
black-dragon74, line 53:
```suggestion
func setFailedPromotionCondition(conditions *[]metav1.Condition, observedGeneration int64, errMessage, errFromCephCSI string) {
```

As both types are `string`, there is no need to repeat it.

internal/controller/replication.storage/status.go
```diff
@@ -140,20 +157,23 @@ func setDemotedCondition(conditions *[]metav1.Condition, observedGeneration int6
 }
 
 // sets conditions when volume demotion was failed.
-func setFailedDemotionCondition(conditions *[]metav1.Condition, observedGeneration int64) {
+func setFailedDemotionCondition(conditions *[]metav1.Condition, observedGeneration int64, errMessage string, errFromCephCSI string) {
```
black-dragon74, line 160: Same here.

internal/controller/replication.storage/volumereplication_controller.go
```diff
@@ -805,14 +805,14 @@ func getCurrentReplicationState(instance *replicationv1alpha1.VolumeReplication)
 	return instance.Status.State
 }
 
-func setFailureCondition(instance *replicationv1alpha1.VolumeReplication) {
+func setFailureCondition(instance *replicationv1alpha1.VolumeReplication, errMessage string, errFromCephCSI string) {
```
black-dragon74, line 808: And this one too.

internal/controller/replication.storage/status.go
Madhu-1:
```suggestion
func setFailedPromotionCondition(conditions *[]metav1.Condition, observedGeneration int64, msg, msgFromDriver string) {
```

internal/controller/replication.storage/status.go
```diff
@@ -76,24 +84,28 @@ func setFailedPromotionCondition(conditions *[]metav1.Condition, observedGenerat
...
 	})
 	setStatusCondition(conditions, &metav1.Condition{
+		Message:            v1alpha1.MessageNotResyncing,
 		Type:               v1alpha1.ConditionResyncing,
 		Reason:             v1alpha1.NotResyncing,
 		ObservedGeneration: observedGeneration,
 		Status:             metav1.ConditionFalse,
 	})
 	setStatusCondition(conditions, &metav1.Condition{
+		Message:            v1alpha1.MessageFailedPreCondition,
```
Madhu-1, line 108: dont we need to get the error/message from the driver added to this failed preCondition?

internal/controller/replication.storage/status.go
Madhu-1:
```suggestion
func setFailedDemotionCondition(conditions *[]metav1.Condition, observedGeneration int64, msg, msgFromDriver string)
```

api/replication.storage/v1alpha1/volumereplication_types.go
```diff
@@ -36,6 +36,24 @@ const (
 	ConditionValidated = "Validated"
 )
 
+// These are valid messages for various
+// conditions and states of Volume
+// replication.
```
Madhu-1, line 41:
```suggestion
// These are valid messages for various conditions and states of volume replication.
```

lets make it one line comment

api/replication.storage/v1alpha1/volumereplication_types.go
```diff
@@ -36,6 +36,24 @@ const (
...
+	MessageValidated          = "volume is validated and met all prerequisites"
+	MessageFailedPromoted     = "failed to promote volume"
+	MessageFailedDemoted      = "failed to demote volume"
+	MessageFailedPreCondition = "failed to meet prerequisite"
+	MessageDemoted            = "volume is demoted to secondary"
+	MessageDegraded           = "volume is degraded"
+	MessageResyncTriggered    = "volume is resyncing changes from primary to secondary"
+	MessageResyncFailed       = "failed to resync"
+	MessageEnableVolumeFailed = "failed to enable volume replication"
+)
```
Madhu-1, line 55: These messages should also work fine for volume groups as well, Instead of Volume can we call it as `source` so that we don't need to change the messages if it's for volume or volume group.

yati1998: sounds good to me, will update

yati1998: I think lets take opinion from @nirs as well on how this messages are used in VGR, cause AFSIK VGR do add messages based on all the VR conditions.

nirs: I think when using volume group replication (VGR) we will have one VR per PVC, so nothing should change in the VR. @ELENAGER do you want review this?

internal/controller/replication.storage/status.go
```diff
@@ -46,26 +50,30 @@ func setPromotedCondition(conditions *[]metav1.Condition, observedGeneration int
...
+func setFailedPromotionCondition(conditions *[]metav1.Condition, observedGeneration int64, message, messageFromDriver string) {
 	setStatusCondition(conditions, &metav1.Condition{
+		Message:            message,
 		Type:               v1alpha1.ConditionCompleted,
 		Reason:             v1alpha1.FailedToPromote,
 		ObservedGeneration: observedGeneration,
 		Status:             metav1.ConditionFalse,
 	})
 	setStatusCondition(conditions, &metav1.Condition{
+		Message:            fmt.Sprintf("%s: %s", v1alpha1.MessageFailedPromoted, messageFromDriver),
```
nirs, line 62: It is little bit confusing - we have:
- message
- messageFromDriver

It seems that message is related to Completed condition, and messageFromDriver is related to DegradtedCondition. But the name do not give any hint about these relations.

nirs (commented):
Looks like we create useful message now, but it is not clear to me when and why messageFromDriver apply to which condition.

Can you share example VR status in all the error cases? It can be simulated by faking failures, or reproducing failure conditions on a real system (e.g. trying to replicate a cloned pvc with the wrong dr policy).

If you provide an image, I can test this in ramen testing env.

yati1998:
> Looks like we create useful message now, but it is not clear to me when and why messageFromDriver apply to which condition.
> 
> Can you share example VR status in all the error cases? It can be simulated by faking failures, or reproducing failure conditions on a real system (e.g. trying to replicate a cloned pvc with the wrong dr policy).
> 
> If you provide an image, I can test this in ramen testing env.

sure, let me create an image and share with you

internal/controller/replication.storage/status.go
yati1998: message is any error message that is being detected in csiaddons, and messagefromdriver is the rpc msgs that comes from ceph csi driver

internal/controller/replication.storage/status.go
```diff
@@ -74,26 +82,30 @@ func setFailedPromotionCondition(conditions *[]metav1.Condition, observedGenerat
...
 	})
 	setStatusCondition(conditions, &metav1.Condition{
+		Message:            v1alpha1.MessageNotResyncing,
 		Type:               v1alpha1.ConditionResyncing,
 		Reason:             v1alpha1.NotResyncing,
 		ObservedGeneration: observedGeneration,
 		Status:             metav1.ConditionFalse,
 	})
 	setStatusCondition(conditions, &metav1.Condition{
+		Message:            fmt.Sprintf("%s: %s", v1alpha1.MessageFailedPreCondition, messageFromDriver),
```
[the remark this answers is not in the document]
yati1998: it can't be for same conditions, cause there are cases like for failed promote, the message will contain the reason for failure to promote, while in case of resync, it will printed in resync part.  based on which message belongs where, I have added it.

api/replication.storage/v1alpha1/volumereplication_types.go
ELENAGER: As far as I understand, when using volume group replication, CSI is creating one VR per PVC internally, is that right? And ramen is reading status and all the info from VGR

Madhu-1: Nope, we are going to have single VR created for a volumegroupReplication as the datasource, its one per group in case of group replication

nirs: ok, so the message should be:

- volumes: "failed to {action} volume: {details}"
- volume groups: "failed to {action} volume group: {details}"

nirs:
I install the image provided by @yati1998 using this:
https://github.com/RamenDR/ramen/pull/1639

It passed ramen e2e job:
https://github.com/RamenDR/ramen/actions/runs/11683242163/job/32532029150

I'm testing the message for validation issue now.

ShyamsundarR:
Changes to reflect message in condition to include errors from the RPCs is good, moves it closer to the condition that faced the error.

This still does not catch certain errors as @yati1998 pointed out, e.g [MirrorEnable](https://github.com/ceph/go-ceph/blob/a0de92354f217f516ec41374c5c485b80b41cfd4/rbd/mirror.go#L141-L153) returns the -22 and no description, so we miss the context further that the error was actually something like this:

```
$ rbd mirror image enable ocs-storagecluster-cephblockpool/csi-vol-7c561e23-151b-4f8d-a219-41abd252d713 snapshot 
2024-10-25T13:56:25.349+0000 7f3006be3c00 -1 librbd::api::Mirror: image_enable: image has a parent, snapshot based mirroring is not supported
``` 

Which is being addressed in https://github.com/ceph/ceph-csi/pull/4941 we may need other corner cases as it may arise to report better failures/errors. Otherwise the move to report known errors and failures in messages with this PR looks good.

(not adding a +1 as I am not looking at specific of the change overall)

yati1998:
> Changes to reflect message in condition to include errors from the RPCs is good, moves it closer to the condition that faced the error.
> 
> This still does not catch certain errors as @yati1998 pointed out, e.g [MirrorEnable](https://github.com/ceph/go-ceph/blob/a0de92354f217f516ec41374c5c485b80b41cfd4/rbd/mirror.go#L141-L153) returns the -22 and no description, so we miss the context further that the error was actually something like this:
> 
> ```
> $ rbd mirror image enable ocs-storagecluster-cephblockpool/csi-vol-7c561e23-151b-4f8d-a219-41abd252d713 snapshot 
> 2024-10-25T13:56:25.349+0000 7f3006be3c00 -1 librbd::api::Mirror: image_enable: image has a parent, snapshot based mirroring is not supported
> ```
> 
> Which is being addressed in [ceph/ceph-csi#4941](https://github.com/ceph/ceph-csi/pull/4941) we may need other corner cases as it may arise to report better failures/errors. Otherwise the move to report known errors and failures in messages with this PR looks good.
> 
> (not adding a +1 as I am not looking at specific of the change overall)

Hey @ShyamsundarR  this error was updated further, I think you are using older version of cephcsi, due to which it seems just the error code is returned. https://github.com/ceph/ceph-csi/pull/4678/files#diff-d3713dcbb075d877c69230700478ec60c5fe029d742dae0f72499d9f6833617f

ShyamsundarR:
> Hey @ShyamsundarR this error was updated further, I think you are using older version of cephcsi, due to which it seems just the error code is returned. https://github.com/ceph/ceph-csi/pull/4678/files#diff-d3713dcbb075d877c69230700478ec60c5fe029d742dae0f72499d9f6833617f

Quite possible, the builds in use were downstream and possibly ones in release-4.16. Overall, the new PR to improve the error message from Ceph-CSI is helpful and relates to this PR, hence cross posting it here.

nirs:
> I does not work. I deployed a deployment using pvc restored from snapshot, and tried to protect is. This is expected to fail since the dr policy is not enabled for flattening. https://github.com/nirs/ocm-ramen-samples/tree/flatten/workloads/flatten/regional-dr
> 
> I get empty messages for all conditions instead of the expected error message.

The test was wrong, we used the manifests form 0.10.0. Testing again with the manifests from this branch and a new version of the image:
https://github.com/RamenDR/ramen/compare/c6135d3dcbf198f302c1bc1f98ebd6a4bb2e8e2e..d65c6eb9ed478c3affd8b18adb061094c078798b

nirs:
Works now with fixed ramen change:
https://github.com/RamenDR/ramen/pull/1639/commits/c496fe945f5d33a42753622d2f4f6bfa5de3c10b

```
% kubectl get vr -A --context dr1 -o yaml              
apiVersion: v1
items:
- apiVersion: replication.storage.openshift.io/v1alpha1
  kind: VolumeReplication
  metadata:
    creationTimestamp: "2024-11-05T13:42:41Z"
    finalizers:
    - replication.storage.openshift.io
    generation: 1
    labels:
      ramendr.openshift.io/owner-name: flatten-drpc
      ramendr.openshift.io/owner-namespace-name: ramen-ops
    name: restored-pvc
    namespace: flatten
    resourceVersion: "40696"
    uid: 9bf65b35-b3da-42dc-a9c0-fbb6a3927b1b
  spec:
    autoResync: false
    dataSource:
      apiGroup: ""
      kind: PersistentVolumeClaim
      name: restored-pvc
    replicationHandle: ""
    replicationState: primary
    volumeReplicationClass: vrc-sample
  status:
    conditions:
    - lastTransitionTime: "2024-11-05T13:42:41Z"
      message: failed to promote volume
      observedGeneration: 1
      reason: FailedToPromote
      status: "False"
      type: Completed
    - lastTransitionTime: "2024-11-05T13:42:41Z"
      message: failed to enable volume replication
      observedGeneration: 1
      reason: Error
      status: "True"
      type: Degraded
    - lastTransitionTime: "2024-11-05T13:42:41Z"
      message: volume is not resyncing
      observedGeneration: 1
      reason: NotResyncing
      status: "False"
      type: Resyncing
    - lastTransitionTime: "2024-11-05T13:42:41Z"
      message: 'failed to meet prerequisite: rpc error: code = FailedPrecondition
        desc = system is not in a state required for the operation''s execution: failed
        to enable mirroring on image "replicapool/csi-vol-f4737b6e-eeff-4137-8248-301cf37a3368":
        parent image "replicapool/csi-snap-e7c91292-a272-4278-9ee9-6be7a4c8bfe0" is
        not enabled for mirroring'
      observedGeneration: 1
      reason: PrerequisiteNotMet
      status: "False"
      type: Validated
    message: 'system is not in a state required for the operation''s execution: failed
      to enable mirroring on image "replicapool/csi-vol-f4737b6e-eeff-4137-8248-301cf37a3368":
      parent image "replicapool/csi-snap-e7c91292-a272-4278-9ee9-6be7a4c8bfe0" is
      not enabled for mirroring'
    observedGeneration: 1
    state: Unknown
kind: List
metadata:
  resourceVersion: ""
```

The duplicate .Status.message can be eliminating, but maybe someone is depending on this?

yati1998:
> The duplicate .Status.message can be eliminating, but maybe someone is depending on this?

Yes we can't eliminate that.

yati1998:
since @nirs has also tested the change is statisfied with the current modifications, I would request a final review on this.
I have addressed all the comments above .

cc @nixpanic @nirs @ShyamsundarR @Madhu-1

api/replication.storage/v1alpha1/volumereplication_types.go
Madhu-1: @yati1998 this is still not addressed yet? Is it going to be a followup, If yes please open an issue for tracking.

api/replication.storage/v1alpha1/volumereplication_types.go
```diff
@@ -36,6 +36,21 @@ const (
...
+	MessageValidated          = "is validated and met all prerequisites"
+	MessageFailedPromoted     = "failed to promote "
+	MessageFailedDemoted      = "failed to demote "
+	MessageFailedPreCondition = "failed to meet prerequisite"
+	MessageDemoted            = "is demoted to secondary"
+	MessageDegraded           = "is degraded"
+	MessageResyncTriggered    = "is resyncing changes from primary to secondary"
+	MessageResyncFailed       = "failed to resync"
+)
+
```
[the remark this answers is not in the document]
Madhu-1:
```go
type Source string

const (
	Volume Source = "volume"
	VolumeGroup Source = volume group"
)

```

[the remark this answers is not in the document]
yati1998: thanks, will update the PR in sometime

internal/controller/replication.storage/status.go
```diff
@@ -140,20 +157,23 @@ func setDemotedCondition(conditions *[]metav1.Condition, observedGeneration int6
 }
 
 // sets conditions when volume demotion was failed.
-func setFailedDemotionCondition(conditions *[]metav1.Condition, observedGeneration int64) {
+func setFailedDemotionCondition(conditions *[]metav1.Condition, observedGeneration int64, message, messageFromDriver, dataSource string) {
```
nirs, line 160: For consistency, dataSource should be before message and messageFromDriver. It is easier to work with code that have consistent structure. If a group of function have common arguments, they should be in the same order.

To make the other arguments more clear, maybe:
- completedMessage
- degradedDetailedMessage?

With this is is very clear which messages belong to which condition and it is clear that completedMessage is not the same content as degradedDetails. One is message and the other is something longer and more detailed.

Same for the other functions getting datasource, message, messageFromDriver.

internal/controller/replication.storage/volumereplication_controller.go
```diff
@@ -371,7 +371,7 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	default:
 		replicationErr = fmt.Errorf("unsupported volume state")
 		logger.Error(replicationErr, "given volume state is not supported", "ReplicationState", instance.Spec.ReplicationState)
-		setFailureCondition(instance)
+		setFailureCondition(instance, "unsupported volume state", replicationErr.Error(), instance.Spec.DataSource.Kind)
```
Madhu-1, line 374: use `replicationErr.Error() here?

Madhu-1 (approved):

internal/controller/replication.storage/volumereplication_controller.go
yati1998: yes, we dont have any other error here to be passed.

yati1998:
@nirs @nixpanic @ShyamsundarR can you please give your final review here?

nirs (approved):
Looks awesome, thanks!

PR #704, opened 2024-11-12
Fix key in connection pool

Madhu-1:
In the connection pool we store the connection with unique identifier, the pod name and the namespace is used as the key to the connection  in the pool and GetLeaderByDriver get the leader based on the driver name and uses the identity of the leader which is pod name as the key to get the connection.

### DIFF
--- a/internal/controller/csiaddons/csiaddonsnode_controller.go
+++ b/internal/controller/csiaddons/csiaddonsnode_controller.go
@@ -93,8 +93,19 @@ func (r *CSIAddonsNodeReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	nodeID := csiAddonsNode.Spec.Driver.NodeID
 	driverName := csiAddonsNode.Spec.Driver.Name
 
-	key := csiAddonsNode.Namespace + "/" + util.NormalizeLeaseName(csiAddonsNode.Name)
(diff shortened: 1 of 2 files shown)


internal/controller/csiaddons/csiaddonsnode_controller.go
```diff
@@ -93,9 +93,20 @@ func (r *CSIAddonsNodeReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
-	key := csiAddonsNode.Namespace + "/" + util.NormalizeLeaseName(csiAddonsNode.Name)
 	logger = logger.WithValues("NodeID", nodeID, "DriverName", driverName)
 
+	podName, endPoint, err := r.resolveEndpoint(ctx, csiAddonsNode.Spec.Driver.EndPoint)
+	if err != nil {
+		logger.Error(err, "Failed to resolve endpoint")
+		return ctrl.Result{}, fmt.Errorf("failed to resolve endpoint %q: %w", csiAddonsNode.Spec.Driver.EndPoint, err)
+	}
+
+	// namespace + "/" + leases identity(pod name) is the key for the connection.
```
nixpanic, line 104: `leases identity`, maybe you mean _leader identity_, or just use _pod name_ only.

nixpanic (approved):

yati1998 (approved):

PR #718, opened 2024-11-27
Fix key and ownerRef bug

Madhu-1:
csi-lib-utils/leaderelection:sanitizeName() is used to sanitize the leader identity used for the leases csiaddonsnode need to store the key with same format so that it can be used to get the connection.

The pod owned by deployment might move to new nodes and this might create the stale CSIAddonsNode object on the old node if we set Deployment as the ownerRef (this could leave stale connection in connection map and a CR).
so, we need to set the pod as the owner for the deployment as we dont store any details that are required later on for any other operations like Fencing etc.

### DIFF
--- a/internal/controller/csiaddons/csiaddonsnode_controller.go
+++ b/internal/controller/csiaddons/csiaddonsnode_controller.go
@@ -105,7 +105,9 @@ func (r *CSIAddonsNodeReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 
 	// namespace + "/" + leader identity(pod name) is the key for the connection.
 	// this key is used by GetLeaderByDriver to get the connection
-	key := csiAddonsNode.Namespace + "/" + podName
+	// util.NormalizeLeaseName() is used to sanitize the leader identity used for the leases
+	// csiaddonsnode need to store the key with same format so that it can be used to get the connection.
+	key := csiAddonsNode.Namespace + "/" + util.NormalizeLeaseName(podName)
 
 	logger = logger.WithValues("EndPoint", endPoint)
 
--- a/sidecar/internal/csiaddonsnode/csiaddonsnode.go
+++ b/sidecar/internal/csiaddonsnode/csiaddonsnode.go
@@ -30,6 +30,7 @@ import (
 
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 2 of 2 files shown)


internal/controller/csiaddons/csiaddonsnode_controller.go
```diff
@@ -105,7 +105,9 @@ func (r *CSIAddonsNodeReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 
 	// namespace + "/" + leader identity(pod name) is the key for the connection.
 	// this key is used by GetLeaderByDriver to get the connection
-	key := csiAddonsNode.Namespace + "/" + podName
+	// csi-lib-utils/leaderelection:sanitizeName() is used to sanitize the leader identity used for the leases
```
nixpanic, line 108: `csi-lib-utils/leaderelection:sanitizeName()` is never called, that makes the comment very confusing. Maybe mention `util.NormalizeLeaseName()` instead?

Madhu-1: its called internall when we create the leader election object using csi-lib-utils packages, i updated to mention to use util.NormalizeLeaseName(). PTAL

sidecar/internal/csiaddonsnode/csiaddonsnode.go
```diff
@@ -198,14 +204,28 @@ func (mgr *Manager) getCSIAddonsNode() (*csiaddonsv1alpha1.CSIAddonsNode, error)
 		if len(rs.OwnerReferences) == 0 {
 			return nil, fmt.Errorf("%w: replicaset has no owner", errInvalidConfig)
 		}
-		ownerReferences = append(ownerReferences, rs.OwnerReferences[0])
+		ownerKindForCSIAddonsName = rs.OwnerReferences[0].Kind
+		ownerNameForCSIAddonsName = rs.OwnerReferences[0].Name
+
+		// The pod might move to new nodes and this might create the stale CSIAddonsNode object.
+		// So, we need to set the pod as the owner for the deployment as we dont store any details
```
nixpanic, line 211: I don't understand this comment. `set the pod as the owner for the deployment`... Which deployment? The Deployment creates a ReplicaSet which creates the Pods :thinking:

Looking at the code, you mean `set the pod as the owner for the CSIAddonsNode` instead.

Madhu-1: Yes thats correct, updated the comment to refect the same where we need to set pod as the owner to avoid garbage connections and CR's

nixpanic (approved):

Madhu-1:
@mergifyio rebase

Madhu-1:
@mergifyio rebase

Issue #643, opened 2024-08-16
Consider using NetworkPolicies to allow/deny access to the CSI-driver sidecar


Madhu-1:
the kube-proxy container has a warning for insecure access https://github.com/kubernetes-sigs/kubebuilder/pull/3899, we need to adopt as per https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/discontinue_usage_of_kube_rbac_proxy.md

nixpanic:
Fome rge [summary](https://github.com/kubernetes-sigs/kubebuilder/blob/master/designs/discontinue_usage_of_kube_rbac_proxy.md#summary):

> Existing users are encouraged to switch to images hosted by the project on [quay.io](https://quay.io/repository/brancz/kube-rbac-proxy?tab=tags&tag=latest) OR to adapt their projects to utilize [Network Policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/), following the updated scaffold guidelines.

NetworkPolicies would require additional configuration on the deployment by users. They need to open-up the port of the CSI-Addons sidecar that is running as part of their CSI-driver. This requires a bit more consideration than replacing the container-image repository.

PR #721, opened 2024-12-04
Discontinue kube-rbac-proxy and disable metrics service

black-dragon74:
This patch drops support for kube-rbac-proxy
and uses controller manager's
WithAuthenticationAndAuthorization.

Closes: https://github.com/csi-addons/kubernetes-csi-addons/issues/643

For more details please refer [here](https://book.kubebuilder.io/reference/metrics.html)

In addition to migration, some of the resources have been re-categorized for better placement and the metrics service is now disabled by default as it did not contain any significant information specific to csi-addons. It can be re-enabled by following the steps outlined in https://github.com/csi-addons/kubernetes-csi-addons/pull/721#issuecomment-2548141096


black-dragon74:
## Testing

### 1. Access `/metrics` endpoint as controller-manager's SA
```sh
❯ oc auth can-i get /metrics --as=system:serviceaccount:csi-addons-system:csi-addons-controller-manager
yes
```

### 2. Get the metrics using valid SA token
```sh
❯ oc exec -it po/metrics-consumer -- sh
$ curl -k -H "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" https://csi-addons-controller-manager-metrics-service.csi-addons-system.svc.cluster.local:8443/metrics
# HELP certwatcher_read_certificate_errors_total Total number of certificate read errors
# TYPE certwatcher_read_certificate_errors_total counter
certwatcher_read_certificate_errors_total 0
# HELP certwatcher_read_certificate_total Total number of certificate reads
...
...
...
stripped....
```

### 3. Get the metrics using invalid SA token

```sh
$ curl -k -H "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/INVALID)" https://csi-addons-controller-manager-metrics-service.csi-addons-system.svc.cluster.local:8443/metrics
Unauthorized
```

### 4. Check POD's containers (no kube-rbac-proxy-container)

```sh
❯ oc get po/csi-addons-controller-manager-7886c6d5c9-f2vwh -o jsonpath='{.spec.containers[*].name}'
manager
```

Regards

cmd/manager/main.go
```diff
@@ -128,8 +135,27 @@ func main() {
...
+		tlsOpts = append(tlsOpts, disableHTTP2)
+	}
+
+	metricsServerOptions := metricsserver.Options{
+		BindAddress:   metricsAddr,
+		SecureServing: secureMetrics,
+		TLSOpts:       tlsOpts,
+	}
+
+	if secureMetrics {
```
Madhu-1, line 153: why someone will use insecure metrics, cant we always use secureMetircs?

cmd/manager/main.go
```diff
@@ -128,8 +135,27 @@ func main() {
 		os.Exit(1)
 	}
 
+	disableHTTP2 := func(config *tls.Config) {
```
Madhu-1, line 138: Is this one for test clusters or means for production clusters also?

black-dragon74: Production clusters as well. As of now, `curl` misses support for HTTP2 on some distros. We can enable it by default once we are sure that all the clients support HTTP2.

Kubebuilder's default scaffolding also follows the same procedure as of now.

cmd/manager/main.go
black-dragon74: It is enabled by default. The reason why it is controlled by a flag of its own is, in prod env it is recommended to setup proper certs and keys before opting in for secure metrics.

If someone does not want to do that and just wants to check something, they may disable it.

cmd/manager/main.go
nixpanic, line 138: please leave a comment in the code why it is disabled by default

cmd/manager/main.go
nixpanic: can you document the steps to provide certificates, so that production deployments may replace automatic generated ones?

nixpanic (changes requested):
looks good overall, just a few nits regarding comments and documentation

Madhu-1 (commented):
on other thought, Do we need the metrics? if not lets disable it?

black-dragon74:
> on other thought, Do we need the metrics? if not lets disable it?

It is enabled in all of the other projects by default. Still, we could disable it if not needed. Thoughts @nixpanic ?

nixpanic:
> > However, the metrics should be useful though. Not sure what the metrics contain, and if there is anything related to the CSI-Addons procedures.
> 
> It includes [these](https://book.kubebuilder.io/reference/metrics-reference) metrics and also the metrics for all the CRs that are reconciled by the controllers.

In that case, disable it by default. These are not metrics about CSI-Addons procedures, it is too incomplete for that. Once we add metrics about the different CSI-Addons procedures, it can be enabled by default.

cmd/manager/main.go
black-dragon74: Done

cmd/manager/main.go
black-dragon74: Done

black-dragon74:
The metrics service is now disabled by default (by setting its bind address to 0). In order to enable it modify `config/manager/kustomization.yaml` and uncomment the lines 13-18:
```yaml
#- metrics_service.yaml

#patches:
#- path: manager_metrics_patch.yaml
#  target:
#    kind: Deployment
```

Ensure the necessary RBACs are present by uncommenting the lines 27-30 in `config/rbac/kustomization.yaml`
```yaml
  # The following RBAC configurations are used to protect
  # the metrics endpoint with authn/authz. These configurations
  # ensure that only authorized users and service accounts
  # can access the metrics endpoint.
  # - metrics_auth_role.yaml
  # - metrics_auth_role_binding.yaml
  # - metrics_reader_role.yaml
  # - metrics_reader_role_binding.yaml
```

cc @nixpanic @Madhu-1

cmd/manager/main.go
```diff
@@ -138,7 +174,8 @@ func main() {
 		RenewDeadline:          &leaderElectionRenewDeadline,
 		RetryPeriod:            &leaderElectionRetryPeriod,
 		WebhookServer: webhook.NewServer(webhook.Options{
-			Port: 9443,
+			Port:    9443,
```
Madhu-1, line 177: we might not need webhook server as well, lets check and address this in followup as well.

Madhu-1 (approved):

PR #722, opened 2024-12-04
csiaddons: allow deleting of the CR

Madhu-1:
The pod is set as the owner of the csiaddons node object, and when the owner pod is deleted the csiaddons need to be deleted as well, When the delete Reconciles happens we cannot get the pod deletails to resolve the endpoint and we dont
need that as well when we are deleting the CR.


Nikhil-Ladha (approved):
lgtm

Rakshith-R (approved):

iPraveenParihar (approved):

PR #782, opened 2025-03-18
api: add volumeGroupAttributes field to VGRContent API

iPraveenParihar:
This commit introduces `volumeGroupAttributes` field to
`VolumeGroupReplicationContent` API.

The field is intended to store key-value pairs derived
from the VolumeGroupContext returned by CSI drivers in the
CreateVolumeGroupResponse.

### DIFF
--- a/docs/volumegroupreplicationclass.md
+++ b/docs/volumegroupreplicationclass.md
@@ -19,6 +19,7 @@ metadata:
 spec:
   provisioner: example.provisioner.io
   parameters:
+    clusterID: my-cluster
     replication.storage.openshift.io/group-replication-secret-name: secret-name
     replication.storage.openshift.io/group-replication-secret-namespace: secret-namespace
 ```
--- a/docs/volumegroupreplicationcontent.md
+++ b/docs/volumegroupreplicationcontent.md
@@ -2,6 +2,9 @@
 
 VolumeGroupReplicationContent is a cluster scoped resource that contains volume grouping related information.
 
+`volumeGroupAttributes` contains key-value pairs from the VolumeGroupContext field available in the
+CreateVolumeGroupResponse returned by the CSI driver.
+
 `volumeGroupReplicationRef` contains object reference of the volumeGroupReplication resource that created this resource.
 
 `volumeGroupReplicationHandle` (optional) is an existing (but new) group replication ID.
@@ -19,6 +22,8 @@ kind: VolumeGroupReplicationContent
 metadata:
   name: volumegroupreplicationcontent-sample
 spec:
+  volumeGroupAttributes:
+    clusterID: my-cluster
   volumeGroupReplicationRef:
     kind: VolumeGroupReplication
     name: volumegroupreplication-sample


nixpanic (changes requested):
What kind of attributes are these, and where do they come from? Would be good to see something about them in the documentation and/or protocol specification.

iPraveenParihar:
> What kind of attributes are these, and where do they come from? Would be good to see something about them in the documentation and/or protocol specification.

_fyi: https://github.com/ceph/ceph-csi/issues/5220_

nixpanic:
> _fyi: [ceph/ceph-csi#5220](https://github.com/ceph/ceph-csi/issues/5220)_

Thanks for the reference. That is just one implementation of a csi-driver with csi-addons support. Please included a description/design for the need of attributes, and what attributes are somewhere in this csi-driver-independent project.

iPraveenParihar:
> > _fyi: [ceph/ceph-csi#5220](https://github.com/ceph/ceph-csi/issues/5220)_
> 
> Thanks for the reference. That is just one implementation of a csi-driver with csi-addons support. Please included a description/design for the need of attributes, and what attributes are somewhere in this csi-driver-independent project.

Instead of specifying what parameters to add or adding all, I'd like to do for volume‑groups what CSI already does for individual volumes, when we create a PV, the external provisioner takes the key‑value pairs returned by the CSI driver in `CreateVolumeResponse.Volume.VolumeContext` and stuffs them into `pv.spec.csi.volumeAttributes`. 

Can we do the same thing for a `VolumeGroup`? IOW, have the CSI drivers return its own attributes in `CreateVolumeGroupResponse.VolumeGroup.VolumeGroupContext`, and then let the VolumeGroupReplicationContent controller copy those entries into `VolumeGroupReplicationContent.spec.volumeGroupAttributes`? 

WDYT? @Madhu-1 @Nikhil-Ladha

Madhu-1:
> > > _fyi: [ceph/ceph-csi#5220](https://github.com/ceph/ceph-csi/issues/5220)_
> > 
> > 
> > Thanks for the reference. That is just one implementation of a csi-driver with csi-addons support. Please included a description/design for the need of attributes, and what attributes are somewhere in this csi-driver-independent project.
> 
> Instead of specifying what parameters to add or adding all, I'd like to do for volume‑groups what CSI already does for individual volumes, when we create a PV, the external provisioner takes the key‑value pairs returned by the CSI driver in `CreateVolumeResponse.Volume.VolumeContext` and stuffs them into `pv.spec.csi.volumeAttributes`.
> 
> Can we do the same thing for a `VolumeGroup`? IOW, have the CSI drivers return its own attributes in `CreateVolumeGroupResponse.VolumeGroup.VolumeGroupContext`, and then let the VolumeGroupReplicationContent controller copy those entries into `VolumeGroupReplicationContent.spec.volumeGroupAttributes`?
> 
> WDYT? @Madhu-1 @Nikhil-Ladha

Yes thats what we need to do, its the context parameters returned by csi driver during create operation

nixpanic:
> Instead of specifying what parameters to add or adding all, I'd like to do for volume‑groups what CSI already does for individual volumes, when we create a PV, the external provisioner takes the key‑value pairs returned by the CSI driver in `CreateVolumeResponse.Volume.VolumeContext` and stuffs them into `pv.spec.csi.volumeAttributes`.

Sure. that is a reasonable explanation. As `CreateVolumeGroupResponse.VolumeGroup.VolumeGroupContext` already exists, it might even have been an oversight and this could be marked as a bug.

Nikhil-Ladha:
That is a better idea.
But, in that case we would still have to depend on VolumeGroupReplicationClass in the omap generator to fetch the secrets, as we won't be storing them in VGRContent. The secrets won't be returned in those attributes, right?
We have `controllerExpandSecretRef` in PV that contains the secrets, so are we planning to have something like that in VGRContent as well?

Madhu-1:
> But, in that case we would still have to depend on VolumeGroupReplicationClass in the omap generator to fetch the secrets, as we won't be storing them in VGRContent. The secrets won't be returned in those attributes, right?
> We have `spec.csi.controllerExpandSecretRef` in PV that contains the secrets, so are we planning to have something like that in VGRContent as well?

@Nikhil-Ladha IMO we dont need to introduce a new fields for it. we could store it as annotation in the VGR not at the VGRC. we pass the parameters from the VGRC always for all the Group RPC calls.  we can revisit the secrets later

Nikhil-Ladha:
> > But, in that case we would still have to depend on VolumeGroupReplicationClass in the omap generator to fetch the secrets, as we won't be storing them in VGRContent. The secrets won't be returned in those attributes, right?
> > We have `spec.csi.controllerExpandSecretRef` in PV that contains the secrets, so are we planning to have something like that in VGRContent as well?
> 
> @Nikhil-Ladha IMO we dont need to introduce a new fields for it. we could store it as annotation in the VGR not at the VGRC. we pass the parameters from the VGRC always for all the Group RPC calls. we can revisit the secrets later

But, isn't that was the initial purpose of the issue?
That is to avoid dependency on VGRClass for omap generation?

RPC calls are fine, we filter the secret params and pass it to the RPC.
But, even after the proposed change we would continue to use VGRClass in omap generator. So, I am not sure how are we planning to address that?

Madhu-1:
> RPC calls are fine, we filter the secret params and pass it to the RPC.
> But, even after the proposed change we would continue to use VGRClass in omap generator. So, I am not sure how are we planning to address that?

Yes we need to address it in different PR's. as i mentioned earlier we can use annotation for it.

Nikhil-Ladha:
> > RPC calls are fine, we filter the secret params and pass it to the RPC.
> > But, even after the proposed change we would continue to use VGRClass in omap generator. So, I am not sure how are we planning to address that?
> 
> Yes we need to address it in different PR's. as i mentioned earlier we can use annotation for it.

Ack.
Yeah, in case of annotation it would be VGRContent then instead of VGR, as that is referred for omap generation.

iPraveenParihar:
> Sure. that is a reasonable explanation. As CreateVolumeGroupResponse.VolumeGroup.VolumeGroupContext already exists, it might even have been an oversight and this could be marked as a bug.

we should be good with changes in this PR?

Nikhil-Ladha:
> i would store the secrets on the VGR not on VGRC (like we store the provisioner secret on the PVC) secrets are optional and not a mandatory fields .

In that case, we should update the issue description to reflect the same for awareness.

iPraveenParihar:
@Madhu-1 @Nikhil-Ladha PTAL
If there is nothing more to be done here, we can merge it?

Nikhil-Ladha (approved):
lgtm

nixpanic:
> Yes thats what we need to do, its the context parameters returned by csi driver during create operation

This is still something that needs to be set in the sidecar/controller, right? It would be good to include that in this PR as well. Otherwise the API suggests it is available, but drivers returning a VolumeGroupContext won't find it in the VGRContent.

iPraveenParihar:
@Nikhil-Ladha, Currently we are not passing the `secretName` & `secretNamespace` for CreateVolumeGroupRequest. 
So, the CSI drivers can't add these details to the `CreateVolumeGroupResponse.Volume.VolumeContext`. 

One way am thinking is to add secret details as annotation in VolumeGroupReplicationContent CR
```
replication.storage.openshift.io/group-replication-secret-name: <secret-name>
replication.storage.openshift.io/group-replication-secret-namespace: <secret-namespace>
```

this should be okay?

Nikhil-Ladha:
> @Nikhil-Ladha, Currently we are not passing the `secretName` & `secretNamespace` for CreateVolumeGroupRequest. So, the CSI drivers can't add these details to the `CreateVolumeGroupResponse.Volume.VolumeContext`.
> 
> One way am thinking is to add secret details as annotation in VolumeGroupReplicationContent CR
> 
> ```
> replication.storage.openshift.io/group-replication-secret-name: <secret-name>
> replication.storage.openshift.io/group-replication-secret-namespace: <secret-namespace>
> ```
> 
> this should be okay?

We are passing these details to the CreateVolumeGroup RPC call, ref: https://github.com/csi-addons/kubernetes-csi-addons/blob/main/internal/controller/replication.storage/volumegroupreplicationcontent_controller.go#L184

internal/controller/replication.storage/volumegroupreplication_controller.go
```diff
@@ -673,9 +679,7 @@ func (r *VolumeGroupReplicationReconciler) createOrUpdateVolumeGroupReplicationC
 			return errors.New("dependent VGRContent resource is not yet created, waiting for it to be created")
 		}
 		if vgrContentObj.CreationTimestamp.IsZero() {
-			vgrContentObj.Annotations = map[string]string{
-				volumeGroupReplicationRef: vgrRef,
-			}
+			metav1.SetMetaDataAnnotation(&vgrContentObj.ObjectMeta, volumeGroupReplicationRef, vgrRef)
```
Nikhil-Ladha, line 682: Is there any new lint changes that require this change?

nixpanic: it is not done consistently either, below is
```go
vgrContentObj.Annotations[volumeGroupReplicationRef] = vgrRef
```

iPraveenParihar: > Is there any new lint changes that require this change?

No, this is a just a helper function available in k8s API. Takes care of handling initialising Annotations object if nil.
```go
// SetMetaDataAnnotation sets the annotation and value
func SetMetaDataAnnotation(obj *ObjectMeta, ann string, value string) {
	if obj.Annotations == nil {
		obj.Annotations = make(map[string]string)
	}
	obj.Annotations[ann] = value
}
```

Nikhil-Ladha: If that's the only reason then I don't think we need this helper function at all over here.
At all the places where we are calling this function is already defining the struct just before, or at the same time. By, using this function there is an extra overhead of function call, and condition check which we can avoid completely.

iPraveenParihar: IMO, without the helper function, we either need to repeat the nil-check at places or assume it's already handled, which increases the risk of inconsistency and bugs. The overhead of using the helper is extremely minimal compared to the safety and clarity it brings.

Nikhil-Ladha: That's true when we are adding annotation at the fly, in this case we are adding annotation when we are initializing and not later on.

api/replication.storage/v1alpha1/volumegroupreplicationcontent_types.go
```diff
@@ -57,6 +57,10 @@ type VolumeGroupReplicationContentSpec struct {
 	// +kubebuilder:validation:Required
 	VolumeGroupReplicationClassName string `json:"volumeGroupReplicationClassName"`
 
+	// volumeGroupAttributes holds the contextual information of the volume group.
+	// +kubebuilder:validation:Optional
+	VolumeGroupAttributes map[string]string `json:"volumeGroupAttributes,omitempty"`
```
Madhu-1, line 62: Please add a rule that once its created cannot be updated (immutable)

internal/controller/replication.storage/volumegroupreplication_controller.go
```diff
@@ -137,6 +140,9 @@ func (r *VolumeGroupReplicationReconciler) Reconcile(ctx context.Context, req ct
 			Name: vgrContentName,
 		},
 	}
+	metav1.SetMetaDataAnnotation(&vgrContentObj.ObjectMeta, prefixedGroupReplicationSecretNameKey, secretName)
+	metav1.SetMetaDataAnnotation(&vgrContentObj.ObjectMeta, prefixedGroupReplicationSecretNamespaceKey, secretNamespace)
```
Madhu-1, line 144: Secrets are optional set it  in annotation only if its set.

internal/controller/replication.storage/volumegroupreplication_controller.go
Madhu-1: @Nikhil-Ladha lets use the helper provided by the kubernetes also it will help us to preserve the annotation if users/admin might have added

Madhu-1 (changes requested):

docs/volumegroupreplicationclass.md
```diff
@@ -19,6 +19,7 @@ metadata:
 spec:
   provisioner: example.provisioner.io
   parameters:
+    clusterID: my-cluster
```
Nikhil-Ladha, line 22: Well, if we are adding this then we should add the `pool` param as well.

PR #793, opened 2025-04-14
fix: avoid panic due to incorrect logger key-value pairs

iPraveenParihar:
```
This commit fixes PANIC caused by an odd number of arguments passed
as key-value pairs to the structured logger.
This ensures that all key-value pairs are correctly formatted to
prevent runtime issues during logging.
```

### DIFF
--- a/internal/controller/csiaddons/reclaimspacejob_controller.go
+++ b/internal/controller/csiaddons/reclaimspacejob_controller.go
@@ -394,7 +394,7 @@ func (r *ReclaimSpaceJobReconciler) controllerReclaimSpace(
 	target *targetDetails) (bool, *int64, error) {
 	controllerClient, err := r.getLeadingRSClient(ctx, r.Client, target.driverName)
 	if err != nil {
-		logger.Info("Controller Client not found: %v", err)
+		logger.Error(err, "Failed to get controller client", "driverName", target.driverName)
 		return false, nil, err
 	}
 


black-dragon74 (approved):

internal/controller/csiaddons/reclaimspacejob_controller.go
```diff
@@ -394,7 +394,7 @@ func (r *ReclaimSpaceJobReconciler) controllerReclaimSpace(
 	target *targetDetails) (bool, *int64, error) {
 	controllerClient, err := r.getLeadingRSClient(ctx, r.Client, target.driverName)
 	if err != nil {
-		logger.Info("Controller Client not found: %v", err)
+		logger.Error(err, "Failed to get controller client", "driverName", target.driverName)
```
Nikhil-Ladha, line 397: Let's log the actual error as well.

Madhu-1: `logger.Error(err,` is already logging the error, anything is missing?

Madhu-1 (approved):

internal/controller/csiaddons/reclaimspacejob_controller.go
Nikhil-Ladha: Ah, sorry.
Misread it as `ctx`.

Issue #117, opened 2022-02-11
idle GRPC connections in controller


Madhu-1:
As we already know currently, a csiaddons node object is created, we create the connections and keep it until the addons node object is deleted. there could be advantages/disadvantages of this one. As csiaddons is meant to be a generic component and it will be used by multiple csi drivers. Just for an example of 10 nodes cluster and 2 csidrivers are using the csiaddons. We have 20 or 2 connections  (for both provisioner and node plugin sidecars are deployed) opened and kept in the in-memory, thinking about the scale what about the 100 nodes clusters or even more csidrivers in a cluster?

## Advantages
* Reuse of connection for faster communication
* (anything else?)

## Disadvantages
* More number of idle connections in in-memory (if there are no addons operations)
* More resources utilization in the controller pod
* More network calls as connection keep-alive need to be checked in intervals to make sure the connection is not broken

I would like to hear thoughts from everyone on this one. cc @nixpanic @humblec @Rakshith-R  @pkalever

Rakshith-R:
> As we already know currently, a csiaddons node object is created, we create the connections and keep it until the addons node object is deleted. there could be advantages/disadvantages of this one. As csiaddons is meant to be a generic component and it will be used by multiple csi drivers. Just for an example of 10 nodes cluster and 2 csidrivers are using the csiaddons. We have 20 or 2 connections (for both provisioner and node plugin sidecars are deployed) opened and kept in the in-memory, thinking about the scale what about the 100 nodes clusters or even more csidrivers in a cluster?

Consider x number of csiaddons node object of the same csidriver, 
Without connections being in memory. 

#### For each operation: 
- The specific controller would have to list all the csiaddons node object 
- For each csiaddonsnode object: 
  - Open a grpc connection with the pod
  - Fetch capability, 
  - if it does not match, close it. pick the next one
  - if it does match, use it.
  
I think the above is enough reason to keep single pool of open connections which can be shared by all operations.

Madhu-1:
[...]

PR #794, opened 2025-04-16
connection: Disconnect idle gRPC connections on inactivity

black-dragon74:
This patch adds support to close the lingering gRPC connections
that have not been used for a certain amount of time, default to 5 mins.

An interceptor is used to track the last access time in a reliable
manner.

If a expired connection is requested, it will first be re-created and
then returned to the caller.

Since gRPC connections are not simple disconnects but a complete
tear down, re-creating the connections is necessary.

Closes: #117

### DIFF
--- a/internal/connection/connection.go
+++ b/internal/connection/connection.go
@@ -19,6 +19,8 @@ package connection
 import (
 	"context"
 	"crypto/tls"
+	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/csi-addons/kubernetes-csi-addons/internal/kubernetes/token"
@@ -29,9 +31,16 @@ import (
 	"google.golang.org/grpc/credentials/insecure"
 )
 
+const (
+	// The duration after which a gRPC connection is closed due to inactivity
+	idleTimeout = time.Minute * 5
+)
+
 // Connection struct consists of to NodeID, DriverName, Capabilities for the controller
 // to pick sidecar connection and grpc Client to connect to the sidecar.
 type Connection struct {
+	sync.Mutex
+
 	Client       *grpc.ClientConn
 	Capabilities []*identity.Capability
 	Namespace    string
@@ -39,13 +48,50 @@ type Connection struct {
 	NodeID       string
 	DriverName   string
 	Timeout      time.Duration
+
+	// Holds the internal state of the connection
+	connected  bool
+	enableAuth bool
+	endpoint   string
+	podName    string
+
+	// Used to cancel any existing timers in case of re-connects
+	cancelIdle context.CancelFunc
+	// Holds the last access time of the gRPC Connection
+	// It is tracked and updated by the `accessTimeInterceptor`
+	lastAccessTime atomic.Int64
 }
 
-// NewConnection establishes connection with sidecar, fetches capability and returns Connection object
-// filled with required information.
-func NewConnection(ctx context.Context, endpoint, nodeID, driverName, namespace, podName string, enableAuth bool) (*Connection, error) {
-	var opts []grpc.DialOption
-	if enableAuth {
+// accessTimeInterceptor is an unary interceptor which updates the lastAccessTime
+// on `Connection` struct to `time.Now()` on each RPC call
+func accessTimeInterceptor(conn *Connection) grpc.UnaryClientInterceptor {
+	return func(ctx context.Context, method string, req, reply any, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+		conn.lastAccessTime.Store(time.Now().UnixNano())
+
+		return invoker(ctx, method, req, reply, cc, opts...)
+	}
+}
+
+// Connect creates a new grpc.ClientConn object and sets it as the
+// client property on Connection struct. If a connection is already
+// connected, it is reused as is. It also spawns a go routine to tear
+// down the connection if it has been idling for a specified threshold.
+//
+// In cases where a new connection is created from the scratch Connect
+// also calls fetchCapabilities on the connection object.
+func (c *Connection) Connect() error {
+	c.Lock()
+	defer c.Unlock()
+
+	// Return early if already connected, the connection will be reused
+	if c.connected {
+		return nil
+	}
+
+	opts := []grpc.DialOption{
+		grpc.WithUnaryInterceptor(accessTimeInterceptor(c)),
+	}
+	if c.enableAuth {
 		opts = append(opts, token.WithServiceAccountToken())
 		tlsConfig := &tls.Config{
 			// Certs are only used to initiate HTTPS connections; authorization is handled by SA tokens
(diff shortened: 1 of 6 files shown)


black-dragon74:
### Reserved for testing..

internal/connection/connection.go
```diff
@@ -29,23 +31,60 @@ import (
 	"google.golang.org/grpc/credentials/insecure"
 )
 
+const (
+	// The duration after which a gRPC connection is closed due to inactivity
+	IDLE_TIMEOUT = time.Minute * 5
```
nixpanic, line 36: I guess `defaultIdleTimeout` would more suitable, and maybe have it as an optional configuration option in the configmap?

internal/connection/connection.go
```diff
@@ -29,23 +31,60 @@ import (
...
+// on `Connection` struct to `time.Now()` on each RPC call
+func accessTimeInterceptor(conn *Connection) grpc.UnaryClientInterceptor {
+	return func(ctx context.Context, method string, req, reply any, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+		conn.lastAccessTime.Store(time.Now().UnixNano())
+
+		return invoker(ctx, method, req, reply, cc, opts...)
+	}
+}
+
+func (c *Connection) Connect(ctx context.Context) error {
```
nixpanic, line 76: Add a small comment for this function. Maybe mention when/what calls it?

nixpanic, line 76: is the ctx parameter really useful (is it used at all?), maybe it should be allocated within this `Connect()` function instead

internal/connection/connection_pool.go
```diff
@@ -82,34 +84,49 @@ func (cp *ConnectionPool) Delete(key string) {
...
 // getByDriverName returns map of connections filtered by driverName. This function
 // must be called with read lock held.
-func (cp *ConnectionPool) getByDriverName(driverName string) map[string]*Connection {
+func (cp *ConnectionPool) getByDriverName(driverName string) (map[string]*Connection, error) {
 	newPool := make(map[string]*Connection)
 	for k, v := range cp.pool {
 		if v.DriverName != driverName {
 			continue
 		}
+		if !v.connected {
```
nixpanic, line 93: This if-block is repeated quite a few times. It would be nicer to put it in an `Connection.EnsureConnected()` function, or something like that.

internal/connection/connection.go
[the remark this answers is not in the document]
black-dragon74: `defaultIdleTimeout` is used inside `fetchCapabilities`, setting this to a higher value would leave us hanging in cases where the network is unreliable. Also, we would ideally always want the gRPC connection timeout to be lesser than gRPC idle timeout.

I see the benefits of having it configurable via a configmap but that would take away the `connection` package's independence as it would require a `ClientSet` to fetch and read the configmap. What do you suggest?

internal/connection/connection_pool.go
black-dragon74: Done!

internal/connection/connection.go
black-dragon74: Done

black-dragon74: Done

internal/connection/connection.go
```diff
@@ -56,34 +95,106 @@ func NewConnection(ctx context.Context, endpoint, nodeID, driverName, namespace,
...
 	conn := &Connection{
-		Client:     cc,
 		Namespace:  namespace,
 		Name:       podName,
 		NodeID:     nodeID,
 		DriverName: driverName,
 		Timeout:    time.Minute,
+
+		endpoint:   endpoint,
+		nodeID:     nodeID,
```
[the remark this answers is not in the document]
black-dragon74: Overlook on my part, fixed :)

internal/connection/connection.go
```diff
@@ -29,23 +31,68 @@ import (
...
 	Name         string
 	NodeID       string
 	DriverName   string
 	Timeout      time.Duration
+
+	// Holds the internal state of the connection
+	connected  bool
+	enableAuth bool
+	endpoint   string
+	driverName string
```
nixpanic, line 56: `driverName` is a duplicate of `DriverName`

internal/connection/connection.go
```diff
@@ -29,23 +31,68 @@ import (
...
+// down the connection if it has been idling for a specified threshold.
+//
+// In cases where a new connection is created from the scratch Connect
+// also calls fetchCapabilities on the connection object.
+func (c *Connection) Connect() error {
+	c.Lock()
+	defer c.Unlock()
+
+	// Return early if already connected, the connection will be reused
+	if c.connected {
```
nixpanic, line 88: maybe update `conn.lastAccessTime.Store(time.Now().UnixNano())` here?

It might be cleaner to have that as a function like the below, so that storing a new timestamp is done only in one place.
```go
func (c *Connection) touch() {
    c.lastAccessTime.Store(time.Now().UnixNano())
}
```

internal/connection/connection.go
```diff
@@ -56,34 +103,110 @@ func NewConnection(ctx context.Context, endpoint, nodeID, driverName, namespace,
...
+// monitoring the idle timeout by calling cancelIdle()
 func (c *Connection) Close() error {
-	if c.Client == nil {
-		return nil
+	c.Lock()
+	defer c.Unlock()
+
+	if c.cancelIdle != nil {
+		// Cancel the context as well
+		c.cancelIdle()
```
nixpanic, line 197: set `c.cancelIdle = nil` after calling it?

internal/connection/connection.go
black-dragon74: Is that really necessary here? The interceptor (middleware) will update it on every request it processes? After returning early if already connected, ideally there will be subsequent requests, which in turn will update the last access time.

internal/connection/connection.go
```diff
@@ -56,34 +103,110 @@ func NewConnection(ctx context.Context, endpoint, nodeID, driverName, namespace,
...
 	conn := &Connection{
-		Client:     cc,
 		Namespace:  namespace,
 		Name:       podName,
 		NodeID:     nodeID,
 		DriverName: driverName,
 		Timeout:    time.Minute,
+
+		endpoint:   endpoint,
+		driverName: driverName,
```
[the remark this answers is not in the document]
black-dragon74: Done!

internal/connection/connection.go
black-dragon74: Done

internal/connection/connection.go
black-dragon74: The ticker doesn't send an event until the first tick which is in 30 seconds.

internal/connection/connection.go
```diff
@@ -29,23 +31,67 @@ import (
 	"google.golang.org/grpc/credentials/insecure"
 )
 
+const (
+	// The duration after which a gRPC connection is closed due to inactivity
+	IDLE_TIMEOUT = time.Minute * 5
```
nixpanic, line 36: Call this `IdleTimeout`, or, if this is internal to this package, `idleTimeout`. Capitalization like `IDLE_TIMEOUT` is really not common in Go.

nixpanic (commented):
Just one remaining nit, looks good otherwise.

nixpanic (approved):

internal/connection/connection.go
```diff
@@ -102,6 +224,16 @@ func (c *Connection) fetchCapabilities(ctx context.Context) error {
...
 
+// ensureConnected re-eastablishes a connection if it is in a disconnected
+// state. An error is returned if it fails to renew the connection.
+func (c *Connection) ensureConnected() error {
+	if !c.connected {
+		return c.Connect()
+	}
+
+	return nil
+}
```
Rakshith-R, line 235: There is very slight possibility that c.connected maybe set to false after it is checked here.

Any checks on c.connected need to be performed with mutex locks held.

I think just calling c.Connect() directly would be sufficient since you have
`if c.connected{}` within it already in line 87,

Rakshith-R (changes requested):
A tiny nit regarding ensureConnected() function

internal/connection/connection.go
black-dragon74: Done

nixpanic (dismissed):

internal/connection/connection.go
```diff
@@ -102,6 +224,12 @@ func (c *Connection) fetchCapabilities(ctx context.Context) error {
 	return nil
 }
 
+// ensureConnected re-eastablishes a connection if it is in a disconnected
+// state. An error is returned if it fails to renew the connection.
+func (c *Connection) ensureConnected() error {
+	return c.Connect()
+}
```
Rakshith-R, line 231: Directly call `c.Connect()` ?

Rakshith-R (changes requested):

internal/connection/connection.go
black-dragon74: Let's keep `ensureConnected`, for readability purposes? It is _a little_ more verbose about the intent in places outside of `connection.go`.

nixpanic (approved):

nixpanic:
@mergifyio rebase

PR #801, opened 2025-05-05
check vgrcontent before removing finalizer

Nikhil-Ladha:
we should check if vgrcontent is present before trying to remove finalizer. It is possible that it might have been already deleted in a previous reconcile loop.


internal/controller/replication.storage/volumereplication_controller.go
```diff
@@ -173,7 +173,7 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		vgr, vgrc, vgrErr = r.getVolumeGroupReplicationDataSource(logger, nameSpacedName)
 		if vgrErr != nil {
 			if errors.IsNotFound(vgrErr) && !instance.DeletionTimestamp.IsZero() {
-				logger.Info("volumeGroupReplication resource not found, as volumeReplication resource is getting garbage collected")
+				logger.Info("volumeGroupReplicationContent resource not found, as volumeReplication resource is getting garbage collected")
```
black-dragon74, line 176: vgrErr may denote both volumeGroupReplication and volumeGroupReplicationContent?

yati1998 (approved):

internal/controller/replication.storage/volumereplication_controller.go
Nikhil-Ladha: It can, though the possibility of that happening is _almost_ nil and can only happen if someone manually delete the VGR by removing the finalizers from it.
But, yeah no harm in having both the CRs mentioned in the log.

Nikhil-Ladha: Hence, only mentioned vgrContent in the log.

black-dragon74 (approved):

Issue #787, opened 2025-03-28
make pvc count in a group configurable


Nikhil-Ladha:
Let's make this configurable with default as 100  ? 
similar to https://github.com/csi-addons/kubernetes-csi-addons/blob/main/docs/csi-addons-config.md

_Originally posted by @Rakshith-R in https://github.com/csi-addons/kubernetes-csi-addons/pull/610#discussion_r2005354328_

PR #803, opened 2025-05-06
make pvc count in group configurable

Nikhil-Ladha:
Added support for configuring the pvc count in  a group. The min and max values are 1 and 100, anything less and more than that would result in an error in the controller manager.
Also, added necessary unit tests to validate the values in the config.

Fixes: #787


nixpanic:
Looks good, but 100 PVCs in a group sounds like really a lot. Are there any use-cases known that make groups, and do we know how large such a group is common?

Please update the documentation in `docs/csi-addons-config.md` too.

nixpanic:
> This is the upper limit that we would allow, and that is the max count that has been tested by ceph team too, some DR use cases would require somewhere around 60-80 at max.

_would allow_ in case of Ceph, I assume? If Ceph is the reference for the value, it would be good to mention that in a comment where the default is set in the code.

That is the only thing that I'd like to see improved, looks good otherwise.

Nikhil-Ladha:
> > This is the upper limit that we would allow, and that is the max count that has been tested by ceph team too, some DR use cases would require somewhere around 60-80 at max.
> 
> _would allow_ in case of Ceph, I assume? If Ceph is the reference for the value, it would be good to mention that in a comment where the default is set in the code.
> 
> That is the only thing that I'd like to see improved, looks good otherwise.

Ack, a note like the following in the `docs/csi-addons-config.yaml` should I think?
```
NOTE: `max-group-pvcs` default value is set based on ceph's support/testing. User can tweak this value based on the supported count for their storage vendor.
```

Nikhil-Ladha:
Added the note, please take a look now.
Thanks :)

docs/csi-addons-config.md
```diff
@@ -4,11 +4,13 @@ CSI-Addons Operator can consume configuration from a ConfigMap named `csi-addons
...
+| Option                      | Default value | Description                                      |
+| --------------------------- | ------------- | ------------------------------------------------ |
+| `reclaim-space-timeout`     | `"3m"`        | Timeout for reclaimspace operation               |
+| `max-concurrent-reconciles` | `"100"`       | Maximum number of concurrent reconciles          |
+| `max-group-pvcs`            | `"100"`       | Maximum number of PVCs allowed in a volume group |
 
 [`csi-addons-config` ConfigMap](../deploy/controller/csi-addons-config.yaml) is provided as an example.
 
 > Note: The operator pod needs to be restarted for any change in configuration to take effect.
+> NOTE: `max-group-pvcs` default value is set based on ceph's support/testing. User can tweak this value based on the supported count for their storage vendor.
```
nixpanic, line 16: nit: put an empty line between the two notes. When you check the visual diff, you can see that both **Note:** and **NOTE:** are on the same line. (Also use **Note:** like the previous one, not in ALL CAPS.)

image

nixpanic (approved):

Rakshith-R (approved):
Thanks

PR #820, opened 2025-06-20
Refactor schedule-precedence to allow user overrides

black-dragon74:
This patch refactors the schedule-precedence logic
to allow the value to be overridden by a ConfigMap.

The default mode of operation still prefers PVCs first
as source of key rotation and reclaimspace annotations.

The values for `schedule-precedence` keys are now concretely
defined as either `pvc` or `storageclass`.

### DIFF
--- a/PendingReleaseNotes.md
+++ b/PendingReleaseNotes.md
@@ -1,0 +1,15 @@
+# v0.13.0 Pending Release Notes
+
+## Breaking changes
+
+## Features
+
+- Allow override of precedence of key rotation and reclaim space related annotations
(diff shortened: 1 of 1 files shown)


Rakshith-R (changes requested):
> This patch modifies the manager deployment to include the default ConfigMap for csi-addons.
> 
> Having it this way will enable us to configure default values on Install without requiring any code changes. It will also enable us to have different values if needed between upstream and downstream.

If it comes in as part of csv or install yamls, won't the default values override whatever values the user has set in the configmap during upgrade ?

Rakshith-R (approved):
Thanks

PR #822, opened 2025-06-25
remove namespace name from pvc owner annotation

Nikhil-Ladha:
Remove namespace name from pvc owner annotation.
Name should be enough to identify vgr/vr the pvc belongs to as the annotation key for vr and vgr is different and both are namespaced scope CRs which would belong to the same namespace as PVC.

### DIFF
--- a/internal/controller/replication.storage/pvc_test.go
+++ b/internal/controller/replication.storage/pvc_test.go
@@ -18,7 +18,6 @@ package controller
 
 import (
 	"context"
-	"fmt"
 	"testing"
 
 	replicationv1alpha1 "github.com/csi-addons/kubernetes-csi-addons/api/replication.storage/v1alpha1"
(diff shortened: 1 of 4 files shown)


Nikhil-Ladha:
Created PR #823 to fix the Mergify config

nixpanic:
@mergifyio rebase

Nikhil-Ladha:
The CI looks good now, can we please have the changes reviewed as well?

Nikhil-Ladha:
> Looks good to me.
> 
> Is there a problem with the namespace prefix in the owner reference?

The change to have namespace in the annotation is not backward compatible. Upgrade won't work, as csi-addons would return a name mismatch for pvcs

Rakshith-R (approved):

Rakshith-R:
@mergifyio queue

PR #826, opened 2025-06-27
Add new status conditions to replication status

Nikhil-Ladha:
Added new status conditions which reflect the current replication status as returned from the storage vendor.


internal/controller/replication.storage/volumereplication_controller.go
```diff
@@ -449,6 +449,11 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 					instance.Status.LastSyncBytes = &tb
 				}
 			}
+			msg := info.StatusMessage
+			if msg == "" {
+				msg = "message not provided by vendor"
```
Madhu-1, line 454: Let keep it empty if its not set by csi driver.

Madhu-1:
@ShyamsundarR PTAL. @Nikhil-Ladha please check with Ramen team to ensure there is no regression in Ramen because of new condition

internal/controller/replication.storage/volumereplication_controller.go
Madhu-1: I think we should leave the behavior as it is, we just advertise what is been provider by the csi driver, i think message is optional one and omit empty, if nothing is set, we will skip it. If they dont see the value in the status its the same assumption that csi driver doesnt provided any value to disable to the user, may be thats what they want convey to the users.

internal/controller/replication.storage/volumereplication_controller.go
```diff
@@ -772,6 +773,17 @@ func setFailureCondition(instance *replicationv1alpha1.VolumeReplication, errMes
 	}
 }
 
+func setReplicationCondition(instance *replicationv1alpha1.VolumeReplication, dataSource, statusMessage string, replicationStatus proto.GetVolumeReplicationInfoResponse_Status) {
```
nixpanic, line 776: it is a bit unclear what `dataSource` is (untyped string), a comment or specific type would be helpful. That counts for this function, but also for the others that are called from here. A comment for a function should not just repeat it's name with spaces between the words, but be useful for someone to understand how/why it can be used.

api/replication.storage/v1alpha1/volumereplication_types.go
```diff
@@ -29,26 +29,31 @@ const (
 // "ConditionCompleted" means the condition is fulfilled.
 // "ConditionDegraded" means the condition is not fulfilled.
 // "ConditionResyncing" means the condition is resyncing.
+// "ConditionReplicating" means the condition is replicating.
```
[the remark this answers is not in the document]
Nikhil-Ladha: Removed the comment for now.

Nikhil-Ladha:
Updated the sidecar code to return the new status fields in the response.

Test results:
```
    - lastTransitionTime: "2025-07-03T05:00:17Z"
      message: 'volume is replicating: local image is primary'
      observedGeneration: 2
      reason: Replicating
      status: "True"
      type: Replicating
```

Nikhil-Ladha:
If everything looks can we get this PR merged, so that ramen PR can also be reviewed and tested?

internal/controller/replication.storage/volumereplication_controller.go
```diff
@@ -449,6 +449,10 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 					instance.Status.LastSyncBytes = &tb
 				}
 			}
+			// Check for nil, as info might be nil for knownGRPCErrors
+			if info != nil {
+				setReplicationCondition(vr.instance, vr.instance.Spec.DataSource.Kind, info.StatusMessage, info.Status)
```
Madhu-1, line 454: Add a comment saying that these conditions only will be set when the VR is in primary state not in secondary state. later this can be moved outside when we want this condition to be set irrespective of the state.

Nikhil-Ladha: Done

internal/controller/replication.storage/volumereplication_controller.go
```diff
@@ -772,6 +777,18 @@ func setFailureCondition(instance *replicationv1alpha1.VolumeReplication, errMes
...
+// setReplicationCondition sets the replication condition in VR's status.Conditions for the PVC/VolumeGroup dataSource based on the replicationStatus value.
+func setReplicationCondition(instance *replicationv1alpha1.VolumeReplication, dataSource, statusMessage string, replicationStatus proto.GetVolumeReplicationInfoResponse_Status) {
+	switch replicationStatus {
+	case proto.GetVolumeReplicationInfoResponse_UNKNOWN:
+		setUnknownReplicationCondition(&instance.Status.Conditions, instance.Generation, dataSource, statusMessage)
+	case proto.GetVolumeReplicationInfoResponse_HEALTHY:
+		setHealthyReplicationCondition(&instance.Status.Conditions, instance.Generation, dataSource, statusMessage)
+	case proto.GetVolumeReplicationInfoResponse_DEGRADED, proto.GetVolumeReplicationInfoResponse_ERROR:
+		setDegradedReplicationCondition(&instance.Status.Conditions, instance.Generation, dataSource, statusMessage)
+	}
```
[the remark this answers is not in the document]
nixpanic: Oh, wow, lol, maybe it is really needed in that case? There are only 3 of those 4 cases listed here :thinking:

PR #829, opened 2025-07-09
fix promotion condtion update for VR

Nikhil-Ladha:
Fix promotion condition update for VR due to intermittent condition updates


Madhu-1 (approved):
LGTM

Madhu-1:
The VR is already in primary state
The VRC is deleted and when the VR reconcile happens it sets the status to promote state as its not able to get the VRC
VRC is recreated and the reconcile triggers again
Here as the VR state is primary in both status and in the CR , calling the volumereplication is skipped and we dont update the status.

Please test the changes as per the RCA and update the result here?

Issue #815, opened 2025-05-27
Optimize memory footprint of csi-addons


black-dragon74:
With the option to specify `precedence` in the `ConfigMap`, we have an opportunity to optimize the memory footprint of `csi-addons`.

If the precedence is set to `sc-only`, there is no need to watch `PVC` resources. The current `PVC` controller handles most of the heavy lifting, but it can be refactored into two separate controllers. 

One controller will watch `StorageClass` objects, and the other will watch `PersistentVolumeClaim` objects. Only one of these controllers needs to run, depending on the specified `precedence`.

We can further reduce memory usage by watching only [PartialObjectMetadata](https://github.com/kubernetes/apimachinery/blob/d5c9711b77ee5a0dde0fef41c9ca86a67f5ddb4e/pkg/apis/meta/v1/types.go#L1496) instead of the full object.

**To summarize:**

- Split the controller into separate ones for `StorageClass` and `PVC`
- Run only one controller based on the configured precedence
- Watch only one resource `Kind`, not both
- Use `PartialObjectMetadata` to avoid caching full objects

nixpanic:
Is there any loss in functionality when only one of the controllers will be enabled? What would be the advantage to enable one over an other? Is there a need to have both controllers at all?

Nit: `precedence` suggests to me that something needs to _exist before_ something new, maybe an other name for the option would be clearer.

PR #833, opened 2025-07-17
add support for agnostic storage vendor in VGR

Nikhil-Ladha:
add a new field named `External` which when set to `true` means the VGR won't be reconciled by csi-addons controller and instead it should be managed by the storage vendor specific controller.

Also, make VolumeReplicationClassName optional in VGR's spec such that external vendors are not tied up to VR specific implementation.

### DIFF
--- a/api/replication.storage/v1alpha1/volumegroupreplication_types.go
+++ b/api/replication.storage/v1alpha1/volumegroupreplication_types.go
@@ -34,7 +34,7 @@ type VolumeGroupReplicationSpec struct {
 
 	// volumeReplicationClassName is the volumeReplicationClass name for the VolumeReplication object
 	// created for this volumeGroupReplication
-	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Optional
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="volumeReplicationClassName is immutable"
 	VolumeReplicationClassName string `json:"volumeReplicationClassName"`
 
@@ -64,6 +64,12 @@ type VolumeGroupReplicationSpec struct {
 	// ReplicationState is "secondary"
 	// +kubebuilder:default:=false
 	AutoResync bool `json:"autoResync"`
+
+	// External represents if VolumeGroupReplication should be reconciled by the csi-addons controller
+	// or an external controller managed by the storage vendor.
+	// +kubebuilder:default:=false
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="source is immutable"
+	External bool `json:"external,omitempty"`
 }
 
 // VolumeGroupReplicationSource specifies the source for the the volumeGroupReplication
--- a/internal/controller/replication.storage/volumegroupreplication_controller.go
+++ b/internal/controller/replication.storage/volumegroupreplication_controller.go
@@ -111,6 +111,12 @@ func (r *VolumeGroupReplicationReconciler) Reconcile(ctx context.Context, req ct
 		return reconcile.Result{}, err
 	}
 
+	// Skip reconcile if external VGR
+	if instance.Spec.External {
+		r.log.Info("skipping reconcile for non csi-addons managed VGR")
+		return reconcile.Result{}, nil
+	}
+
 	// Get VolumeGroupReplicationClass instance
 	vgrClassObj, err := r.getVolumeGroupReplicationClass(instance.Spec.VolumeGroupReplicationClassName)
 	if err != nil {
@@ -424,6 +430,10 @@ func (r *VolumeGroupReplicationReconciler) SetupWithManager(mgr ctrl.Manager) er
 
 			// Check if the pvc labels match any VGRs based on selectors present in it's annotation
 			for _, vgr := range vgrObjsList.Items {
+				// Skip reconcile for PVC that belongs to an external VGR
+				if vgr.Spec.External {
+					continue
+				}
 				if vgr.Annotations != nil && vgr.Annotations[pvcSelector] != "" {
 					labelSelector, err := labels.Parse(vgr.Annotations[pvcSelector])
 					if err != nil {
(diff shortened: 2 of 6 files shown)


black-dragon74 (commented):
Why does this need to be a API change and not something like: https://github.com/csi-addons/kubernetes-csi-addons/blob/5100f91e506c3c33364d28a644c4e432299318b4/internal/controller/csiaddons/persistentvolumeclaim_controller.go#L78

When set to `unmanaged` it means we won't reconcile.

Madhu-1:
> Why does this need to be a API change and not something like:
> 
> https://github.com/csi-addons/kubernetes-csi-addons/blob/5100f91e506c3c33364d28a644c4e432299318b4/internal/controller/csiaddons/persistentvolumeclaim_controller.go#L78
> 
> When set to `unmanaged` it means we won't reconcile.

This is more of a workaround implementation not a direct API change. Having it in the CR spec is strict rule

api/replication.storage/v1alpha1/volumegroupreplication_types.go
```diff
@@ -64,6 +63,11 @@ type VolumeGroupReplicationSpec struct {
 	// ReplicationState is "secondary"
 	// +kubebuilder:default:=false
 	AutoResync bool `json:"autoResync"`
+
+	// External represents if VolumeGroupReplication should be reconciled by the csi-addons controller
```
Madhu-1, line 67: should not be reconciled by the csi-addons controller?

api/replication.storage/v1alpha1/volumegroupreplication_types.go
```diff
@@ -64,6 +63,11 @@ type VolumeGroupReplicationSpec struct {
...
+	// or an external controller managed by the storage vendor.
+	// +kubebuilder:default:=false
+	External bool `json:"external,omitempty"`
```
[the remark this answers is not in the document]
Nikhil-Ladha: Yes, updated it now.

Madhu-1 (approved):

internal/controller/replication.storage/volumegroupreplication_controller.go
```diff
@@ -111,6 +111,12 @@ func (r *VolumeGroupReplicationReconciler) Reconcile(ctx context.Context, req ct
 		return reconcile.Result{}, err
 	}
 
+	// Skip reconcile if external VGR
+	if instance.Spec.External {
+		r.log.Info("skipping reconcile for non csi-addons managed VGR")
```
iPraveenParihar, line 116: could be improved as below WDYT?

```suggestion
		r.log.Infof("reconciliation skipped: VGR %s is externally managed", instance.Name)
```

Nikhil-Ladha: If it's just for name then, we are loggin the name just before this check.

iPraveenParihar: Ahh, missed the context above 👍

api/replication.storage/v1alpha1/volumegroupreplication_types.go
```diff
@@ -34,8 +34,7 @@ type VolumeGroupReplicationSpec struct {
 
 	// volumeReplicationClassName is the volumeReplicationClass name for the VolumeReplication object
 	// created for this volumeGroupReplication
-	// +kubebuilder:validation:Required
-	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="volumeReplicationClassName is immutable"
+	// +kubebuilder:validation:Optional
```
ShyamsundarR, line 37: We can have immutability built into the field, need not lose that. Just the field needs to be optional and not mandatory.

internal/controller/replication.storage/volumegroupreplication_controller.go
```diff
@@ -111,6 +111,12 @@ func (r *VolumeGroupReplicationReconciler) Reconcile(ctx context.Context, req ct
 		return reconcile.Result{}, err
 	}
 
+	// Skip reconcile if external VGR
+	if instance.Spec.External {
```
ShyamsundarR, line 115: I would add a similar check [here](https://github.com/csi-addons/kubernetes-csi-addons/blob/5100f91e506c3c33364d28a644c4e432299318b4/internal/controller/replication.storage/volumegroupreplication_controller.go#L426) such that PVCs that belong to  VGR that is external does not trigger a reconcile for the VGR by this controller.

ShyamsundarR (commented):
LGTM! couple of improvements, will let others decide on applicability and merge.

internal/controller/replication.storage/volumegroupreplication_controller.go
```diff
@@ -424,6 +430,10 @@ func (r *VolumeGroupReplicationReconciler) SetupWithManager(mgr ctrl.Manager) er
 
 			// Check if the pvc labels match any VGRs based on selectors present in it's annotation
 			for _, vgr := range vgrObjsList.Items {
+				// Skip reconcile for PVC that belongs to an external VGR
+				if vgr.Spec.External {
```
Madhu-1, line 434: For now this is good but even we need more enhancements for example not even cache this object when the External field is set

Madhu-1: https://github.com/csi-addons/kubernetes-csi-addons/issues/815 pointer it here so that the main issue contains a link to this PR where we need to address cache problem

Nikhil-Ladha: For the time being I have added the check in PVC predicate, it is beneficial in all the cases.

Nikhil-Ladha:
> https://github.com/csi-addons/kubernetes-csi-addons/blob/main/docs/volumegroupreplication.md
> 
> Document the behavior please and add release notes.

@Rakshith-R done.

internal/controller/replication.storage/volumegroupreplication_controller.go
Nikhil-Ladha: Done

Madhu-1 (approved):

PR #844, opened 2025-07-25
config: add terminationMessagePolicy to csi-addons

ShravaniVangur:
In Kubernetes, termination messages allow containers to communicate the reason for termination, which might be useful for debugging.

#### By Default:
Unless explicitly configured, Kubernetes defaults to the following:
```
terminationMessagePath: /dev/termination-log
terminationMessagePolicy: File
```
This indicates that if a container exits with an error and the `/dev/termination-log` file is empty, the termination message in the Pod status will be blank.

#### By setting `terminationMessagePolicy` to `FallbackToLogsOnError`:
If `/dev/termination-log` is empty and the container exits with an error, Kubernetes will use the last few lines of the container’s logs as the termination message, thus providing information when the termination log file does not have any.

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -50,6 +50,7 @@ spec:
           securityContext:
             allowPrivilegeEscalation: false
             readOnlyRootFilesystem: true
+          terminationMessagePolicy: FallbackToLogsOnError
           livenessProbe:
             httpGet:
               path: /healthz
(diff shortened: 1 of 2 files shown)


ShravaniVangur:
> @ShravaniVangur Can you please explain the details about the changes and why it is set what happens if its not set and what is the default behaviour in kubernetes

@Madhu-1 Sure. Addressed this in the description. PTAL.

Madhu-1 (approved):

PR #862, opened 2025-08-14
add replicating condition if error returned for GetVolumeReplicationInfo call

Nikhil-Ladha:
We should add replicating condition to VR/VGR status if GetVolumeReplicationInfo call returned an error, so that we are able to report the current condition to the user.

### DIFF
--- a/internal/controller/replication.storage/volumereplication_controller.go
+++ b/internal/controller/replication.storage/volumereplication_controller.go
@@ -463,6 +463,12 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 			requeueForInfo = true
 		} else if !util.IsUnimplementedError(err) {
 			logger.Error(err, "Failed to get volume replication info")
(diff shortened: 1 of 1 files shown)


internal/controller/replication.storage/volumereplication_controller.go
```diff
@@ -463,6 +463,12 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 			requeueForInfo = true
 		} else if !util.IsUnimplementedError(err) {
 			logger.Error(err, "Failed to get volume replication info")
+			// Set UNKNOWN Replicating condition if any error occurred while fetching volume info.
+			setReplicationCondition(vr.instance, vr.instance.Spec.DataSource.Kind, "status not found", proto.GetVolumeReplicationInfoResponse_UNKNOWN)
```
nixpanic, line 467: I am not sure "status not found" is correct, why not use the contents of `err`?

PR #867, opened 2025-08-26
remove replicating condition when state is secondary

Nikhil-Ladha:
When a VR/VGR is demoted to secondary, we don't call the `VolumeInfo` on the volume/group anymore, so the older conditions are not updated and will show stale data. We should instead remove replicating condition from VR/VGR status when the image/group is demoted to secondary state.


Nikhil-Ladha:
> @Nikhil-Ladha, could you please add why we need to do this in the PR description? Just for the context

Done

PR #868, opened 2025-09-02
remove pvc finalizers only after VGR is deleted

Nikhil-Ladha:
finalizers added to PVCs part of VGR should only be removed after the VGR is deleted, else there can be a race condition where the VGR is not deleted yet as the group is still undergoing disabling mirroring and modification operations to remove volume from group and then deletion. Therefore, we shouldn't be removing the finalizers from PVC before the group is deleted.


iPraveenParihar (approved):

Rakshith-R (changes requested):
What if the pod restarts after removing finalizer on VGR ? 
The finalizers on PVCs will stay forever.

Nikhil-Ladha:
> What if the pod restarts after removing finalizer on VGR ? The finalizers on PVCs will stay forever.

The probability of such a scenario is very less, but anyway I have updated the code to handle the situation in a better way.
PTAL, thanks.

Nikhil-Ladha:
I have updated the code to wait for both the VR, VGRContent to be deleted before removing finalizers and owner annotations from PVC by moving the owner annotation check in RemoveVGRFinalizer to a separate individual function named `isSafeToDeleteVGR` which basically checks if both the dependent CRs are deleted and their owner annotations are removed from the VGR.
This will make sure that the PVC finalizers are removed only when VR, VGRContent are deleted and VGR deletion can happen as well

PR #907, opened 2025-11-24
Optimizations for controller memory usage

black-dragon74:
This PR aims to optimize memory consumption and allocation for csi-addons controller. This goal is achieved mainly by:

- Disabling cache for resources that we do not need frequently.
- Setting up shared informers on every resource that we Get/List frequently (VAs / PVs)
- A new simplified PVC controller without the overheads that came with old way of operation, mainly:
  - Supports and enabled only when `schedule-precedence` is set to `storageclass`.
  - Uses deterministic naming for resources to avoid manual housekeeping of stale resources**
  - Ditches the extra field indexers on created resources (side-effect of above)
  - No update operations on PVCs, state-to-reconcile is always calculated with SC as source of truth.
- A little housekeeping by moving common functions/variables into their own `controller/utils` package.

Closes: #815 

---

### DIFF
--- a/internal/controller/csiaddons/pvc_new_controller.go
+++ b/internal/controller/csiaddons/pvc_new_controller.go
@@ -105,13 +105,25 @@ func (r *PVCReconiler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Re
 
 	// Reconcile for dependent features
 	// Reconcile - Key rotation
+	keyRotationName := fmt.Sprintf("%s-keyrotation", pvc.Name)
 	keyRotationSched := sc.Annotations[utils.KrcJobScheduleTimeAnnotation]
 	keyRotationEnabled := sc.Annotations[utils.KrEnableAnnotation]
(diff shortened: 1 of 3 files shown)


internal/controller/utils/reclaimspace_keyrotation_utils.go
```diff
@@ -0,0 +1,164 @@
...
+		Spec: csiaddonsv1alpha1.ReclaimSpaceJobSpec{
+			Target:               csiaddonsv1alpha1.TargetSpec{PersistentVolumeClaim: pvcName},
+			BackoffLimit:         DefaultBackoffLimit,
+			RetryDeadlineSeconds: DefaultRetryDeadlineSeconds,
+		},
+	}
+}
+
+func SetSpec(obj client.Object, schedule, pvcName string) {
+
```
Madhu-1: Remove new line

Madhu-1 (approved):

PR #949, opened 2026-01-13
feat: Use staggered schedule for CronJobs

black-dragon74:
This patch includes the following changes around the determined schedule for created (ReclaimSpace/KeyRotation)CronJobs:

- Prevents thundering-herd behavior by staggering the schedule by a max offset of 2hrs.
  - If a job's schedule requires it to run more frequently, the stagger window is adjusted accordingly. 
  - IOW, if schedule < 2hrs; staggerWindow = schedule
- The offset is calculated deterministically using UID to ensure the same offset b/w multiple reconciles.
- If a job has missed its last execution, the stagger is not applied to let the job run immediately (might lead to thundering-herd but a fair tradeoff in happy path?)

Staggering can be disabled by setting the launch arg/config map key `cronjob-stagger-window` to `0`.

### DIFF
--- a/test/e2e/encryptionkeyrotation/encryptionkeyrotation_test.go
+++ b/test/e2e/encryptionkeyrotation/encryptionkeyrotation_test.go
@@ -169,7 +169,7 @@ var _ = ginkgo.Describe("EncryptionKeyRotation", ginkgo.Ordered, func() {
 			gomega.Expect(cronJob).NotTo(gomega.BeNil())
 
 			ginkgo.By("Waiting for the first EncryptionKeyRotationJob to be created")
-			createdJob := f.WaitForEncryptionKeyRotationJobCreation(cronJob.Name, 90*time.Second)
+			createdJob := f.WaitForEncryptionKeyRotationJobCreation(cronJob.Name, 3*time.Minute)
 			gomega.Expect(createdJob).NotTo(gomega.BeNil())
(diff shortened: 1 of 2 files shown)


internal/controller/utils/reclaimspace_keyrotation.go
```diff
@@ -185,3 +189,33 @@ func CleanOldJobs(
 
 	return shouldRequeue, nil
 }
+
+// GetStaggeredNext returns a deterministic, UID-based staggered time computed from a schedule.
```
nixpanic: Maybe place this in a file for cronjob or scheduler functions?

internal/controller/utils/reclaimspace_keyrotation.go
```diff
@@ -128,3 +132,33 @@ func ExtractOwnerNameFromPVCObj[T client.Object](rawObj client.Object) []string
 
 	return []string{owner.Name}
 }
+
+// GetStaggeredNext returns a deterministic, UID-based staggered time computed from a schedule.
+// The staggered window is capped at a maximum of 2hrs but is adjusted for smaller intervals.
+func GetStaggeredNext(uid apiTypes.UID, nextTime time.Time, sched cron.Schedule) time.Time {
```
Rakshith-R: Move this to a new file ?

Madhu-1 (approved):

PR #951, opened 2026-01-19
bundle: add ability to inject priorityClassName

black-dragon74:
This patch adds the ability to inject a `priorityClassName` to the generated bundle.

Ref: https://issues.redhat.com/browse/DFBUGS-394

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -126,7 +126,9 @@ gen-csv-base:
 
 .PHONY: bundle
 bundle: gen-csv-base manifests operator-sdk
+	cd config/manifests && $(KUSTOMIZE) edit add patch --path manager_priority_patch.yaml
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -126,7 +126,18 @@ gen-csv-base:
 
 .PHONY: bundle
 bundle: gen-csv-base manifests operator-sdk
-	$(KUSTOMIZE) build config/manifests | $(OPERATOR_SDK) generate bundle --manifests --metadata --package=$(PACKAGE_NAME) $(BUNDLE_VERSION)
+	@TEMP=$$(mktemp -d) && \
```
nixpanic: This is a `Makefile`, there should be no need for `&& \` kind of formatting. Each command is executed, and if there is an error the remaining commands will not run.

Running this in a tempdir is rather ugly. Is that really required?

It looks like this is always done when building the bundle. There is no difference in including the `priorityClassName` by default... Maybe just add the patch to `kustomize.yaml` but leave it commented out?

PR #955, opened 2026-01-20
controller: remove the shim for GCing old resources

black-dragon74:
We can remove this shim which was used when upgrading from v0.13.0 to v0.14.0. 

We had a release of v0.14.0 a while ago and the next version does not require this shim.

Reverts (changes of): 0b93f399e3a660653bebd88a043f8af6a6b8f393


nixpanic (changes requested):
The commit message mentions _4.19 to 4.20_ versions. These are not versions available in this project. Please use correct version numbers, and not downstream product ones.

nixpanic (approved):

Rakshith-R (approved):

PR #957, opened 2026-01-21
bundle: add ability to inject priorityClassName

black-dragon74:
This patch adds the ability to inject a priorityClassName to the generated bundle.


Rakshith-R (commented):
Let's set it to system-cluster-critical by default going by how important functions csi-addons handle ?
Users' can override it if they want to anything else.

@Madhu-1  ?

Nikhil-Ladha:
Let's have this enabled by default instead of commenting it out, we already have an old bug that needs this fix: https://issues.redhat.com/browse/DFBUGS-394

PR #985, opened 2026-03-06
docs: update design doc to include VR/VGR status samples

Nikhil-Ladha:
add VR/VGR status,status.Condition samples for different states.

### DIFF
--- a/docs/design/volumegroupreplication.md
+++ b/docs/design/volumegroupreplication.md
@@ -65,7 +65,918 @@ For reference, one can check the implementation of csi-addons controllers for th
 - [volumegroupreplication_controller](https://github.com/csi-addons/kubernetes-csi-addons/blob/main/internal/controller/replication.storage/volumegroupreplication_controller.go) - It manages the flow of VGR and necessary validations to keep the VGR as the source of truth for the user.
 - [volumegroupreplicationcontent_controller](https://github.com/csi-addons/kubernetes-csi-addons/blob/main/internal/controller/replication.storage/volumegroupreplicationcontent_controller.go) - It manages the group level operations for the VGR.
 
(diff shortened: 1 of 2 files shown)


Madhu-1 (commented):
i think we need to make a workflow to Failover and failback. What conditions should DR orchestrator look for 

For example, like Below

* Create VR
* Wait for the conditions to match below
* Change the VR state to secondary
* Wait for the specific condition to match
* Create VR on secondary cluster
* Wait for the conditions to match

somethink like above for both failover and Relocate

docs/design/volumegroupreplication.md
```diff
@@ -65,7 +65,533 @@ For reference, one can check the implementation of csi-addons controllers for th
...
+2. **Secondary Cluster Healthy**: The secondary cluster is operational and accessible
+3. **VolumeGroupReplication Exists**: A VolumeGroupReplication resource must exist on the secondary cluster (see note below)
+4. **VolumeGroupReplication Status**: The VGR on the secondary cluster should show one of these states:
+   - `state: Secondary` with `Degraded: True` and `Completed: True`
+   - `state: Secondary` with `Resyncing: True` (data may be slightly out of sync)
+
+**Important Note**: The VolumeGroupReplication resource might not be created on the secondary cluster in some deployment scenarios. If the VGR does not exist on the secondary cluster, you must create it before proceeding with the failover.
+The VGR should use the same label selector to identify PVCs and reference the appropriate VolumeReplicationClass and VolumeGroupReplicationClass with `replicationState: secondary`.
+
+#### Failover Steps
```
Madhu-1: Let's cover some samples for pre-failover on what it looks like.

Madhu-1 (approved):

Madhu-1:
we can improvise as requried later

fOO223Fr:
The workflows deserves a seperate doc of their own. 1 document per workflow and then a indexing doc which links all workflows. otherwise one document become too long.

PR #1020, opened 2026-04-14
Allow ReclaimSpace to be disabled from StorageClasses

black-dragon74:
This patch allows a user to disable reclaimspace from the annotations
on the StorageClass.

This is required for RWX block mode volumes used in VMs (to prevent sparsify op from blacklisting the actual client) .

### DIFF
--- a/docs/reclaimspace.md
+++ b/docs/reclaimspace.md
@@ -180,3 +180,13 @@ To disable reclaim space for a specific PersistentVolumeClaim (PVC), follow thes
    ```
 
 These changes will disable reclaim space for the specified PVC.
+
+### Disabling Reclaim Space for All PersistentVolumeClaims in a StorageClass
+
+To disable reclaim space for all PVCs in a particular StorageClass, annotate the StorageClass with `reclaimspace.csiaddons.openshift.io/enable: "false"`.
+
+```bash
+kubectl annotate storageclass <SC_NAME> "reclaimspace.csiaddons.openshift.io/enable=false"
+```
+
+This action will disable reclaim space across all PVCs in that StorageClass and remove any existing `ReclaimSpaceCronJob` CRs for PVCs within it.
--- a/internal/controller/csiaddons/persistentvolumeclaim_controller.go
+++ b/internal/controller/csiaddons/persistentvolumeclaim_controller.go
@@ -295,6 +295,7 @@ func (r *PersistentVolumeClaimReconciler) storageClassEventHandler() handler.Eve
 				utils.RsCronJobScheduleTimeAnnotation,
 				utils.KrcJobScheduleTimeAnnotation,
 				utils.KrEnableAnnotation,
+				utils.RsEnableAnnotation,
 			}
 
 			var requests []reconcile.Request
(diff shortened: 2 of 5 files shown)


nixpanic (changes requested):
I guess this makes sense. Only need to add some user facing documentation for the new annotation.

black-dragon74:
> I guess this makes sense. Only need to add some user facing documentation for the new annotation.

Done :)

docs/reclaimspace.md
```diff
@@ -180,3 +180,13 @@ To disable reclaim space for a specific PersistentVolumeClaim (PVC), follow thes
...
+
+### Disabling Reclaim Space for All PersistentVolumeClaims in a StorageClass
+
+To disable reclaim space for all PVCs in a particular StorageClass, annotate the StorageClass with `reclaimspace.csiaddons.openshift.io/enable: "false"`.
+
+```bash
+kubectl annotate storageclass <SC_NAME> "reclaimspace.csiaddons.openshift.io/enable=false"
+```
+
+This action will disable reclaim space across all PVCs in that StorageClass and remove any existing `ReclaimSpaceCronJob` CRs for PVCs within it.
```
nixpanic, line 192: what happens when there is an annotation on a Namespace?

Rakshith-R (dismissed):
LGTM

please make sure no edge case is missed

docs/reclaimspace.md
black-dragon74: It will still be disabled. `checkDisabledByAnnotation` checks on PVC > NS > StorageClass when precedence is set to `pvc`(default).

internal/controller/csiaddons/persistentvolumeclaim_controller.go
iPraveenParihar, line 294: you need to add the enable annotation here too?

iPraveenParihar: need to add here too?

internal/controller/csiaddons/persistentvolumeclaim_controller.go
```diff
@@ -585,6 +585,25 @@ func (r *PersistentVolumeClaimReconciler) processReclaimSpace(
...
+		return ctrl.Result{}, err
+	}
+
+	if disabled {
+		if rsCronJob != nil {
+			err = r.Delete(ctx, rsCronJob)
+			if client.IgnoreNotFound(err) != nil {
+				errMsg := "failed to delete ReclaimSpaceCronJob"
+				logger.Error(err, errMsg)
+				return ctrl.Result{}, fmt.Errorf("%w: %s", err, errMsg)
```
iPraveenParihar, line 600: this is inverted order. <error message>: <error>

iPraveenParihar (changes requested):

internal/controller/csiaddons/persistentvolumeclaim_controller.go
black-dragon74: I remember a review comment by Niels where he prefers err and then errMsg. As this is unrelated to this PR, let's keep it as-is?

internal/controller/csiaddons/persistentvolumeclaim_controller.go
black-dragon74: Ah excellent catch! Thank you!

internal/controller/csiaddons/persistentvolumeclaim_controller.go
nixpanic: yes, but that preference is only related to static errors:

```go
var SomeErr = errors.New("some error occured")

func Fails() error {
    err := it.failed()
    if err != nil {
        return fmt.Errorf("%w: %v", SomeErr, err)
    }
 
    return nil
}
```

In all other cases where an error is returned from a function, and some additional context is added without it being an `error` type, the normal formatting has my preference:

```go
func Fails() error {
    err := it.failed()
    if err != nil {
        msg := "something failed"
        return fmt.Errorf("%s: %w", msg, err)
    }
 
    return nil
}
```

Rakshith-R (approved):

nixpanic (approved):

nixpanic:
@iPraveenParihar you have a review outstanding on this PR. Please check and unblock it, thanks!

Madhu-1:
Why to disable the complete feature instead of disabling ControllerReclaim?

black-dragon74:
> Why to disable the complete feature instead of disabling ControllerReclaim?

Node ReclaimSpace just TRIMs, which zeroes out the blocks. Sparsify is the one that discards those blocks. So what's the point of running JUST the `fstrim` when the RADOS objects are still not released and are consuming space on OSDs?

Rakshith-R:
> > Why to disable the complete feature instead of disabling ControllerReclaim?
> 
> Node ReclaimSpace just TRIMs, which zeroes out the blocks. Sparsify is the one that discards those blocks. So what's the point of running JUST the `fstrim` when the RADOS objects are still not released and are consuming space on OSDs?

Not exactly:

From CSI-Addons perspective: it issues controller reclaimspace and if pvc is mounted, node reclaimspace.

In our case with Block Mode:
- node reclaimspace is a no-op
- controller reclaimspace: sparsify is the one causing the problem.

Just disabling controller Reclaimspace will just lead to useless CRs and reconcile loops.

Therefore, this pr precisely solves the problem.

iPraveenParihar (approved):

PR #1022, opened 2026-04-16
fix: reuse healthy connections in CSIAddonsNode reconciler

black-dragon74:
The reconciler created a new gRPC connection on every requeue(1hr) closing the existing one via ConnPool.Put().

This disrupted sidecars using that connect with ExitOnConnectionLoss causing restarts.

This patches fixes the issue by ensuring that a new connection is only created when the pool has no entry, the endpoint changed, or the existing connection is unhealthy.

### DIFF
--- a/internal/connection/connection_pool.go
+++ b/internal/connection/connection_pool.go
@@ -22,10 +22,10 @@ import (
 	"log"
 	"sync"
 
+	"github.com/csi-addons/kubernetes-csi-addons/internal/util"
+
 	coordination "k8s.io/api/coordination/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-
-	"github.com/csi-addons/kubernetes-csi-addons/internal/util"
 )
 
 const failedToReconnectFmtStr = "failed to reconnect an inactive connection due to error: %w"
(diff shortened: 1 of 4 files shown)


internal/controller/csiaddons/csiaddonsnode_controller.go
```diff
@@ -158,15 +158,39 @@ func (r *CSIAddonsNodeReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		return ctrl.Result{}, err
 	}
 
-	logger.Info("Connecting to sidecar")
-	newConn, err := connection.NewConnection(ctx, endPoint, nodeID, driverName, csiAddonsNode.Namespace, csiAddonsNode.Name, r.EnableAuth)
+	// Check if we already have a connection for this endpoint
+	newConn := r.ConnPool.GetByKey(key)
```
nixpanic, line 162: is there a good reason to do all the connection handling here, and not immediately in `connection.NewConnection()`? That could return an existing connection in case it is healthy. It feels cleaner to do all the connection checking inside the connection package.

black-dragon74: I thought of it initially but then `NewConnection` becomes misleading as it should always return a "new" connection, created from the scratch?

Moreover, this quirk is specific to CSIAddonsNode reconciler.

_I am happy to move things around if required :)_

nixpanic: It feels cleaner to me if all the weirdness of connection handling is done inside the connection package. If you don't like to do this in `connection.NewConnection()`, maybe add `connection.GetConnection()` and describe the differences in the function comments?

black-dragon74: PTAL. Should me much better now

nixpanic (approved):
Thanks, much nicer to read the usage in `Reconcile()`.

internal/connection/connection_pool.go
```diff
@@ -50,6 +50,42 @@ func NewConnectionPool() *ConnectionPool {
...
+	cp.rwlock.RLock()
+	existing := cp.pool[key]
+	cp.rwlock.RUnlock()
+
+	// If a connection exists for the same endpoint
+	if existing != nil && existing.Endpoint() == endpoint {
+		// Check if it is actually healthy by calling Connect(), err will be nil
+		if err := existing.Connect(); err == nil {
+			return existing, nil
+		}
```
Rakshith-R, line 74: should we close unhealthy existing connection ?

black-dragon74: `Put` call down the lines does that.

Rakshith-R: 🤔 seems to get closed on error in https://github.com/black-dragon74/kubernetes-csi-addons/blob/d7275ebff022bfe2692df280ea2d2ebe39e973f8/internal/connection/connection.go#L65

Rakshith-R (approved):
Thanks, LGMT !

internal/connection/connection_pool.go
black-dragon74: We close a healthy connection when we cannot fetch caps, Put will attempt to close the connection again but will ignore the errors. It's a safetynet to ensure we terminate old connection before we store a new one.

PR #993, opened 2026-03-10
doc: add design for replication destination info

Madhu-1:
Adding a design document on how to handle the DR when the volumeID's and volumeGroupID's are not same on the destination cluster.

PR #1029, opened 2026-04-29
feat: Add support for GetReplicationDestinationInfo 

nicknevin:
Add support for GetReplicationDestinationInfo based on the design [here](https://github.com/csi-addons/kubernetes-csi-addons/pull/993).

Note: This should not be merged until the corresponding spec changes (see https://github.com/csi-addons/spec/pull/79) have been merged and go.mod updated to refer to them.


api/replication.storage/v1alpha1/volumegroupreplicationcontent_types.go
```diff
@@ -78,12 +78,46 @@ type VolumeGroupReplicationContentSource struct {
...
+// PersistentVolumeMapping contains the PV reference along with its
+// source and destination volume handles. The destination handle is
+// populated only when the SP supports GET_REPLICATION_DESTINATION_INFO
+// and destination info is available.
+type PersistentVolumeMapping struct {
+	// Name is the name of the PersistentVolume.
+	Name string `json:"name"`
+
+	// VolumeHandle is the CSI volume handle of this PV on the
+	// source/primary cluster (i.e. PV.Spec.CSI.VolumeHandle).
```
Nikhil-Ladha: The CR could be in secondary state and still have the source handle.
Let's not mention "primary" here.

build/Containerfile.tools
```diff
@@ -1,6 +1,6 @@
 # DL3007 ignored as it asks to pin the version explicitly to a release tag
 # hadolint ignore=DL3007
-FROM quay.io/fedora/fedora:latest
+FROM quay.io/fedora/fedora:44
```
Nikhil-Ladha: Why this change?

internal/controller/replication.storage/volumegroupreplication_controller.go
```diff
@@ -326,6 +329,25 @@ func (r *VolumeGroupReplicationReconciler) Reconcile(ctx context.Context, req ct
...
+				allCovered = false
+				break
+			}
+		}
+		if allCovered {
+			setDestinationInfoAvailableCondition(&instance.Status.Conditions,
+				instance.Generation, volumeGroupReplicationDataSource)
+		}
+	}
+
```
Nikhil-Ladha: We should do this only if the driver support the `GET_REPLICATION_DESTINATION_INFO` capability, otherwise we shouldn't be adding/checking for the condition.
Same applies above as well where we are sending the `setDestinationInfoPendingCondition` condition.

Nikhil-Ladha: We should call the `supportsGetReplicationDestinationInfo` function at the start of the reconcile function and use the returned bool var to check at different stages before trying to add/update the conditions in the CR.

internal/controller/replication.storage/volumegroupreplicationcontent_controller.go
```diff
@@ -211,17 +211,62 @@ func (r *VolumeGroupReplicationContentReconciler) Reconcile(ctx context.Context,
...
+				VolumeHandle: handle,
+			})
 		}
 	}
 	instance.Status.PersistentVolumeRefList = pvRefList
+	instance.Status.PersistentVolumeMappingList = pvMappings
+
+	// Fetch destination info if the driver supports it
+	if r.supportsGetReplicationDestinationInfo(ctx, instance.Spec.Provisioner) && groupID != "" {
+		repClient, clientErr := r.getGroupReplicationClient(ctx, instance.Spec.Provisioner)
```
Nikhil-Ladha: We are already fetching the groupClient at the start of the reconcile(), it can be reused here.

internal/controller/replication.storage/volumegroupreplicationcontent_controller.go
```diff
@@ -239,6 +284,38 @@ func (r *VolumeGroupReplicationContentReconciler) SetupWithManager(mgr ctrl.Mana
...
+// replication RPCs at the group level.
+func (r *VolumeGroupReplicationContentReconciler) getGroupReplicationClient(ctx context.Context, driverName string) (grpcClient.VolumeReplication, error) {
+	conn, err := r.Connpool.GetLeaderByDriver(ctx, r.Client, driverName)
+	if err != nil {
+		return nil, fmt.Errorf("no leader for the ControllerService of driver %q", driverName)
+	}
+
+	return grpcClient.NewVolumeGroupReplicationClient(conn.Client, r.Timeout), nil
+}
+
```
Nikhil-Ladha: We already have a getVolumeGroupClient function just below this, we don't need a separate groupReplicationClient.
Moreover, there is no VolumeGroupReplicationClient as such.

internal/controller/replication.storage/volumegroupreplicationcontent_controller.go
```diff
@@ -211,17 +211,62 @@ func (r *VolumeGroupReplicationContentReconciler) Reconcile(ctx context.Context,
...
+			resp, destErr := repClient.GetReplicationDestinationInfo(
+				groupID,
+				secretName,
+				secretNamespace,
+			)
+			if destErr != nil {
+				// SP may return UNAVAILABLE if destination details are not yet available.
+				// Requeue to retry -- destination fields remain empty, so VGR keeps
+				// DestinationInfoAvailable=False.
+				logger.Info("failed to get destination info, will retry", "error", destErr)
```
Nikhil-Ladha: Missing return with requeue

internal/controller/replication.storage/volumegroupreplicationcontent_controller.go
```diff
@@ -211,17 +211,62 @@ func (r *VolumeGroupReplicationContentReconciler) Reconcile(ctx context.Context,
...
+			} else if resp != nil {
```
Nikhil-Ladha: If a driver supports this capability and there is no error, then it should never return `nil` as response, right?

internal/controller/replication.storage/volumereplication_controller.go
```diff
@@ -560,6 +595,27 @@ func (r *VolumeReplicationReconciler) getReplicationClient(ctx context.Context,
...
+		}
+
+		if cap.GetVolumeReplication().GetType() == identity.Capability_VolumeReplication_GET_REPLICATION_DESTINATION_INFO {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
```
Nikhil-Ladha: This is also defined and used in vgrcontent_controller.
Let's move it to `utils.go` file under `internal/controller/replication.storage/utils.go` to avoid duplication.

internal/proto/replication.proto
```diff
@@ -244,3 +248,55 @@ message ReplicationSource {
...
+// get the destination volume or volume group details for an existing
+// replication.
+message GetReplicationDestinationInfoRequest {
+  // The source volume or volume group for which destination
+  // details are requested.
+  // This field is REQUIRED.
+  ReplicationSource replication_source = 1;
+  // Secrets required by the plugin to complete the request.
+  string secret_name = 2;
+  string secret_namespace = 3;
```
Nikhil-Ladha: Let's move the replication_source to the end and secrets in the beginning.

internal/sidecar/service/volumereplication.go
```diff
@@ -266,6 +266,71 @@ func (rs *ReplicationServer) GetVolumeReplicationInfo(
...
+	}
+	err = setReplicationSource(&repReq.ReplicationSource, req.GetReplicationSource())
+	if err != nil {
+		logger.Error(err, "Failed to set replication source")
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	resp, err := rs.controllerClient.GetReplicationDestinationInfo(ctx, repReq)
+	if err != nil {
+		logger.Error(err, "Failed to get volume replication info")
```
Nikhil-Ladha:
```suggestion
		logger.Error(err, "Failed to get replication destination info")
```

internal/sidecar/service/volumereplication.go
```diff
@@ -266,6 +266,71 @@ func (rs *ReplicationServer) GetVolumeReplicationInfo(
...
+		return nil, err
+	}
+
+	dest, err := convertReplicationDestination(resp.GetReplicationDestination())
+	if err != nil {
+		logger.Error(err, "Failed to convert replication destination")
```
Nikhil-Ladha:
```suggestion
		logger.Error(err, "Failed to convert replication destination info")
```

Nikhil-Ladha (commented):
Some initial review.

build/Containerfile.tools
nicknevin: From the commit message. If there is a better way to address this I'm open to suggestions.

> commit f526b993fbaef2352b901dc240186c424af4e01c
> Author: Nick Nevin <nnevin@redhat.com>
> Date:   Mon Apr 27 17:40:03 2026 -0400
> 
>     test: use a version of fedora with a newer version of go
> 
>     fedora:latest has go 1.25 which doesn't meet the project requirement of go 1.27

nicknevin:
@Nikhil-Ladha Thanks for the review. I have done some refactoring and I believe all your comments have been addressed.

build/Containerfile.tools
Nikhil-Ladha: We still don't have 1.27 support available in product builds, we shouldn't be bumping the go version at all.
Please revert to 1.25.z

nicknevin: I misspoke earlier. The project requirement is 1.25.7 (not 1.27) and this PR is not changing the go version. Without this change from fedora:latest to fedora:44 I get an error when running make test. 
```
$ make test
cd ./tools && go build -o /sandbox/idmap/kubernetes-csi-addons/bin/controller-gen ./vendor/sigs.k8s.io/controller-tools/cmd/controller-gen
cd ./tools && go build -o /sandbox/idmap/kubernetes-csi-addons/bin/kustomize ./vendor/sigs.k8s.io/kustomize/kustomize/v5
/sandbox/idmap/kubernetes-csi-addons/bin/controller-gen rbac:roleName=manager-role crd webhook paths="{./api/...,./cmd/...,./internal/controller/...,./internal/...,./sidecar/...}" output:crd:artifacts:config=config/crd/bases
cd config/manager && /sandbox/idmap/kubernetes-csi-addons/bin/kustomize edit set image controller=quay.io/csiaddons/k8s-controller:latest
/sandbox/idmap/kubernetes-csi-addons/bin/kustomize build config/crd > deploy/controller/crds.yaml
/sandbox/idmap/kubernetes-csi-addons/bin/kustomize build config/rbac > deploy/controller/rbac.yaml
/sandbox/idmap/kubernetes-csi-addons/bin/kustomize build config/manager > deploy/controller/setup-controller.yaml
/sandbox/idmap/kubernetes-csi-addons/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths="./api/..."
podman build -f ./build/Containerfile.tools -t quay.io/csiaddons/tools:latest .
STEP 1/5: FROM quay.io/fedora/fedora:latest
STEP 2/5: ENV GOPATH=/go
--> Using cache cd876826b95474fa4dfddce60ee31322bd6b6f8897cde2d4aea9203423a7dc00
--> cd876826b954
STEP 3/5: ENV PATH="${GOPATH}/bin:${PATH}"
--> Using cache 1abc105c2070bdf7bd80b49e58ec803de1b20013bccb0dddc36158aa1dfafd59
--> 1abc105c2070
STEP 4/5: RUN dnf -y install 	git 	make 	golang 	protobuf-compiler     && dnf -y update     && dnf clean all     && rm -rf /var/cache/yum     && true
--> Using cache 25d12f3fbe65801cc2347f76d4a505439de2ffbfb405708d99128b858512aec9
--> 25d12f3fbe65
STEP 5/5: WORKDIR "/go/src/github.com/csi-addons/kubernetes-csi-addons"
--> Using cache 5a4750f2d52218dd2df53f2be3156b77ec84587f35cb33b383e795d22d53a896
COMMIT quay.io/csiaddons/tools:latest
--> 5a4750f2d522
Successfully tagged quay.io/csiaddons/tools:latest
Successfully tagged quay.io/csiaddons/tools:canary
5a4750f2d52218dd2df53f2be3156b77ec84587f35cb33b383e795d22d53a896
podman run --rm -ti --volume=/sandbox/idmap/kubernetes-csi-addons:/go/src/github.com/csi-addons/kubernetes-csi-addons:Z quay.io/csiaddons/tools:latest make generate-protobuf
cd ./tools && go build -o /go/src/github.com/csi-addons/kubernetes-csi-addons/bin/protoc-gen-go ./vendor/google.golang.org/protobuf/cmd/protoc-gen-go
go: go.mod requires go >= 1.25.7 (running go 1.25.5; GOTOOLCHAIN=local)
make: *** [Makefile:253: protoc-gen-go] Error 1
make: *** [Makefile:199: docker-generate-protobuf] Error 2
```

Nikhil-Ladha: I think the latest tag would always be ahead of (or equal to) any versioned tag, and that said probably you might have an older copy of the `fedora:latest` image in your local system. So, please delete the image locally and re-run the `make test` command, it should work.
I just tried locally and it works.

internal/controller/replication.storage/status_test.go
```diff
@@ -0,0 +1,154 @@
+/*
+Copyright 2025 The Kubernetes-CSI-Addons Authors.
```
Madhu-1: 2026

internal/controller/replication.storage/status_test.go
```diff
@@ -0,0 +1,154 @@
...
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"testing"
+
+	"github.com/csi-addons/kubernetes-csi-addons/api/replication.storage/v1alpha1"
+	"github.com/stretchr/testify/assert"
```
Madhu-1, line 22: add new line for import separation

internal/controller/replication.storage/volumegroupreplication_controller.go
```diff
@@ -44,7 +44,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	replicationv1alpha1 "github.com/csi-addons/kubernetes-csi-addons/api/replication.storage/v1alpha1"
+	conn "github.com/csi-addons/kubernetes-csi-addons/internal/connection"
 	"github.com/csi-addons/kubernetes-csi-addons/internal/controller/replication.storage/replication"
+	"github.com/csi-addons/spec/lib/go/identity"
```
Madhu-1: move this to line 45 for import grouping

internal/controller/replication.storage/volumegroupreplicationcontent_controller.go
```diff
@@ -211,17 +211,54 @@ func (r *VolumeGroupReplicationContentReconciler) Reconcile(ctx context.Context,
...
+		if pv.Spec.CSI != nil {
+			handleToPV[pv.Spec.CSI.VolumeHandle] = pv
+		}
+	}
+
+	pvRefList := []corev1.LocalObjectReference{}
+	pvMappings := []replicationv1alpha1.PersistentVolumeMapping{}
+	for _, handle := range instance.Spec.Source.VolumeHandles {
+		if pv, ok := handleToPV[handle]; ok {
 			pvRefList = append(pvRefList, corev1.LocalObjectReference{
```
Madhu-1, line 229: can we add a comment to skip adding to the pvReflist?

internal/controller/replication.storage/volumereplication_controller.go
```diff
@@ -481,6 +494,29 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+		destReplication := replication.Replication{
+			Params: vr.commonRequestParameters,
+		}
+		destResp := destReplication.GetDestinationInfo()
+		if destResp.Error != nil {
+			logger.Info("failed to get replication destination info, will retry", "error", destResp.Error)
+			setDestinationInfoFailedCondition(&instance.Status.Conditions,
+				instance.Generation, instance.Spec.DataSource.Kind, replication.GetMessageFromError(destResp.Error))
+		} else if resp, ok := destResp.Response.(*proto.GetReplicationDestinationInfoResponse); ok && resp != nil {
+			if vol := resp.GetReplicationDestination().GetVolume(); vol != nil {
```
Madhu-1: dont we need a nil check for `GetReplicationDestination()`?

internal/controller/replication.storage/volumereplication_controller.go
```diff
@@ -481,6 +494,29 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+	// This runs on every reconcile where destination info is not yet available,
+	// ensuring retries are decoupled from replication state changes.
+	cond := findCondition(instance.Status.Conditions, v1alpha1.ConditionDestinationInfoAvailable)
+	if cond != nil && cond.Status != metav1.ConditionTrue {
+		destReplication := replication.Replication{
+			Params: vr.commonRequestParameters,
+		}
+		destResp := destReplication.GetDestinationInfo()
+		if destResp.Error != nil {
+			logger.Info("failed to get replication destination info, will retry", "error", destResp.Error)
```
Madhu-1: log as error?

internal/controller/replication.storage/volumegroupreplicationcontent_controller.go
```diff
@@ -211,17 +211,54 @@ func (r *VolumeGroupReplicationContentReconciler) Reconcile(ctx context.Context,
...
+	repClient, destinationInfoSupported, err := getReplicationClient(ctx, r, instance.Spec.Provisioner, volumeGroupReplicationDataSource)
+	if err != nil {
+		logger.Error(err, "failed to get replication client")
+		return reconcile.Result{}, err
+	} else if destinationInfoSupported {
+		resp, err := repClient.GetReplicationDestinationInfo(groupID, secretName, secretNamespace)
+		if err != nil {
+			// SP may return UNAVAILABLE if destination details are not yet available.
+			// Requeue to retry -- destination fields remain empty, so VGR keeps DestinationInfoAvailable=False.
+			logger.Info("failed to get replication destination info", "error", err)
```
Madhu-1: log as error

internal/controller/replication.storage/volumereplication_controller.go
```diff
@@ -481,6 +494,29 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+			setDestinationInfoFailedCondition(&instance.Status.Conditions,
+				instance.Generation, instance.Spec.DataSource.Kind, replication.GetMessageFromError(destResp.Error))
+		} else if resp, ok := destResp.Response.(*proto.GetReplicationDestinationInfoResponse); ok && resp != nil {
```
Madhu-1: lets log if we hit `!ok` case?

internal/controller/replication.storage/volumegroupreplicationcontent_controller.go
```diff
@@ -211,17 +211,54 @@ func (r *VolumeGroupReplicationContentReconciler) Reconcile(ctx context.Context,
...
+			return reconcile.Result{}, err
+		} else if vgDest := resp.GetReplicationDestination().GetVolumegroup(); vgDest != nil {
+			instance.Status.DestinationVolumeGroupID = vgDest.GetVolumeGroupId()
+			// Enrich PersistentVolumeMappingList with destination handles
+			destVolumeIDs := vgDest.GetVolumeIds()
```
Madhu-1: This can also be nil, we need to handle add nil case as well

internal/controller/replication.storage/volumegroupreplicationcontent_controller.go
```diff
@@ -211,17 +211,54 @@ func (r *VolumeGroupReplicationContentReconciler) Reconcile(ctx context.Context,
...
+		logger.Error(err, "failed to get replication client")
+		return reconcile.Result{}, err
+	} else if destinationInfoSupported {
+		resp, err := repClient.GetReplicationDestinationInfo(groupID, secretName, secretNamespace)
+		if err != nil {
+			// SP may return UNAVAILABLE if destination details are not yet available.
+			// Requeue to retry -- destination fields remain empty, so VGR keeps DestinationInfoAvailable=False.
+			logger.Info("failed to get replication destination info", "error", err)
+			return reconcile.Result{}, err
+		} else if vgDest := resp.GetReplicationDestination().GetVolumegroup(); vgDest != nil {
```
Madhu-1: `GetReplicationDestination()` can also be nil, we need to check for nil before accessing the `GetVolumeGroup()`

internal/controller/replication.storage/volumereplication_controller.go
```diff
@@ -481,6 +494,29 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
...
+	cond := findCondition(instance.Status.Conditions, v1alpha1.ConditionDestinationInfoAvailable)
+	if cond != nil && cond.Status != metav1.ConditionTrue {
+		destReplication := replication.Replication{
+			Params: vr.commonRequestParameters,
+		}
+		destResp := destReplication.GetDestinationInfo()
+		if destResp.Error != nil {
+			logger.Info("failed to get replication destination info, will retry", "error", destResp.Error)
+			setDestinationInfoFailedCondition(&instance.Status.Conditions,
+				instance.Generation, instance.Spec.DataSource.Kind, replication.GetMessageFromError(destResp.Error))
```
Madhu-1: dont we need to requeu or return `ctrl.Result{}, err` in case of failure here?

internal/controller/replication.storage/volumereplication_controller.go
```diff
@@ -306,6 +306,19 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		return ctrl.Result{}, nil
 	}
 
+	if destinationInfoSupported {
```
Madhu-1: can we move the `instance.Status.LastStartTime = getCurrentTime()` before this if check so that we will not have stale LastStartTime if the Update fail below

internal/controller/replication.storage/status_test.go
nicknevin: Ok I read up a bit about formatting of imports in go. What convention does this project follow? Do you make use of one of the tools like goimport or goimport-reviser, and if so what options do you pass to it?

internal/controller/replication.storage/volumegroupreplicationcontent_controller.go
nicknevin: Sorry I don't understand what you are asking for here. Can you elaborate further please?

internal/controller/replication.storage/volumereplication_controller.go
nicknevin: Sure, can do. BTW after setting instance.Status.LastStartTime the current (original) code calls r.Update(). Shouldn't this rather be a call to r.Status().Update()?

build/Containerfile.tools
nicknevin: Doh! That worked thanks. I'll revert the change.

internal/controller/replication.storage/status_test.go
Madhu-1: We follow below

```
import (
	"standard"

	"third-party"

	"local"
)
```

internal/controller/replication.storage/volumegroupreplicationcontent_controller.go
Madhu-1: We have deprecated the `pvRefList` in the API as we have new field, a comment to mention that the field is deprecreated and we can stop adding pvRefList in future

internal/controller/replication.storage/volumereplication_controller.go
Madhu-1: Yes i think so, i would leave that for now and take it as a bug fix later on to have minimal changes as part of this PR

internal/controller/replication.storage/volumereplication_controller.go
nicknevin: I have added nil checks in the places you suggested. I take it this is basically protection against a bad driver returning a response with missing fields. A question though, when a nil check fails should we requeue with backoff? I would think so, to guarantee there is a retry to get the destination info as I don't see any guarantee that one of the watches will fire and lead to a retry.

nicknevin:
@Madhu-1. In the volumereplication controller we are currently getting and updating destination info in both the volume and volume group cases. It occurs to me that it isn't necessary and can be omitted for the volume group case because there it is handled in the volumegroupreplicationcontent controller. Do I have this right?

internal/controller/replication.storage/volumereplication_controller.go
Madhu-1: retry sounds good if we return error it automatically get requeued with exponential backoff

Madhu-1:
> @Madhu-1. In the volumereplication controller we are currently getting and updating destination info in both the volume and volume group cases. It occurs to me that it isn't necessary and can be omitted for the volume group case because there it is handled in the volumegroupreplicationcontent controller. Do I have this right?

@nicknevin yes you are right, we should skip GetDestinationInfo in replication controller if the VR is created for a group

nicknevin:
With the latest commit I believe all review comments to date have been addressed. I'll do a final rebase and squash to clean up the history before any merge.

internal/sidecar/service/volumereplication.go
```diff
@@ -266,6 +266,71 @@ func (rs *ReplicationServer) GetVolumeReplicationInfo(
 	}, nil
 }
 
+// GetVolumeReplicationInfo fetches required information from kubernetes cluster and calls
+// CSI-Addons GetVolumeReplicationInfo service.
```
Rakshith-R:
```suggestion
// GetReplicationDestinationInfo fetches required information from kubernetes cluster and calls
// CSI-Addons GetReplicationDestinationInfo service.
```

internal/controller/replication.storage/volumereplication_controller.go
```diff
@@ -481,6 +493,13 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		updatedConditions := removeCondition(&vr.instance.Status.Conditions, replicationv1alpha1.ConditionReplicating)
 		vr.instance.Status.Conditions = updatedConditions
 	}
+
+	if instance.Spec.DataSource.Kind == pvcDataSource {
```
Rakshith-R:
```suggestion
	if destinationInfoSupported && instance.Spec.DataSource.Kind == pvcDataSource {
```

internal/controller/replication.storage/volumegroupreplication_controller.go
```diff
@@ -816,3 +836,47 @@ func (r *VolumeGroupReplicationReconciler) cleanupVR(vgr *replicationv1alpha1.Vo
 
 	return err
 }
+
+func (r *VolumeGroupReplicationReconciler) updateReplicationDestinationCondition(vgr *replicationv1alpha1.VolumeGroupReplication,
+	persistentVolumeMappingList []replicationv1alpha1.PersistentVolumeMapping) {
+	if persistentVolumeMappingList != nil {
```
Rakshith-R:
```suggestion
	if len(persistentVolumeMappingList) > 0 {
```
Should this check also consider empty list ?

internal/controller/replication.storage/status_test.go
```diff
@@ -0,0 +1,155 @@
...
+	assert.False(t, isDestinationInfoAvailable(conditions))
+	assert.Len(t, conditions, 1) // should update in place, not add
+
+	// Transition to available
+	setDestinationInfoAvailableCondition(&conditions, 2, pvcDataSource)
+	assert.True(t, isDestinationInfoAvailable(conditions))
+	assert.Len(t, conditions, 1) // still one condition, updated in place
+
+	// Transition back to pending (e.g., membership change)
+	setDestinationInfoPendingCondition(&conditions, 3, pvcDataSource)
```
Rakshith-R, line 152: Should this be used in cases of errors too ? 
or re-sync ?

Rakshith-R (commented):
some nits

internal/proto/replication.proto
Nikhil-Ladha: This is yet to be updated.

internal/controller/replication.storage/volumegroupreplication_controller.go
Nikhil-Ladha: Yes, we should allow empty list as we allow creation of an empty group.

internal/controller/replication.storage/status_test.go
Nikhil-Ladha: For errors, the pending condition will already be present as we add it in the top of the reconcile loop when the capability is present.
For re-sync, I don't think we need it imo.

internal/controller/replication.storage/volumegroupreplication_controller.go
nicknevin: So should persistentVolumeMappingList == nil be treated the same as persistentVolumeMappingList == [ ] ?
I.e. that all volume mappings exist is vacuously true because the group is empty.

Nikhil-Ladha: Not exactly, as the pvVoumeMappingList would be nil in the first reconcile, but it won't be nil from next time if the group is not empty as we are defining an empty list in VGRContent reconciler. It can be empty though, till the grouping is complete and that transmits to `PendingCondition`.
We should be passing the `pvcRefList` to this function and check for it's `len()` and then set the condition, something like this:

```
if len(pvcRefList) > 0 {
  if len(pvMapping) > 0 {
	check if mappings available
	if yes, set available condition and return
  }
  set Pending condition
  return
} else {
  group is empty, we should set it to available condition
}
```

internal/controller/replication.storage/status_test.go
Rakshith-R: :thinking: 
@Madhu-1 do you think the current implementation will handle all the corner cases ?
of resync / change in underlying image id etc ?

Madhu-1: Please note that we issue resync only during secondary state not on primary state, this RPC is mostly for the primate state not for secondary, do you still think we need to address it? let me know if any case is missed. we need to be very sure about this one

Rakshith-R: > Please note that we issue resync only during secondary state not on primary state, this RPC is mostly for the primate state not for secondary

On resync, volume handle at secondary changes ? 
This change will be propogated back to primary cluster's RPC call and then it'll used to recreate PVCs in secondary cluster ? 

Or did I miss something crucial ? :thinking:

Madhu-1: The volume Handle will remain same only the internal image ID changes.

internal/controller/replication.storage/volumegroupreplication_controller.go
nicknevin: Got it. Fixed in the latest commit.

Rakshith-R (dismissed):
LGTM, Thanks !

internal/proto/replication.proto
Nikhil-Ladha: @nicknevin can you please fix this?

Nikhil-Ladha (changes requested):
Just the order in the proto file needs to be fixed, rest looks good.

internal/proto/replication.proto
nicknevin: Done.

Nikhil-Ladha (approved):
lgtm, thanks

Rakshith-R (approved):
@mergifyio queue

PR #1041, opened 2026-05-25
fail early when PVC is part of both VR, VGR

Nikhil-Ladha:
Currently, if a PVC is part of VR and we try to create a VGR which includes that PVC, then it will try to group the volumes as the annotation check will pass for the first time and fail in subsequent reconciles. The storage level validation was the only safeguard.
We should fail early if we see a PVC is part of a VR/VGR already through owner annotations.

### DIFF
--- a/internal/controller/replication.storage/pvc.go
+++ b/internal/controller/replication.storage/pvc.go
@@ -73,8 +73,8 @@ func annotatePVCWithOwner(client client.Client, ctx context.Context, logger logr
 		pvc.Annotations = map[string]string{}
 	}
 
-	if pvc.Annotations[replicationv1alpha1.VolumeReplicationNameAnnotation] != "" &&
-		pvc.Annotations[replicationv1alpha1.VolumeGroupReplicationNameAnnotation] != "" {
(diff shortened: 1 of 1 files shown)


internal/controller/replication.storage/pvc.go
```diff
@@ -73,8 +73,8 @@ func annotatePVCWithOwner(client client.Client, ctx context.Context, logger logr
 		pvc.Annotations = map[string]string{}
 	}
 
-	if pvc.Annotations[replicationv1alpha1.VolumeReplicationNameAnnotation] != "" &&
-		pvc.Annotations[replicationv1alpha1.VolumeGroupReplicationNameAnnotation] != "" {
+	if (pvc.Annotations[replicationv1alpha1.VolumeReplicationNameAnnotation] != "" && pvcAnnotation == replicationv1alpha1.VolumeGroupReplicationNameAnnotation) ||
```
Madhu-1, line 76: @Nikhil-Ladha why do we need to have check for both key and vaules? isnt checking key was enough?

Nikhil-Ladha: We are checking the key only.
Those are 2 different keys one is for VR, other is VGR, i.e we are checking if the VR key is already present in the PVC and if the new annotation being added is VGR and vice versa.

Madhu-1 (approved):

black-dragon74 (approved):

PR #1048, opened 2026-06-18
feat: (sidecar) annotate PVCs with volume health status

SanjalKatiyar:
Add a PVC annotation recorder that writes per-node keys (`csiaddons.openshift.io/volumehealth.<hostUID>`).
Records both `healthy` and `unhealthy` states, updates `lastChecked` on each tick/run, adds `since` while state is unchanged and adds `node` which is the node of which PV is attached.

Ref: https://github.com/csi-addons/kubernetes-csi-addons/pull/1048#discussion_r3443180800 for more details.

Other related changes:
`flags: (sidecar) enable volume condition reporting by default (GA)` https://github.com/csi-addons/kubernetes-csi-addons/pull/1050
`feat: (controller) add periodic PVC stale volume health cleanup` https://github.com/csi-addons/kubernetes-csi-addons/pull/1055

## 1. Normal Testing

> CephFS PVCs: 3
> CephRBD PVCs: 62
> CephRBD Virt PVCs: 3

##### Annotated RWX PVC:
```
metadata:
  name: cephfs-rwx-health
  namespace: volume-health-test
  uid: 73a964ac-e133-4c8b-bc43-b999ee930f4c
  resourceVersion: '579908'
  creationTimestamp: '2026-07-13T10:19:26Z'
  annotations:
    csiaddons.openshift.io/volumehealth.6aed389b-2965-463d-ba31-960fa627bba8: '{"state":"healthy","lastChecked":"2026-07-13T10:20:47Z","since":"2026-07-13T10:19:48Z","node":"ip-10-0-42-8.ec2.internal"}'
    csiaddons.openshift.io/volumehealth.7c4f36ac-b7f2-4eda-9fb4-a5ddb4126a86: '{"state":"healthy","lastChecked":"2026-07-13T10:21:15Z","since":"2026-07-13T10:20:15Z","node":"ip-10-0-89-115.ec2.internal"}'
```

##### Annotated RWO PVC:
```
metadata:
  name: cephfs-rwo-health
  namespace: volume-health-test
  uid: 4e836f1a-5bca-491b-800a-6024f0dd698c
  resourceVersion: '581661'
  creationTimestamp: '2026-07-13T10:19:24Z'
  annotations:
    csiaddons.openshift.io/volumehealth.6aed389b-2965-463d-ba31-960fa627bba8: '{"state":"healthy","lastChecked":"2026-07-13T10:22:47Z","since":"2026-07-13T10:19:47Z","node":"ip-10-0-42-8.ec2.internal"}'
```

##### openshift-storage.cephfs.csi.ceph.com-nodeplugin-csi-addonztvzj Logs:
```
2026-07-13T08:49:47.624Z INFO client Probing CSI driver for readiness
2026-07-13T08:49:47.630Z INFO setup The CSI-plugin does not have the CSI-Addons CONTROLLER_SERVICE capability, not running leader election
2026-07-13T08:49:47.816Z INFO setup Starting volume condition reporter {"driver": "openshift-storage.cephfs.csi.ceph.com"}
2026-07-13T08:49:47.936Z INFO csiaddonsnode Starting watcher for CSIAddonsNode {"name": "ip-10-0-42-8.ec2.internal-openshift-storage-daemonset-openshift-storage.cephfs.csi.ceph.com-nodeplugin-csi-addons"}
2026-07-13T08:49:48.619Z INFO server Listening for CSI-Addons requests {"address": "[::]:9071"}
2026-07-13T10:19:47.844Z INFO volume-condition Persistent volume is healthy {"pvName": "pvc-4e836f1a-5bca-491b-800a-6024f0dd698c", "message": "volume is in a healthy condition"}
2026-07-13T10:19:47.874Z INFO volume-condition Persistent volume is healthy {"pvName": "pvc-2af09987-c33a-44aa-8673-a34b35c94add", "message": "volume is in a healthy condition"}
2026-07-13T10:19:48.022Z INFO volume-condition Persistent volume is healthy {"pvName": "pvc-73a964ac-e133-4c8b-bc43-b999ee930f4c", "message": "volume is in a healthy condition"}
2026-07-13T10:22:02.525Z INFO csiaddonsnode Watcher exited gracefully, will be restarted soon {"name": "ip-10-0-42-8.ec2.internal-openshift-storage-daemonset-openshift-storage.cephfs.csi.ceph.com-nodeplugin-csi-addons"}
2026-07-13T10:22:02.552Z INFO csiaddonsnode Starting watcher for CSIAddonsNode {"name": "ip-10-0-42-8.ec2.internal-openshift-storage-daemonset-openshift-storage.cephfs.csi.ceph.com-nodeplugin-csi-addons"}
```

##### csi-addons-controller-manager-bc56675d-f8w4f Logs:
```
2026-07-13T10:48:18.842Z INFO Starting EventSource {"controller": "encryptionkeyrotationcronjob", "controllerGroup": "csiaddons.openshift.io", "controllerKind": "EncryptionKeyRotationCronJob", "source": "kind source: *v1alpha1.EncryptionKeyRotationCronJob"}
2026-07-13T10:48:18.843Z INFO volume-health-cleanup-worker Completed volume health cleanup {"pvcWithVolumeHealth": 65, "pvcCleaned": 0, "malformedSkipped": 0, "errors": 0}
2026-07-13T10:48:18.930Z INFO Starting Controller {"controller": "csiaddonsnode", "controllerGroup": "csiaddons.openshift.io", "controllerKind": "CSIAddonsNode"}
```

##### Events:
Screenshot 2026-07-13 at 3 54 23 PM
Screenshot 2026-07-13 at 3 54 47 PM

## 2. Load Testing

> Total PVCs (with corresponding Pods) ~ 1000-1100
> Bound PVCs (attached to running Pods) ~ 700-800
> Bound PVCs (attached to pending Pods) ~ 200-300
> CephFS PVCs ~ 50%
> CephRBD PVCs ~ 50%
> Sidecar PVCs condition interval - default 1m

##### Cluster Dashboards:
Screenshot 2026-07-13 at 5 13 26 PM
Screenshot 2026-07-13 at 5 13 33 PM
Screenshot 2026-07-13 at 5 14 19 PM

##### Generated Report:
```
================================================================
  Volume Health Feature — Load Test Report
  Generated: 2026-07-13T12:14:15Z
================================================================

CLUSTER OVERVIEW
────────────────────────────────────────────────────────────────

  Nodes:
    ip-10-0-11-132.ec2.internal              master/infra   Ready    v1.35.5 (max 250 pods)
    ip-10-0-152-236.ec2.internal             worker         v1.35.5  250 (max  pods)
    ip-10-0-169-134.ec2.internal             worker         v1.35.5  250 (max  pods)
    ip-10-0-24-142.ec2.internal              worker         v1.35.5  250 (max  pods)
    ip-10-0-42-8.ec2.internal                worker         v1.35.5  250 (max  pods)
    ip-10-0-47-81.ec2.internal               master/infra   Ready    v1.35.5 (max 250 pods)
    ip-10-0-89-115.ec2.internal              worker         v1.35.5  250 (max  pods)
    ip-10-0-90-58.ec2.internal               master/infra   Ready    v1.35.5 (max 250 pods)

  Pod count per node:
    ip-10-0-90-58.ec2.internal               70 pods
    ip-10-0-89-115.ec2.internal              262 pods
    ip-10-0-47-81.ec2.internal               71 pods
    ip-10-0-42-8.ec2.internal                248 pods
    ip-10-0-24-142.ec2.internal              250 pods
    ip-10-0-169-134.ec2.internal             15 pods
    ip-10-0-152-236.ec2.internal             249 pods
    ip-10-0-11-132.ec2.internal              58 pods
    9h                                       4 pods
    10h                                      27 pods
    (unscheduled/pending)                        279 pods


API SERVER & ETCD METRICS (from Prometheus)
NOTE: VolumeCondition interval used by the Sidecar - (default) 1m
────────────────────────────────────────────────────────────────
[0;36m[INFO][0m Capturing API server metrics...

=== API Server Metrics ===
[0;36m[INFO][0m Found route: thanos-querier -> thanos-querier-openshift-monitoring.apps.dr1-jul-13-26.odf-devcluster.com
[0;36m[INFO][0m Querying Prometheus at thanos-querier-openshift-monitoring.apps.dr1-jul-13-26.odf-devcluster.com...

╔══════════════════════════════════════════════════════════════╗
║  SECTION A: Sidecar → API Server Write Load                ║
║  The sidecar PATCHes PVC annotations with health status.   ║
║  These metrics show how many writes/sec and how fast       ║
║  the API server processes them.                            ║
╚══════════════════════════════════════════════════════════════╝

  [A1] PVC PATCH rate — annotation writes from sidecar (req/s, 5m avg)
  PromQL: sum(rate(apiserver_request_total{verb="PATCH",resource="persistentvolumeclaims"}[5m])) by (instance)
    10.0.11.132:6443: 3.2370 req/s
    10.0.47.81:6443: 0.8667 req/s
    10.0.90.58:6443: 4.3613 req/s
    ── total: 8.4650 req/s

  [A2] PVC PATCH p99 latency — worst-case annotation write time
  PromQL: histogram_quantile(0.99,sum(rate(apiserver_request_duration_seconds_bucket{verb="PATCH",resource="persistentvolumeclaims"}[5m])) by (instance,le))
    10.0.11.132:6443: 24.8 ms
    10.0.47.81:6443: 24.8 ms
    10.0.90.58:6443: 24.8 ms

╔══════════════════════════════════════════════════════════════╗
║  SECTION B: Sidecar → API Server Read Load                 ║
║  Before patching, the sidecar GETs the PVC to read its     ║
║  current annotations. This shows read amplification.       ║
╚══════════════════════════════════════════════════════════════╝

  [B1] PVC GET rate — sidecar reads PVC before patching (req/s, 5m avg)
  PromQL: sum(rate(apiserver_request_total{verb="GET",resource="persistentvolumeclaims"}[5m])) by (instance)
    10.0.11.132:6443: 5.4741 req/s
    10.0.47.81:6443: 3.2667 req/s
    10.0.90.58:6443: 10.4089 req/s
    ── total: 19.1497 req/s

╔══════════════════════════════════════════════════════════════╗
║  SECTION C: PVC Impact — Share of Total API Server Load    ║
║  Compares PVC request rates against the total API server   ║
║  request rate to show what fraction of cluster load comes  ║
║  from the volume health feature.                           ║
╚══════════════════════════════════════════════════════════════╝

  [C1] All PVC operations vs total API server requests (req/s, 5m avg)
  PromQL: sum(rate(apiserver_request_total{resource="persistentvolumeclaims"}[5m])) by (verb)
    PVC GET           18.9678 req/s
    PVC PATCH          8.5485 req/s
    PVC LIST           0.0586 req/s
    PVC WATCH          0.0472 req/s
    PVC DELETE         0.0000 req/s
    PVC PUT            0.0000 req/s
    PVC POST           0.0000 req/s
                   ----------
    PVC total         27.6221 req/s

  [C2] Total API server request rate (all resources, all verbs)
  PromQL: sum(rate(apiserver_request_total[5m]))
    All resources:     188.7716 req/s

  [C3] PVC requests as % of total API server load
  PromQL: sum(rate(apiserver_request_total{resource="persistentvolumeclaims"}[5m])) / sum(rate(apiserver_request_total[5m])) * 100
    PVC share:       14.63%
    Verdict:         MODERATE — PVC operations are a noticeable portion

╔══════════════════════════════════════════════════════════════╗
║  SECTION D: API Server Saturation                          ║
║  How many requests are in-flight right now? High numbers   ║
║  (close to max) indicate the API server is under stress.   ║
╚══════════════════════════════════════════════════════════════╝

  [D1] Inflight requests — summed across all API server instances
  PromQL: sum(apiserver_current_inflight_requests) by (request_kind)
    mutating: 9 inflight
    readOnly: 13 inflight

╔══════════════════════════════════════════════════════════════╗
║  SECTION E: WATCH Fan-out                                  ║
║  Every PVC annotation change triggers WATCH events to all  ║
║  controllers watching PVCs. High rates indicate fan-out    ║
║  amplification from volume health writes.                  ║
╚══════════════════════════════════════════════════════════════╝

  [E1] PVC WATCH rate — watch events triggered by annotation changes (req/s, 5m avg)
  PromQL: sum(rate(apiserver_request_total{verb="WATCH",resource="persistentvolumeclaims"}[5m])) by (instance)
    10.0.11.132:6443: 0.0333 req/s
    10.0.47.81:6443: 0.0111 req/s
    10.0.90.58:6443: 0.0283 req/s
    ── total: 0.0728 req/s
```

### DIFF
--- a/config/rbac/csiaddons_pvc_editor_role.yaml
+++ b/config/rbac/csiaddons_pvc_editor_role.yaml
@@ -1,0 +1,14 @@
+---
+# permissions for csi-addons sidecar to patch PVC annotations.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: csiaddons-pvc-editor-role
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - persistentvolumeclaims
+    verbs:
+      - get
+      - patch
--- a/deploy/controller/rbac.yaml
+++ b/deploy/controller/rbac.yaml
@@ -119,6 +119,19 @@ rules:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
+  name: csi-addons-csiaddons-pvc-editor-role
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - persistentvolumeclaims
+  verbs:
+  - get
+  - patch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
   name: csi-addons-manager-role
 rules:
 - apiGroups:
--- a/docs/volume-condition.md
+++ b/docs/volume-condition.md
@@ -13,8 +13,29 @@ Condition Reporter.
 
 ## Abnormal Volume Condition reporting
 
-Once enabled, the healthy and abnormal volume condition is reported in the logs
-of the CSI-Addons sidecar, and as an Event for the PersistentVolumeClaim.
+Once enabled, the sidecar reports the healthy and abnormal volume conditions as follows:
+
+- logs in the CSI-Addons sidecar
+- Event for the PersistentVolumeClaim
+- a health annotation on the PersistentVolumeClaim
+
+The health annotation on the PersistentVolumeClaim is written per node:
+
+```yaml
+csiaddons.openshift.io/volumehealth.<node-uid>: '{"state":"healthy|unhealthy","lastChecked":"<RFC3339>","since":"<RFC3339>","node":"<node-name>"}'
+```
+
+The sidecar always writes both states explicitly (`healthy` and `unhealthy`) and
+updates `lastChecked` on every tick.
+
+- `state` is always `healthy` or `unhealthy`
+- `lastChecked` is refreshed on every tick
+- `since` is set when the sidecar first observes the current state, and is
+  kept unchanged while that state does not change
+
+Each sidecar instance only updates its own
+`csiaddons.openshift.io/volumehealth.<node-uid>` key and never modifies keys
+written by sidecars running on other nodes.
 
 Users will see the Event in their Namespace, and also when they describe (with
 `kubectl describe ...`) the PersistentVolumeClaim.
@@ -33,7 +54,6 @@ Additional options for reporting include:
 - annotate one or more of
 
   1. the PersistentVolume
-  1. the PersistentVolumeClaim
   1. the Pod that uses the PersistentVolumeClaim
   1. the Node where the volume condition is abnormal
      > unlikely acceptable, needs permissions to the Node object
@@ -88,10 +108,23 @@ rules:
   - apiGroups:
       - ""
     resources:
-      - persistentvolumes
       - persistentvolumeclaims
     verbs:
       - get
+---
+# permissions for csi-addons sidecar to patch PVC health annotation.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: csiaddons-pvc-editor-role
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - persistentvolumeclaims
+    verbs:
+      - get
+      - patch
 ```
 
 [nodegetvolumestats]: https://github.com/container-storage-interface/spec/blob/master/spec.md#nodegetvolumestats
--- a/sidecar/internal/volume-condition/pvc_annotation_recorder.go
+++ b/sidecar/internal/volume-condition/pvc_annotation_recorder.go
@@ -1,0 +1,188 @@
+/*
+Copyright 2026 The Kubernetes-CSI-Addons Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package condition
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/csi-addons/kubernetes-csi-addons/sidecar/internal/volume-condition/volume"
+)
+
+const (
+	pvcVolumeHealthAnnotationKeyPrefix = "csiaddons.openshift.io/volumehealth."
+
+	pvcVolumeHealthAnnotationStateHealthy   = "healthy"
+	pvcVolumeHealthAnnotationStateUnhealthy = "unhealthy"
+)
+
+type pvcAnnotationRecorder struct {
+	client   *kubernetes.Clientset
+	hostName string
+	nodeUID  string
+}
+
+type pvcVolumeHealthAnnotation struct {
+	State       string `json:"state"`
+	LastChecked string `json:"lastChecked"`
+	Since       string `json:"since"`
+	Node        string `json:"node"`
+}
+
+// assert on conditionRecorder interface
+var _ conditionRecorder = &pvcAnnotationRecorder{}
+
+// WithPVCAnnotationRecorder can be passed to the VolumeConditionReporter so
+// that it annotates a PersistentVolumeClaim with the local node volume-health
+// state and timestamps.
+func WithPVCAnnotationRecorder() RecorderOption {
+	return RecorderOption{
+		newRecorder: func(client *kubernetes.Clientset, hostname, nodeUID string) (conditionRecorder, error) {
+			return &pvcAnnotationRecorder{
+				client:   client,
+				hostName: hostname,
+				nodeUID:  nodeUID,
+			}, nil
+		},
+	}
+}
+
+func (par *pvcAnnotationRecorder) record(
+	ctx context.Context,
+	pv *corev1.PersistentVolume,
+	pvc *corev1.PersistentVolumeClaim,
+	vc volume.VolumeCondition,
+) error {
+	if pvc.GetDeletionTimestamp() != nil {
+		return nil
+	}
+
+	patch, err := buildPVCVolumeHealthAnnotationPatch(
+		pvc.GetAnnotations(),
+		par.nodeUID,
+		par.hostName,
+		vc.IsHealthy(),
+	)
+	if err != nil {
+		return fmt.Errorf("failed to build volume health annotation patch for persistent volume claim %q in namespace %q: %w",
+			pvc.Name, pvc.Namespace, err)
+	}
+
+	_, err = par.client.CoreV1().PersistentVolumeClaims(pvc.Namespace).Patch(
+		ctx,
+		pvc.Name,
+		types.MergePatchType,
+		patch,
+		metav1.PatchOptions{},
+	)
+	if err != nil {
+		return fmt.Errorf(
+			"failed to patch volume health annotation for persistent volume claim %q in namespace %q: %w",
+			pvc.Name,
+			pvc.Namespace,
+			err,
+		)
+	}
+
+	return nil
+}
+
+func (par *pvcAnnotationRecorder) needsPVC() bool {
+	return true
+}
+
+func (par *pvcAnnotationRecorder) recordUnchangedVolumes() bool {
+	// "true" so the sidecar keeps refreshing the PVC health annotation
+	// every reporting tick, even when health stays the same.
+	// Otherwise, "lastChecked" timestamp would become stale.
+	return true
+}
+
+func buildPVCVolumeHealthAnnotationPatch(
+	annotations map[string]string,
+	nodeUID string,
+	nodeName string,
+	healthy bool,
+) ([]byte, error) {
+	if nodeUID == "" {
+		return nil, fmt.Errorf("node UID cannot be empty")
+	}
+
+	state := pvcVolumeHealthAnnotationStateHealthy
+	if !healthy {
+		state = pvcVolumeHealthAnnotationStateUnhealthy
+	}
+
+	annotationKey := getPVCVolumeHealthAnnotationKey(nodeUID)
+	lastCheckedTimestamp := time.Now().Format(time.RFC3339)
+	since := getSinceTimestamp(annotations[annotationKey], state)
+	if since == "" {
+		since = lastCheckedTimestamp
+	}
+
+	annotationValue, err := json.Marshal(pvcVolumeHealthAnnotation{
+		State:       state,
+		LastChecked: lastCheckedTimestamp,
+		Since:       since,
+		Node:        nodeName,
+	})
+	if err != nil {
+		return nil, err
+	}
+	patch := map[string]interface{}{
+		"metadata": map[string]interface{}{
+			"annotations": map[string]interface{}{
+				annotationKey: string(annotationValue),
+			},
+		},
+	}
+
+	data, err := json.Marshal(patch)
+	if err != nil {
+		return nil, err
+	}
+
+	return data, nil
+}
+
+func getPVCVolumeHealthAnnotationKey(nodeUID string) string {
+	return pvcVolumeHealthAnnotationKeyPrefix + nodeUID
+}
+
+func getSinceTimestamp(encodedAnnotation string, state string) string {
+	if encodedAnnotation == "" {
+		return ""
+	}
+
+	annotation := pvcVolumeHealthAnnotation{}
+	if err := json.Unmarshal([]byte(encodedAnnotation), &annotation); err != nil {
+		return ""
+	}
+
+	if annotation.State != state {
+		return ""
+	}
+
+	return annotation.Since
+}
--- a/sidecar/internal/volume-condition/reporter.go
+++ b/sidecar/internal/volume-condition/reporter.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"time"
 
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -75,12 +76,13 @@ func NewVolumeConditionReporter(
 	if err != nil {
 		return nil, fmt.Errorf("failed to get node %q: %w", hostname, err)
 	}
+	nodeUID := n.UID()
 
 	// TODO: use options for the recorders
 	recorders := make([]conditionRecorder, len(recorderOptions))
 	for i, opt := range recorderOptions {
 		var rec conditionRecorder
-		rec, err = opt.newRecorder(client, hostname)
+		rec, err = opt.newRecorder(client, hostname, nodeUID)
 		if err != nil {
 			return nil, fmt.Errorf("failed to create recorder: %w", err)
 		}
@@ -100,7 +102,8 @@ func NewVolumeConditionReporter(
 // Run starts the volume condition reporter. This function never returns, only
 // if a critical error occurs.
 // Only new/updated volume conditions are reported, this prevents noise of
-// recurring conditions.
+// recurring conditions. But, this can be overridden
+// by the "recordUnchangedVolumes" flag on a recorder.
 func (cvr *volumeConditionReporter) Run(ctx context.Context, interval time.Duration) error {
 	running := time.Tick(interval)
 	if running == nil {
@@ -128,12 +131,12 @@ func (cvr *volumeConditionReporter) Run(ctx context.Context, interval time.Durat
 				continue
 			}
 
-			if !cvr.isUpdatedVolumeCondition(v.GetVolumeID(), vc) {
-				// skip recording if there is no update
+			updated := cvr.isUpdatedVolumeCondition(v.GetVolumeID(), vc)
+			if !updated && !cvr.hasAnyRecorderForUnchangedVolume() {
 				continue
 			}
 
-			cvr.recordVolumeCondition(ctx, v.GetVolumeID(), vc)
+			cvr.recordVolumeCondition(ctx, v.GetVolumeID(), vc, updated)
 		}
 
 		cvr.pruneConditionCache(volumes)
@@ -171,7 +174,12 @@ func (cvr *volumeConditionReporter) pruneConditionCache(volumes []volume.CSIVolu
 
 // recordVolumeCondition resolves the PersistentVolume from the given volumeID
 // and loops through all recorders to report the volume condition.
-func (cvr *volumeConditionReporter) recordVolumeCondition(ctx context.Context, volumeID string, vc volume.VolumeCondition) {
+func (cvr *volumeConditionReporter) recordVolumeCondition(
+	ctx context.Context,
+	volumeID string,
+	vc volume.VolumeCondition,
+	updated bool,
+) {
 	pvName, err := platform.GetPlatform().ResolvePersistentVolumeName(cvr.driver.GetDrivername(), volumeID)
 	if err != nil {
 		logger.Error(err, "Failed to resolve persistent volume name")
@@ -184,11 +192,61 @@ func (cvr *volumeConditionReporter) recordVolumeCondition(ctx context.Context, v
 		return
 	}
 
+	claim := pv.Spec.ClaimRef
+	var (
+		pvc    *corev1.PersistentVolumeClaim
+		pvcErr error
+	)
+	// fetch PVC details only once, when at least one active recorder needs it,
+	// instead of fetching it for every recorder, preventing unnecessary API calls.
+	if cvr.hasAnyRecorderNeedingPVC(updated) {
+		if claim == nil {
+			pvcErr = fmt.Errorf("persistent volume %q does not have a claim (unused?)", pv.Name)
+			logger.Error(pvcErr, "Failed to get claim from persistent volume", "pvName", pvName)
+		} else {
+			pvc, pvcErr = cvr.client.CoreV1().PersistentVolumeClaims(claim.Namespace).Get(ctx, claim.Name, metav1.GetOptions{})
+			if pvcErr != nil {
+				logger.Error(pvcErr, "Failed to get persistent volume claim", "pvcName", claim.Name, "namespace", claim.Namespace)
+			}
+		}
+	}
+
 	for _, recorder := range cvr.recorders {
-		err = recorder.record(ctx, pv, vc)
+		if !updated && !recorder.recordUnchangedVolumes() {
+			continue
+		}
+
+		if recorder.needsPVC() && pvcErr != nil {
+			continue
+		}
+
+		err = recorder.record(ctx, pv, pvc, vc)
 		if err != nil {
 			logger.Error(err, "Failed to record volume condition for persistent volume",
 				"recorder", fmt.Sprintf("%T", recorder), "pvName", pv.Name)
 		}
 	}
+}
+
+func (cvr *volumeConditionReporter) hasAnyRecorderForUnchangedVolume() bool {
+	for _, recorder := range cvr.recorders {
+		if recorder.recordUnchangedVolumes() {
+			return true
+		}
+	}
+
+	return false
+}
+
+func (cvr *volumeConditionReporter) hasAnyRecorderNeedingPVC(updated bool) bool {
+	for _, recorder := range cvr.recorders {
+		if !updated && !recorder.recordUnchangedVolumes() {
+			continue
+		}
+		if recorder.needsPVC() {
+			return true
+		}
+	}
+
+	return false
 }
--- a/sidecar/main.go
+++ b/sidecar/main.go
@@ -51,10 +51,12 @@ var (
 	// available recorders for the --volume-condition-recorders flag
 	volumeConditionLogRecorder      = "log"
 	volumeConditionPVCEventRecorder = "pvcEvent"
+	volumeConditionPVCAnnotation    = "pvcAnnotation"
 
 	defaultVolumeConditionRecorders = strings.Join([]string{
 		volumeConditionLogRecorder,
 		volumeConditionPVCEventRecorder,
+		volumeConditionPVCAnnotation,
 	}, ",")
 )
 
@@ -206,6 +208,8 @@ func main() {
 					recorderOptions = append(recorderOptions, condition.WithLogRecorder())
 				case volumeConditionPVCEventRecorder:
 					recorderOptions = append(recorderOptions, condition.WithEventRecorder())
+				case volumeConditionPVCAnnotation:
+					recorderOptions = append(recorderOptions, condition.WithPVCAnnotationRecorder())
 				default:
 					setupLog.Error(fmt.Errorf("condition recorder %q is unknown", vcr), "Unknown condition recorder")
 					os.Exit(1)
--- a/config/rbac/kustomization.yaml
+++ b/config/rbac/kustomization.yaml
@@ -21,6 +21,7 @@ resources:
   - leader_election_role.yaml
   - leader_election_role_binding.yaml
   - csiaddons_event_editor_role.yaml
+  - csiaddons_pvc_editor_role.yaml
   # The following RBAC configurations are used to protect
   # the metrics endpoint with authn/authz. These configurations
   # ensure that only authorized users and service accounts
--- a/sidecar/internal/volume-condition/event_recorder.go
+++ b/sidecar/internal/volume-condition/event_recorder.go
@@ -18,10 +18,8 @@ package condition
 
 import (
 	"context"
-	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	typedcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
@@ -47,7 +45,7 @@ func WithEventRecorder() RecorderOption {
 	}
 }
 
-func newEventRecorder(client *kubernetes.Clientset, hostname string) (conditionRecorder, error) {
+func newEventRecorder(client *kubernetes.Clientset, hostname, _ string) (conditionRecorder, error) {
 	eventBroadcaster := record.NewBroadcaster()
 	eventBroadcaster.StartRecordingToSink(
 		&typedcorev1.EventSinkImpl{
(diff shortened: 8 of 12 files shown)


docs/volume-condition.md
```diff
@@ -13,8 +13,19 @@ Condition Reporter.
...
+Once enabled, the sidecar reports the healthy and abnormal volume conditions as follows:
+
+- logs in the CSI-Addons sidecar
+- Event for the PersistentVolumeClaim
+- a health annotation on the PersistentVolumeClaim
+
+The health annotation is:
+
+```yaml
+csi-addons.openshift.io/volumehealth: unhealthy
```
black-dragon74:
```suggestion
csiaddons.openshift.io/volumehealth: unhealthy
```

sidecar/internal/volume-condition/pvc_annotation_recorder.go
```diff
@@ -0,0 +1,156 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
+
+	"github.com/csi-addons/kubernetes-csi-addons/sidecar/internal/volume-condition/volume"
+)
+
+const (
+	pvcVolumeHealthAnnotationKey       = "csi-addons.openshift.io/volumehealth"
```
black-dragon74:
```suggestion
	pvcVolumeHealthAnnotationKey       = "csiaddons.openshift.io/volumehealth"
```

black-dragon74:
What’s the benefit of having it in an annotation when it’s already emitted as an event?

config/rbac/csiaddons_pvc_editor_role.yaml
```diff
@@ -0,0 +1,15 @@
...
+  name: csiaddons-pvc-editor-role
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - persistentvolumeclaims
+    verbs:
+      - get
+      - patch
+      - update
```
SanjalKatiyar: I still need to find out how these roles are actually applied or whether we need to automate the role bindings via another driver specific repository (eg: ceph-csi-operator ??). I didn't find any recorder role bindings in this csi-addons repo.

deploy/controller/rbac.yaml
```diff
@@ -118,6 +118,20 @@ rules:
...
+  name: csi-addons-csiaddons-pvc-editor-role
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - persistentvolumeclaims
+  verbs:
+  - get
+  - patch
+  - update
```
SanjalKatiyar: https://github.com/csi-addons/kubernetes-csi-addons/pull/1048/changes/#r3434931348

SanjalKatiyar:
> What’s the benefit of having it in an annotation when it’s already emitted as an event?

This was already a planned enhancement and incorporated now as part of GA: https://github.com/csi-addons/kubernetes-csi-addons/blob/main/docs/volume-condition.md#future-enhancements.

I can add few reasons:
- PV level info is easier to pin point and can lead to faster resolution.
- Events get GC-ed after TTL, annotation is persistent and more reliable.
- It's much easier to query affected PVC rather than filtering and parsing all Events for the required information.
- Events reports "what happened", where as PV level info reports the "current state".
- We can have dynamic dashboards keeping watch and reporting the affected PVC(s) live.

config/rbac/csiaddons_pvc_editor_role.yaml
SanjalKatiyar: cc @nixpanic @Madhu-1 @Rakshith-R

docs/volume-condition.md
```diff
@@ -13,8 +13,19 @@ Condition Reporter.
...
+Once enabled, the sidecar reports the healthy and abnormal volume conditions as follows:
+
+- logs in the CSI-Addons sidecar
+- Event for the PersistentVolumeClaim
+- a health annotation on the PersistentVolumeClaim
+
+The health annotation is:
+
+```yaml
+csiaddons.openshift.io/volumehealth: unhealthy
```
nixpanic: The problem with this is, that a PV(C) can be attached to multiple nodes (RWX, ROX modes) at the same time. It is not sufficient to report `unhealthy` as the PV(C) may work without issues on an other node. Somehow the annotation should include the node.

What is the reason to remove the annotation when the PV(C) is healthy again? Have you considered to update the annotation to `healthy`?

sidecar/internal/volume-condition/pvc_annotation_recorder.go
```diff
@@ -0,0 +1,156 @@
+/*
+Copyright 2025 The Kubernetes-CSI-Addons Authors.
```
nixpanic: use the current year please

config/rbac/csiaddons_pvc_editor_role.yaml
nixpanic: yes, these roles are part of the CSI-driver deployment, the csi-addons sidecar runs with the CSI-driver (either as a sidecar to their Pods, or as a separate Pod for a selected driver).

sidecar/internal/volume-condition/pvc_annotation_recorder.go
SanjalKatiyar: my bad...

docs/volume-condition.md
SanjalKatiyar: > Somehow the annotation should include the node.

As per my previous understanding, the only plan was for the Event to include the node name (https://github.com/csi-addons/kubernetes-csi-addons/blob/main/sidecar/internal/volume-condition/event_recorder.go#L61).

But, on a second look, if I understand correctly, I see a scenario where the PVC status remains "unhealthy" for an extended period while the corresponding K8s Event gets GC-ed due to the default TTL. Since the status never changes, no new Event would be generated, so no node information as well in that case (https://github.com/csi-addons/kubernetes-csi-addons/blob/main/sidecar/internal/volume-condition/reporter.go#L131-L134). 

---

> What is the reason to remove the annotation when the PV(C) is healthy again? Have you considered to update the annotation to healthy?

Based on our previous discussions, my understanding was that we only need to annotate the "unhealthy" state. The absence of any annotation would implicitly indicate a "healthy" state.
cc @Madhu-1 can you plz confirm ??

---

> The problem with this is, that a PV(C) can be attached to multiple nodes (RWX, ROX modes) at the same time. It is not sufficient to report unhealthy as the PV(C) may work without issues on an other node.

Thanks @nixpanic, I hadn't considered that. In such case, looks like multiple sidecars could write conflicting values (one DaemonSet instance adds or removes the annotation while another instance simultaneously does the opposite), right ??

One suggestion would be (meanwhile, let me look into more alternatives), something for example:
`csiaddons.openshift.io/volumehealth.<node-a>: unhealthy`
 but concern would be how to cleanup annotations when say node get destroyed and recreated (hostName changes) or PV gets detached from a node ??

config/rbac/csiaddons_pvc_editor_role.yaml
SanjalKatiyar: got it, thanks !

docs/volume-condition.md
SanjalKatiyar: > but concern would be how to cleanup annotations when say node get destroyed and recreated (hostName changes) or PV gets detached from a node ??

One option is to have a "lastchecked" heartbeat annotation on the PVC itself:
`csiaddons.openshift.io/volumehealth.lastchecked.<node-a>: <timestamp>`, which will usually get updated every minute (https://github.com/csi-addons/kubernetes-csi-addons/blob/main/sidecar/internal/volume-condition/reporter.go#L105).

And for the cleanup, we add a controller to the addons-controller (https://github.com/csi-addons/kubernetes-csi-addons/tree/main/internal/controller) which would cleanup the stale annotations from the PVC(s) as per the default threshold.

@nixpanic @Madhu-1 can you plz share ur thoughts ?? I might be missing something.

nixpanic: > Thanks @nixpanic, I hadn't considered that. In such case, looks like multiple sidecars could write conflicting values (one DaemonSet instance adds or removes the annotation while another instance simultaneously does the opposite), right ??

That is a concern as well.

The main issue is that one node may have a different condition than an other. And a user would need to identify which node is having troubles.

Adding the node name to the annotation key, would definitely help with that. But you do run in a difficult cleanup path :disappointed: Not sure about a nice way to cleanup stale annotations.

Also, when a PVC is not attached, it should not have an unhealthy state anymore either. I don't know if that can be cleaned up easily somehow.

SanjalKatiyar: > Also, when a PVC is not attached, it should not have an unhealthy state anymore either. I don't know if that can be cleaned up easily somehow.

The annotations won't be removed immediately, but, the cleanup controller can remove them "eventually".

Controller can either be -
1. **running periodically (say every 15 or 30 minutes):**
- in each tick it will list all PVCs (IIUC manager default cache already holds all PVCs as populated by another `persistentvolumeclaim_controller`, so we already follow such caching pattern).
- filter the ones with `volumehealth` annotation.
- checks the "lastchecked" timestamp.
- if timestamp exceeds the "threshold", remove the annotation.
2. **or, can have per pvc reconcile and use `RequeueAfter` for next loop:**
- watch all PVCs (similar to existing `persistentvolumeclaim_controller`). 
- use predicate based on `volumehealth` annotation.
- fetch the current PVC.
- checks the "lastchecked" timestamp.
- if timestamp exceeds the "threshold", remove the annotation.
- if any timestamp DOES NOT exceeds the "threshold", `RequeueAfter` earliest expiry among all node timestamps on that PVC.

But, I am happy to discuss any other suggestion as well ?? or any drawbacks in above proposed suggestion ??

Also, instead of introducing 2 diff annotations per node on a PVC (as I proposed earlier in this thread), we can even have a single/combined per node annotation with the required info needed for the cleanup:
`csiaddons.openshift.io/volumehealth.<node-a>: {"state":"unhealthy","lastchecked":"<timestamp>"}'`.

nixpanic: My 1st inclination would be to approach the cleanup with option 1. That probably is the less resource intense solution. The periodic timeout should be configurable (commandline parameter and/or configmap option).

Storing a JSON blob in the annotation does not have my preference, but it would make things a much simpler. There are other annotations that use JSON blobs already, it is less user friendly, but seems acceptable overall. In addition to `state` and `lastchecked` it would be good to have a `since` timestamp as well. In order to identify a critical issue, the starting time of such a problem is important, and a `since` timestamp of unhealthy PVCs can help with that.

Thanks!

SanjalKatiyar: Thanks @nixpanic for your inputs, I noticed we already have an existing flags/ConfigMap flow in place, so I will add a new option there.
I will implement the cleanup controller (option "1") in a separate PR specifically for the addons-controller changes and update this PR with the related/relevant addons-sidecar changes (as discussed above).

cc @Madhu-1.

SanjalKatiyar: updated... I will do a final e2e testing once implementational logic is approved/agreed upon as expected/okay... will test all 3 changes together before merging:
- sidecar (this PR).
- controller (PR#1055).
- default enable VC flag (PR#1050).

docs/volume-condition.md
```diff
@@ -13,8 +13,29 @@ Condition Reporter.
...
+Once enabled, the sidecar reports the healthy and abnormal volume conditions as follows:
+
+- logs in the CSI-Addons sidecar
+- Event for the PersistentVolumeClaim
+- a health annotation on the PersistentVolumeClaim
+
+The health annotation on the PersistentVolumeClaim is written per node:
+
+```yaml
+csiaddons.openshift.io/volumehealth.<node-name>: '{"state":"healthy|unhealthy","lastChecked":"<RFC3339>","since":"<RFC3339>"}'
```
nixpanic: just wondering, why not rfc9557 formatted timestamps?

sidecar/internal/volume-condition/reporter.go
```diff
@@ -41,12 +41,17 @@ type VolumeConditionReporter interface {
 	Run(ctx context.Context, interval time.Duration) error
 }
 
+type configuredRecorder struct {
```
nixpanic: why add a new type for this, isn't it simpler to just add a function to the `conditionRecorder` interface that returns a bool like `recordUnchangedVolumes` here?

Even though it may result in a little more code, it probably is easier to understand. Options for recording recurring values are kept at the recorder itself.

sidecar/internal/volume-condition/pvc_annotation_recorder.go
```diff
@@ -0,0 +1,186 @@
...
+	if err != nil {
+		return fmt.Errorf(
+			"failed to build volume health annotation patch for persistent volume claim %q in namespace %q: %w",
+			claim.Name,
+			claim.Namespace,
+			err,
+		)
+	}
+
+	_, err = par.client.CoreV1().PersistentVolumeClaims(claim.Namespace).Patch(
```
Madhu-1: wont this silently override problem as we are not using the merge path with resourceVersion

docs/volume-condition.md
SanjalKatiyar: Honestly, I just added `RFC3339` without much thought, as it is widely used thoughout this repo and related vendors files.
But, I will probably make this readMe file bit generic and just add instead (without specifying any format related details):
```
csiaddons.openshift.io/volumehealth.<node-name>: 
'{"state":"healthy|unhealthy","lastChecked":"<timestamp>","since":"<timestamp>"}'
```

nixpanic: Ok. it was merely a question. If RFC3339 is used elsewhere in this repo (or standard kubernetes) use facing timestamps, just stick with that.

sidecar/internal/volume-condition/pvc_annotation_recorder.go
```diff
@@ -0,0 +1,186 @@
...
+		},
+	}
+}
+
+func (par *pvcAnnotationRecorder) record(
+	ctx context.Context,
+	pv *corev1.PersistentVolume,
+	vc volume.VolumeCondition,
+) error {
+	claim := pv.Spec.ClaimRef
```
Madhu-1: what about the pvc that are under deletion?

sidecar/internal/volume-condition/pvc_annotation_recorder.go
```diff
@@ -0,0 +1,186 @@
...
+	data, err := json.Marshal(patch)
+	if err != nil {
+		return nil, err
+	}
+
+	return data, nil
+}
+
+func getPVCVolumeHealthAnnotationKey(hostName string) string {
+	return pvcVolumeHealthAnnotationKeyPrefix + hostName
```
Madhu-1: dont we need to sanitize the hostname to the known format so that the patch never fails?

sidecar/internal/volume-condition/pvc_annotation_recorder.go
```diff
@@ -0,0 +1,186 @@
...
+		newRecorder: func(client *kubernetes.Clientset, hostname string) (conditionRecorder, error) {
+			return &pvcAnnotationRecorder{
+				client:   client,
+				hostName: hostname,
+			}, nil
+		},
+	}
+}
+
+func (par *pvcAnnotationRecorder) record(
```
Madhu-1, line 71: can we add unit test for this as well?

Madhu-1 (commented):
* We need to do load/performance testing with lot of PVC to see how much burden we are adding to API server
* We need to call out this whole design works only when some driver are using StagePath not the publish Path, and we should gate these changes only when stagingPath is used or else we will many other problems (it can be a followup)
* Are we going to have a change to garbage collect the stale entries as well?
* we are doing 2 PVC get call for each record, can we simply to have 1 Get call?

docs/volume-condition.md
SanjalKatiyar: controller changes: https://github.com/csi-addons/kubernetes-csi-addons/pull/1055 (adding for reference).

sidecar/internal/volume-condition/reporter.go
SanjalKatiyar: thanks, updated...

sidecar/internal/volume-condition/pvc_annotation_recorder.go
SanjalKatiyar: That should be okay (no override happens), each annotation key is unique and is only touched upon by the daemon running on that particular node, it does not replaces entire annotations map. So, even if PVC object is changed (resourceVersion is updated), patch should still be fine as it won't affect any other key.

But, plz let me know in case I misunderstood and u had some other concern here.

sidecar/internal/volume-condition/pvc_annotation_recorder.go
SanjalKatiyar: thanks, updated...

sidecar/internal/volume-condition/pvc_annotation_recorder.go
SanjalKatiyar: Thanks for pointing this out !!
Format should not be an issue AFAIU, but looks like name length is definitely a concern.

Node name can be upto [253 chars](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names) and annotation "name" (part after the `/` in the key) can be upto [63 chars](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set) only.

If we go ahead with initial plan of using `csiaddons.openshift.io/volumehealth.<NODE_NAME>` as the key, we are only left with upto 50 chars for the `<NODE_NAME>` part :/

That said, I am not sure if I am over complicating things, but only suggestion I can think of right now is to have annotation as:
1. Either `csiaddons.openshift.io/volumehealth.<NODE_NAME_HASH_WITHIN_50_CHARS>: {"state":"unhealthy","lastchecked":"<timestamp>","node":"node-a"}'`.
That is, calculating hash of the node name and adding that to the key and value will have actual node name in the already existing JSON blob.

2. Or, `csiaddons.openshift.io/volumehealth.<NODE_UID>: {"state":"unhealthy","lastchecked":"<timestamp>","node":"node-a"}'`.
That is, somehow fetching/injecting Node resource UID, which is 36 chars long and adding that to the key and value will have actual node name in the already existing JSON blob.

@Madhu-1 @nixpanic wdyt ??

SanjalKatiyar:
> * We need to do load/performance testing with lot of PVC to see how much burden we are adding to API server

Ack ! I will do a final e2e testing once implementational logic is approved/agreed upon as expected/okay.
Is there any toolkit or any utility we commonly already use for such testing ??

> * We need to call out this whole design works only when some driver are using StagePath not the publish Path, and we should gate these changes only when stagingPath is used or else we will many other problems (it can be a followup)

Let me check more into it, I probably need more details on how it currently works and what's the impact (just for my reference purpose). I will fix it in this same PR.

> * Are we going to have a change to garbage collect the stale entries as well?

Yes, I have https://github.com/csi-addons/kubernetes-csi-addons/pull/1055 for the controller changes.

> * we are doing 2 PVC get call for each record, can we simply to have 1 Get call?

Currently we fetch PVC in both "event" and "annotation" recorder separately, do you mean that instead we fetch it once and pass as argument to these recorders ?? Or, something else ??

cc @Madhu-1.

sidecar/internal/volume-condition/pvc_annotation_recorder.go
SanjalKatiyar: ack !

Madhu-1:
> > * We need to do load/performance testing with lot of PVC to see how much burden we are adding to API server
> 
> Ack ! I will do a final e2e testing once implementational logic is approved/agreed upon as expected/okay. Is there any toolkit or any utility we commonly already use for such testing ??
> 

not as of now, we need to create it using some scripts. 

> > * We need to call out this whole design works only when some driver are using StagePath not the publish Path, and we should gate these changes only when stagingPath is used or else we will many other problems (it can be a followup)
> 
> Let me check more into it, I probably need more details on how it currently works and what's the impact (just for my reference purpose). I will fix it in this same PR.
> 
> > * Are we going to have a change to garbage collect the stale entries as well?
> 
> Yes, I have #1055 for the controller changes.
> 
> > * we are doing 2 PVC get call for each record, can we simply to have 1 Get call?
> 
> Currently we fetch PVC in both "event" and "annotation" recorder separately, do you mean that instead we fetch it once and pass as argument to these recorders ?? Or, something else ??
> 
> cc @Madhu-1.

Yes something like that, we can do as enhancement as well in follow up PR.

sidecar/internal/volume-condition/pvc_annotation_recorder.go
Madhu-1: NodeUUID make sense to me. @nixpanic ^^

sidecar/internal/volume-condition/pvc_annotation_recorder.go
Madhu-1: lets see how it behaves, when we have a PVC mounted on 2 different nodes.

sidecar/internal/volume-condition/pvc_annotation_recorder.go
nixpanic: I agree, Node UID should be good. The documentation and examples will need to explain how to identify the Node from the annotation though.

sidecar/internal/volume-condition/pvc_annotation_recorder.go
SanjalKatiyar: I am un-resolving this comment for now :(

We don't have any unit tests for existing log/event recorders which I can refer, looks like I will have to use mock/fake client(s) to add a test for `record`, that would take bit more efforts/research, so decoupling that changes from this PR.

For now I have added relevant tests for each separate function used in PVC recorder.

SanjalKatiyar:
> * We need to call out this whole design works only when some driver are using StagePath not the publish Path, and we should gate these changes only when stagingPath is used or else we will many other problems (it can be a followup)

@Madhu-1 as you suggested, I will keep this as a followup (if really needed), but I didn't get why it would cause problems ??

For example, NFS (plz check PR: https://github.com/ceph/ceph-csi/pull/6383#pullrequestreview-4643402820), it doesn't support `stagePath`, and addons-sidecar sends `publishPath` to the RPC request instead.
And, driver already have capabilities, so it can always advertise if it can't support volume-conditions (we already check that for recorders: https://github.com/csi-addons/kubernetes-csi-addons/blob/main/sidecar/internal/volume-condition/reporter.go#L70-L72).

Why should we gate this work ??

cc @nixpanic

sidecar/internal/volume-condition/pvc_annotation_recorder.go
SanjalKatiyar: got it, thanks.

Madhu-1:
> > * We need to call out this whole design works only when some driver are using StagePath not the publish Path, and we should gate these changes only when stagingPath is used or else we will many other problems (it can be a followup)
> 
> @Madhu-1 as you suggested, I will keep this as a followup (if really needed), but I didn't get why it would cause problems ??
> 
> For example, NFS (plz check PR: [ceph/ceph-csi#6383 (review)](https://github.com/ceph/ceph-csi/pull/6383#pullrequestreview-4643402820)), it doesn't support `stagePath`, and addons-sidecar sends `publishPath` to the RPC request instead. And, driver already have capabilities, so it can always advertise if it can't support volume-conditions (we already check that for recorders: https://github.com/csi-addons/kubernetes-csi-addons/blob/main/sidecar/internal/volume-condition/reporter.go#L70-L72).
> 
> Why should we gate this work ??
> 
> cc @nixpanic

When you send publish Path you will send RPC per pod not per pvc, so if you have 10 pod using 1 PVC you will send 10 RPC instead of 1 as Publish path is per pod and NodeStage is per PVC on a node.

sidecar/main.go
```diff
@@ -206,6 +208,8 @@ func main() {
 					recorderOptions = append(recorderOptions, condition.WithLogRecorder())
 				case volumeConditionPVCEventRecorder:
 					recorderOptions = append(recorderOptions, condition.WithEventRecorder())
+				case volumeConditionPVCAnnotation:
+					recorderOptions = append(recorderOptions, condition.WithPVCAnnotationRecorder())
```
SanjalKatiyar, line 212: @Madhu-1  @nixpanic I am creating a specific thread for discussing the below concern:

> We need to call out this whole design works only when some driver are using StagePath not the publish Path, and we should gate these changes only when stagingPath is used or else we will many other problems (it can be a followup)

**Discussion so far:**
https://github.com/csi-addons/kubernetes-csi-addons/pull/1048#issuecomment-4903857406

> @Madhu-1 as you suggested, I will keep this as a followup (if really needed), but I didn't get why it would cause problems ??
> For example, NFS (plz check PR: https://github.com/ceph/ceph-csi/pull/6383#pullrequestreview-4643402820), it doesn't support stagePath, and addons-sidecar sends publishPath to the RPC request instead.
> And, driver already have capabilities, so it can always advertise if it can't support volume-conditions (we already check that for recorders: https://github.com/csi-addons/kubernetes-csi-addons/blob/main/sidecar/internal/volume-condition/reporter.go#L70-L72).
> Why should we gate this work ??
> cc @nixpanic

https://github.com/csi-addons/kubernetes-csi-addons/pull/1048#issuecomment-4911302975

> When you send publish Path you will send RPC per pod not per pvc, so if you have 10 pod using 1 PVC you will send 10 RPC instead of 1 as Publish path is per pod and NodeStage is per PVC on a node.

SanjalKatiyar: @Madhu-1 Can you plz elaborate on this ??

Even if there are multiple `publishPath` entries per volume, we select only the first one that matches the `volumeID` and driver (ref: https://github.com/csi-addons/kubernetes-csi-addons/blob/main/sidecar/internal/volume-condition/platform/kubelet.go#L174-L198). So, there should **not** be "N" RPCs, if "N" pods are using the same volume.

If the concern is that always selecting the first `publishPath` from multiples could lead to false positives, as the selected `publishPath` (or its corresponding pod) could have an issue while the others are somehow healthy. Then yeah, we might have to gate this volume condition feature.
Plz let me know ur thoughts ??

Also, assuming we do gate it, should we close the NFS PR (https://github.com/ceph/ceph-csi/pull/6383) ?? cause those changes would no longer be needed, CephNFS (or any non-staging driver) wouldn't need to advertise volume conditions.

Madhu-1: from kubelet or kubernetes its always N RPC if we are doing for publish path because publish path is specific to a pod. do we just select one publish path and leave others. if yes that would be a bug in the csi-addons. 

i dont see much helpful for NFS, @nixpanic WDYT?

Madhu-1 (dismissed):

sidecar/main.go
nixpanic: This project should be able to support any CSI-driver.

In general, the cleanest most likely us to use the staging path. This would be the same for all pods that use the same PV.

Not all drivers use stage/unstage, and Ceph-CSI/NFS is just one example of those. If the `RPC_STAGE_UNSTAGE_VOLUME` NodeServiceCapability is not present, the publish path should be used.

---

**[Off topic]**
As far as Ceph-CSI goes, it should support volume health for all six backends/modes:
- CephFS (filesystem mode only)
- RBD in filesystem mode
- RBD in block mode
- NFS (filesystem mode only)
- NVMe-oF in filesystem mode
- NVMe-oF in block mode

sidecar/internal/volume-condition/pvc_annotation_recorder.go
nixpanic: indeed, more testing is very welcome! That is one of the reasons this feature was not promoted from beta state yet.

sidecar/internal/volume-condition/pvc_annotation_recorder.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	return nil
+}
+
+func (par *pvcAnnotationRecorder) needsPVC() bool {
+	return true
+}
+
+func (par *pvcAnnotationRecorder) recordUnchangedVolumes() bool {
```
nixpanic, line 115: Why is this important for the annotation? Might be useful to have that as a comment.

config/rbac/csiaddons_pvc_editor_role.yaml
nixpanic: The code patches the PVC, I think the `update` verb is not required?

sidecar/internal/volume-condition/pvc_annotation_recorder.go
SanjalKatiyar: added a comment...

config/rbac/csiaddons_pvc_editor_role.yaml
SanjalKatiyar: removed...

sidecar/internal/volume-condition/pvc_annotation_recorder.go
SanjalKatiyar: created a GH issue as well: https://github.com/csi-addons/kubernetes-csi-addons/issues/1065.

nixpanic (approved):

Madhu-1 (approved):

PR #1050, opened 2026-06-18
flags: (sidecar) enable volume condition reporting by default (GA)

SanjalKatiyar:
Enable '--enable-volume-condition' by default (GA).

Other related changes:
`feat: (sidecar) annotate PVCs with volume health status` https://github.com/csi-addons/kubernetes-csi-addons/pull/1048
`feat: (controller) add periodic PVC stale volume health cleanup` https://github.com/csi-addons/kubernetes-csi-addons/pull/1055

### DIFF
--- a/sidecar/internal/volume-condition/reporter.go
+++ b/sidecar/internal/volume-condition/reporter.go
@@ -68,7 +68,12 @@ func NewVolumeConditionReporter(
 	}
 
 	if !drv.SupportsVolumeCondition() {
-		return nil, fmt.Errorf("driver %q does not support volume-condition", drivername)
+		// volume condition is enabled by default, but drivers can still
+		// choose not to support it. Logging instead of calling os.Exit(1) (in main.go) so
+		// that only this feature is skipped and the rest of the sidecar
+		// process continues running normally.
+		logger.Info(fmt.Sprintf("driver %q does not support volume-condition", drivername))
+		return nil, nil
 	}
 
 	n, err := node.NewNode(ctx, client, hostname)
--- a/sidecar/main.go
+++ b/sidecar/main.go
@@ -82,7 +82,7 @@ func main() {
 		enableAuthChecks            = flag.Bool("enable-auth", true, "Enable Authorization checks and TLS communication (enabled by default)")
 
 		// volume condition reporting
-		enableVolumeCondition    = flag.Bool("enable-volume-condition", false, "Enable reporting of the volume condition")
+		enableVolumeCondition    = flag.Bool("enable-volume-condition", true, "Enable reporting of the volume condition (enabled by default)")
 		volumeConditionInterval  = flag.Duration("volume-condition-interval", 1*time.Minute, "Interval between volume condition checks")
 		volumeConditionRecorders = flag.String("volume-condition-recorders", defaultVolumeConditionRecorders, "location(s) to report volume condition to")
 	)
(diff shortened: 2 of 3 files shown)


sidecar/internal/volume-condition/reporter.go
```diff
@@ -68,7 +68,12 @@ func NewVolumeConditionReporter(
 	}
 
 	if !drv.SupportsVolumeCondition() {
-		return nil, fmt.Errorf("driver %q does not support volume-condition", drivername)
+		// volume condition is enabled by default, but drivers can still
+		// choose not to support it. Logging instead of calling os.Exit(1) (in main.go) so
+		// that only this feature is skipped and the rest of the sidecar
+		// process continues running normally.
+		logger.Info(fmt.Sprintf("driver %q does not support volume-condition", drivername))
+		return nil, nil
```
SanjalKatiyar, line 76: Added this change because of https://github.com/csi-addons/kubernetes-csi-addons/blob/main/sidecar/main.go#L219.
But, happy to discuss if we have some other expectation here as part of this feature GA.

sidecar/main.go
```diff
@@ -218,6 +218,14 @@ func main() {
...
 				os.Exit(1)
 			}
+			if reporter == nil {
+				// volume condition is enabled by default, but drivers can still
+				// choose not to support it. Logging instead of calling os.Exit(1) so
+				// that only this feature is skipped and the rest of the sidecar
+				// process continues running normally.
+				setupLog.Info("Volume condition reporting is not supported by the driver, skipping", "driver", driver)
+				return
+			}
```
SanjalKatiyar, line 228: https://github.com/csi-addons/kubernetes-csi-addons/pull/1050/changes#r3435348794

sidecar/internal/volume-condition/reporter.go
SanjalKatiyar: cc @nixpanic @Madhu-1 @Rakshith-R

nixpanic:
This looks reasonable to me. Not sure how many CSI-drivers have support for this, and if the feature is tested well enough to be considered completely stable.

sidecar/main.go
```diff
@@ -218,6 +218,14 @@ func main() {
 				setupLog.Error(err, "Failed to create volume condition reporter")
 				os.Exit(1)
 			}
+			if reporter == nil {
+				// volume condition is enabled by default, but drivers can still
+				// choose not to support it. Logging instead of calling os.Exit(1) so
+				// that only this feature is skipped and the rest of the sidecar
+				// process continues running normally.
+				setupLog.Info("Volume condition reporting is not supported by the driver, skipping", "driver", driver)
+				return
```
Madhu-1, line 227: Isnt this need to be error and os.Exit(1)?

SanjalKatiyar: @Madhu-1 plz check: https://github.com/csi-addons/kubernetes-csi-addons/pull/1050#discussion_r3435348794

Madhu-1: missed that one, lets log as error and do os.Exit

SanjalKatiyar: also, @nixpanic's comment here: https://github.com/csi-addons/kubernetes-csi-addons/pull/1050#issuecomment-4770607563

Madhu-1: Thinking more about it, couple of points
1. what happens if this is deployed with the Controller Plugin
2. CSI-addons is used by other SP for DR functionality they might have not implemented it

If we are planning to do major release i am to do breaking changes, if no major release is planned i would like to disable this by default if it causes problem 

@nixpanic @Rakshith-R ^^

SanjalKatiyar: > lets log as error and do os.Exit

If any driver does not support volume conditions, the entire sidecar will fail and K8s will restart the container. Once restarted, the container will probably fail again with exponential backoff, so on.
That's why, I was not doing os.Exit anymore as volume conditions is default enabled now.

> what happens if this is deployed with the Controller Plugin

Will adding `if (!isController && *enableVolumeCondition) { /* start the volume condition reporter */ }` make sense ??

SanjalKatiyar:
Closing this PR (will be sending the fix for enablement in downstream repo).

PR #1054, opened 2026-06-24
controller: Add exponential backoff on gRPC Aborted

black-dragon74:
When a gRPC request returns `codes.Aborted`, the error was returned directly to controller which requeues with its default ~5ms backoff. This exhausted all retries in under 500ms.

This patch adds exponential backoff when `codes.Aborted` is returned, using `RequeueAfter` instead of the default rate-limited requeue.

Commonly encountered with RelcaimSpace on large PVCs where `fstrim` might take longer to complete. The job was marked failed even when the `fstrim` succeeded (retries exhausted before job completion).


black-dragon74:
@iPraveenParihar @Rakshith-R PTAL at this one, QoL change.

Rakshith-R (approved):

nixpanic (approved):

PR #1067, opened 2026-07-15
controller: fix for missed runs due to stagger window

black-dragon74:
When we stagger a ReclaimSpace/KeyRotationJob, we ask the controller to wakeup later than the raw schedule. This caused a conflict with `StartingDeadlineSeconds` where the runs would be silently skipped.

This patch fixes it by:
- Widening the missed run search window by stagger offset.
- Appying the same stagger to the `lastMissed` so the "too late" check compares like with like.

Unit tests around this have also been added.

### DIFF
--- a/internal/controller/csiaddons/encryptionkeyrotationcronjob_controller.go
+++ b/internal/controller/csiaddons/encryptionkeyrotationcronjob_controller.go
@@ -307,17 +307,22 @@ func getNextScheduleForKeyRotation(
 	} else {
 		earliestTime = krcJob.CreationTimestamp.Time
 	}
+
+	rawNext := sched.Next(now)
+	staggeredNext := utils.GetStaggeredNext(krcJob.UID, rawNext, sched, staggerWindow)
+	staggerOffset := staggeredNext.Sub(rawNext)
+
 	if krcJob.Spec.StartingDeadlineSeconds != nil {
-		// controller is not going to schedule anything below this point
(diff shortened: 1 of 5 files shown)


black-dragon74:
## How a reconcile loop looks like with/without the fix

### Variables
Schedule: `@daily`
Stagger: `1.5h`
StartingDeadlineSeconds: `3600` (1h)

Good to know, the controller will wake up at 1:30 (midnight + stagger), an additonal 1hr is allowed by `StartingDeadlineSeconds`.

### Without the fix
- Missed run search starts at now - deadline; which is: 1:30-1:00 = 00:30
- midnight is before 00:30, no missed found, run skipped.
- Even if it was found, since the controller used raw schedule, we are already too late as: midnight + deadline = 1:00 and the time now is 1:30, past the allowed interval.

### With the fix
- Search starts at: now - deadline - offset, 1:30 - 1h - 1.5 = 22:00 (the previous day)
- Missed midnight run found and returned after staggering: 00:00 + 1.5h = 1:30
- Deadline expires at: 1:30 + 1h = 2:30, now is 1:30 which is before that, the job runs.
- If now were 3:00 due to some genuine reason, the run would be correctly skipped.

**Note**: This is an edge case that only happens if a schedule run was missed.

Rakshith-R (approved):

nixpanic (approved):

PR #1071, opened 2026-07-20
requeue reconcile requests with delay if deletion is blocked for vgrcontent

Nikhil-Ladha:
add requeue for vgrcontent reconciler with delay of 15secs, if deletion is stuck due to presence of VolumeReplication finalizer being present in VGRContent and we are waiting for it be removed by volumereplication controller because sometime due to failed update or multiple consecutive reconcile requests for different resources the update gets missed and the deletion can take a long time.
Also, added a new default called `DefaultRequeueIntervalForVGRResources` which aligns with the requeue interval required for other VGR resources.

### DIFF
--- a/internal/controller/replication.storage/volumegroupreplicationcontent_controller.go
+++ b/internal/controller/replication.storage/volumegroupreplicationcontent_controller.go
@@ -51,6 +51,10 @@ type VolumeGroupReplicationContentReconciler struct {
 	Timeout time.Duration
 }
 
+var defaultRequeueForVGResources = ctrl.Result{
+	RequeueAfter: 15 * time.Second,
+}
+
 //+kubebuilder:rbac:groups=replication.storage.openshift.io,resources=volumegroupreplicationcontents,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=replication.storage.openshift.io,resources=volumegroupreplicationcontents/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=replication.storage.openshift.io,resources=volumegroupreplicationcontents/finalizers,verbs=update
@@ -175,7 +179,9 @@ func (r *VolumeGroupReplicationContentReconciler) Reconcile(ctx context.Context,
 		} else {
 			err = fmt.Errorf("cannot delete VolumeGroupReplicationContent resource, until dependent VolumeReplication instance is deleted")
 			logger.Error(err, "failed to delete VolumeGroupReplicationContent resource")
-			return reconcile.Result{}, err
+			// We should requeue with a delay as sometimes due to a failed update or multiple consecutive
+			// reconcile requests for different resources, the update gets missed and the deletion can take a long time.
+			return defaultRequeueForVGResources, err
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


internal/controller/replication.storage/volumegroupreplicationcontent_controller.go
```diff
@@ -175,7 +175,9 @@ func (r *VolumeGroupReplicationContentReconciler) Reconcile(ctx context.Context,
 		} else {
 			err = fmt.Errorf("cannot delete VolumeGroupReplicationContent resource, until dependent VolumeReplication instance is deleted")
 			logger.Error(err, "failed to delete VolumeGroupReplicationContent resource")
-			return reconcile.Result{}, err
+			return reconcile.Result{
```
black-dragon74: Maybe define it as a constant and/or add comments around the behavior?

Nikhil-Ladha: Done, added a const which is reused across other VGR controllers for consistency.

black-dragon74 (dismissed):
Thank you :)

internal/controller/replication.storage/volumegroupreplicationcontent_controller.go
```diff
@@ -51,6 +51,8 @@ type VolumeGroupReplicationContentReconciler struct {
 	Timeout time.Duration
 }
 
+const DefaultRequeueIntervalForVGRResources = 15 * time.Second
```
nixpanic: does this need to be exported?

internal/controller/replication.storage/volumereplication_controller.go
```diff
@@ -373,7 +373,7 @@ func (r *VolumeReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Re
 				return ctrl.Result{
 					Requeue: true,
 					// Setting Requeue time for 15 seconds
```
nixpanic: This comment is now not suitable anymore, it can easily go out of sync when the default it updated.

Have you considered alternative too?

```go
var defaultRequeueForVGResources = ctrl.Result{
	Requeue: true,
	RequeueAfter: 15 * time.Second,
}
```

internal/controller/replication.storage/volumegroupreplicationcontent_controller.go
Nikhil-Ladha: I don't think so, it is used within the same package only.

internal/controller/replication.storage/volumereplication_controller.go
Nikhil-Ladha: We can use this as well, though I feel just having the duration as a constant would be more simpler?

nixpanic: Depends... if `DefaultRequeueIntervalForVGRResources` is only used when the struct has all other matching attributes, a default struct would be a little cleaner/simpler?

Nikhil-Ladha: Sure, updated to use the default struct.

black-dragon74 (commented):
A nit, can we remove timing info from the comments? Future updates would require updating each comment.

Rest LGTM

black-dragon74 (approved):

Madhu-1 (approved):

PR #1082, opened 2026-07-22
pvc_new_controller: Ignore unrelated events on the PVC

black-dragon74:
Since the new controller is only enabled when precedence is set to
"storageclass", meaning annotations are only read from SC, there
is no need to watch for events/annotations other than Create on the
PVC when this controller is running. The owner refs will GC the
child resources if PVC is deleted.

e2e is also added around this.

### DIFF
--- a/internal/controller/utils/predicates.go
+++ b/internal/controller/utils/predicates.go
@@ -40,29 +40,17 @@ func SilentPredicate() predicate.Predicate {
 	}
 }
 
-// PVCPredicate returns a predicate for PVC watch events.
-// Reconciliation on update is triggered only when
-// one of the relevant annotations changes value. Create and Delete
-// events are always reconciled.
+// PVCPredicate returns a predicate that only allows PVC Create events.
+// When schedule precedence is set to StorageClass, the controller reads
+// configuration from the StorageClass, not PVC annotations, so PVC
+// updates do not require reconciliation. Deletes are also ignored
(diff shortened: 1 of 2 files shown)


internal/controller/utils/predicates.go
```diff
@@ -47,22 +47,9 @@ func SilentPredicate() predicate.Predicate {
...
-			relevantAnnotations := []string{
-				KrcJobScheduleTimeAnnotation,
-				KrEnableAnnotation,
-				RsCronJobScheduleTimeAnnotation,
-				RsEnableAnnotation,
-			}
-
-			return AnnotationValueChanged(e.ObjectOld.GetAnnotations(), e.ObjectNew.GetAnnotations(), relevantAnnotations)
-		},
+		DeleteFunc:  func(e event.DeleteEvent) bool { return false },
```
nixpanic, line 51: Please correct the comment above the function. Don't only include the events, also include a reason why only Create needs reconcile.

nixpanic (changes requested):

internal/controller/utils/predicates.go
black-dragon74: Done :)

Rakshith-R (approved):

black-dragon74:
@nixpanic PTAL :)

nixpanic (approved):

PR #1095, opened 2026-08-14
sidecar/volume-condition: support NodeGetVolumeHealth from CSI spec v1.13.0

nixpanic:
The CSI spec v1.13.0 introduces a dedicated NodeGetVolumeHealth RPC for
volume health reporting and removes VolumeCondition from the
NodeGetVolumeStats response. Update the volume-condition sidecar to use
NodeGetVolumeHealth (GET_VOLUME_HEALTH capability) as the default path.

For CSI drivers that still implement the older VOLUME_CONDITION capability
(CSI spec v1.12.0 and earlier), a fallback is kept in the new
legacycsi package. Because the v1.13.0 generated types no longer include
VolumeCondition in NodeGetVolumeStatsResponse, the fallback uses a custom
gRPC codec and raw protobuf wire-format parsing (protowire) to extract
the field without requiring the old generated code.

Reflect the CSI spec v1.13.0 change: NodeGetVolumeHealth is now the
primary RPC for volume health reporting, with NodeGetVolumeStats kept
as a fallback for older drivers. Add a capability table and update the
Dependencies section to list the capability required for each spec version.

This has been tested manually with Ceph-CSI/RBD and NodeGetVolumeStats.

Closes: #1091

## New Features
- Added support for CSI volume health reporting through the newer `NodeGetVolumeHealth` capability.
- Retained compatibility with older drivers using legacy volume-condition reporting.
- Health messages now use a reason fallback when available.
- Health checks automatically select the most appropriate supported reporting method.
- Improved handling of health responses, including malformed or incomplete data.

## Documentation
- Updated volume health guidance with current CSI version requirements and API references.


sidecar/internal/volume-condition/volume/driver.go
```diff
@@ -88,10 +102,63 @@ func (drv *csiDriver) GetDrivername() string {
...
+	req := &csi.NodeGetVolumeHealthRequest{
+		VolumeId: v.GetVolumeID(),
+	}
+
+	if stagingPath, err := platform.GetPlatform().GetStagingPath(drv.name, v.GetVolumeID()); err == nil {
+		req.StagingTargetPath = stagingPath
+	}
+	if publishPath, err := platform.GetPlatform().GetPublishPath(drv.name, v.GetVolumeID()); err == nil {
+		req.VolumePublishPath = publishPath
+	}
```
black-dragon74: Maybe log when both are empty?

sidecar/internal/volume-condition/volume/legacycsi/legacycsi.go
black-dragon74, line 1: We should add unit tests for the key functions in this file.

black-dragon74 (changes requested):

sidecar/internal/volume-condition/volume/driver.go
nixpanic: yes, probably good to do. Added logging.

sidecar/internal/volume-condition/volume/legacycsi/legacycsi.go
nixpanic: Good point, added now.

sidecar/internal/volume-condition/volume/driver.go
```diff
@@ -88,10 +105,73 @@ func (drv *csiDriver) GetDrivername() string {
...
+
+// getVolumeConditionViaVolumeHealth calls NodeGetVolumeHealth (CSI spec v1.13.0+).
+func (drv *csiDriver) getVolumeConditionViaVolumeHealth(v CSIVolume) (VolumeCondition, error) {
+	req := &csi.NodeGetVolumeHealthRequest{
+		VolumeId: v.GetVolumeID(),
+	}
+
+	if stagingPath, err := platform.GetPlatform().GetStagingPath(drv.name, v.GetVolumeID()); err == nil {
+		req.StagingTargetPath = stagingPath
+	}
```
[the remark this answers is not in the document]
black-dragon74: @nixpanic I second this, https://github.com/container-storage-interface/spec/blob/b516e92e3d6c0d27e6c8ffd0a956e3918b50f619/csi.proto#L1823-L1833 dictates this. Gating it will ensure we don't drift away from spec.

black-dragon74 (changes requested):
One minor nit around gating, rest LGTM :)

sidecar/internal/volume-condition/volume/driver.go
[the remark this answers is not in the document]
nixpanic: Yes, that's a good thing to check!

nixpanic:
@coderabbitai resolve

black-dragon74 (approved):
Thank you :)

iPraveenParihar (approved):

PR #1105, opened 2026-08-25
ci: fix golangci-lint errors

nixpanic:
ctrl.Result.Requeue is deprecated in favour of RequeueAfter.  Replace all
Requeue: true usages in the PVC controllers with RequeueAfter using a shared
pvcRequeueDelay constant (30s).

staticcheck SA4023 flagged that the error returned by nodeMgr.DispatchWatcher()
is never nil, making the `if err != nil` comparison always true. Remove the
redundant guard and call setupLog.Error and os.Exit unconditionally.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved persistent volume claim processing by retrying pending or unavailable connections after a consistent 30-second delay.
  - Reduced unnecessary immediate retry activity while claims are waiting to become ready.
  - Improved monitoring behavior when the storage watcher stops, including clear logging and controlled termination.

### DIFF
--- a/internal/controller/csiaddons/persistentvolumeclaim_controller.go
+++ b/internal/controller/csiaddons/persistentvolumeclaim_controller.go
@@ -73,6 +73,11 @@ const (
 
 	relciamSpaceOp Operation = "reclaimspace"
 	keyRotationOp  Operation = "keyrotation"
+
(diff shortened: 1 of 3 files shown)


nixpanic:
Already fixed withj #1103

